PR #7, opened 2024-09-23
feat(worldchain-builder): Add `verified_blockspace_cap` to `WorldChainPayloadBuilder`

0xKitsune:
This PR adds `verified_blockspace_cap` to the `WorldChainPayloadBuilder` which specifies the maximum amount of blockspace that should be occupied by verified transactions.

### DIFF
--- a/world-chain-builder/src/payload/builder.rs
+++ b/world-chain-builder/src/payload/builder.rs
@@ -52,7 +52,7 @@ use crate::pool::tx::WorldChainPoolTransaction;
 pub struct WorldChainPayloadBuilder<EvmConfig> {
     inner: OptimismPayloadBuilder<EvmConfig>,
     // TODO: docs describing that this is a percent, ex: 50 is 50/100
-    verified_blockspace_cap: u64,
+    verified_blockspace_capacity: u64,
     _database_env: Arc<DatabaseEnv>,
 }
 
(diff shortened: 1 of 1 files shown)


world-chain-builder/src/pool/noop.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    fn get_all_blobs_exact(
+        &self,
+        tx_hashes: Vec<TxHash>,
+    ) -> Result<Vec<BlobTransactionSidecar>, BlobStoreError> {
+        if tx_hashes.is_empty() {
+            return Ok(vec![]);
+        }
+        Err(BlobStoreError::MissingSidecar(tx_hashes[0]))
```
Dzejkop, line 235: Missing sidecar?

0xKitsune: The `NoopWorldchainTransactionPool` is essentially the `NoopTransactionPool` ported over from Reth. It is only used when [building empty payloads](https://github.com/worldcoin/world-chain/blob/dev/world-chain-builder/src/payload/builder.rs#L137) (which features the same logic contained in the [OptimismPayloadBuilder::build_empty_payload](https://github.com/paradigmxyz/reth/blob/efa5d45e4ea500b3c4b21755899c22233eb6d61c/crates/optimism/payload/src/builder.rs#L128-L146)). The sidecar in this context relates to blobs rather than the PBS sidecar.

Dzejkop: gotcha, makes sense. Thanks!

PR #18, opened 2024-10-02
Parse external nullifier and validates its hash matches provided external nullifier hash

Dzejkop:


world-chain-builder/src/pbh/semaphore.rs
```diff
@@ -50,11 +58,157 @@ pub struct SemaphoreProof {
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ExternalNullifier {
+    pub month: MonthMarker,
+    pub nonce: u16,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MonthMarker {
```
0xForerunner: This all looks great! Only thing I would change is renaming this to DateMarker. When reading through I assumed this only contained the month at first.

world-chain-builder/src/pool/validator.rs
```diff
@@ -100,31 +96,32 @@ where
...
+            .parse()
+            .map_err(WorldChainTransactionPoolInvalid::InvalidExternalNullifier)
+            .map_err(TransactionValidationError::from)?;
+
+        let current_date = MonthMarker::from(date);
+        if external_nullifier.month != current_date {
+            return Err(WorldChainTransactionPoolInvalid::InvalidExternalNullifierPeriod.into());
         }
 
-        // TODO: Handle edge case where we are at the end of the month
```
0xForerunner, line 118: We should either solve this TODO, leave it, or create a linear task for it, (or maybe a test case). This will very likely cause problems if not handled. I suggest we have a 1 minute buffer period at the end of the month where we accept both date markers.

Dzejkop: All right, will do!

PR #19, opened 2024-10-02
Osiris/e2e tests

0xOsiris:
Tests Block building, and priority ordering e2e through the `WorldChainEthApi`

### DIFF
--- a/world-chain-builder/src/e2e_tests/mod.rs
+++ b/world-chain-builder/src/e2e_tests/mod.rs
@@ -52,6 +52,7 @@ use semaphore::{
 use std::{
     collections::{BTreeMap, HashMap},
     sync::Arc,
+    time::Duration,
 };
 
 pub const DEV_CHAIN_ID: u64 = 8453;
@@ -221,8 +222,7 @@ impl WorldChainBuilderTestContext {
 
 #[tokio::test]
 async fn test_can_build_pbh_payload() -> eyre::Result<()> {
-    // std::env::set_var("RUST_LOG", "info");
-    reth_tracing::init_test_tracing();
+    tokio::time::sleep(Duration::from_secs(1)).await;
     let mut ctx = WorldChainBuilderTestContext::setup().await?;
     let mut pbh_tx_hashes = vec![];
     for signer in ctx.pbh_wallets.iter() {
@@ -240,7 +240,7 @@ async fn test_can_build_pbh_payload() -> eyre::Result<()> {
     let block_hash = payload.block().hash();
     let block_number = payload.block().number;
 
-    let tip = pbh_tx_hashes[0].clone();
+    let tip = pbh_tx_hashes[0];
     ctx.node
         .assert_new_block(tip, block_hash, block_number)
         .await?;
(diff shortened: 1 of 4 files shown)


world-chain-builder/src/e2e_tests/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        )?)
+        .launch()
+        .await?;
+
+    let wallets = Wallet::new(10).with_chain_id(chain_id).gen();
+
+    return Ok((NodeTestContext::new(node).await?, wallets, chain_id, tasks));
+}
+
+type Adapter = NodeAdapter<
```
Dzejkop, line 60: wow

Dzejkop: Can we move this and NodeTestContextType to the top of the file? Right below imports?

world-chain-builder/src/e2e_tests/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    >,
+>;
+
+pub type NodeTestContextType = NodeTestContext<Adapter, WorldChainAddOns>;
+
+#[tokio::test]
+async fn test_can_build_mixed_pbh_payload() -> eyre::Result<()> {
+    reth_tracing::init_test_tracing();
+    // Create a raw signed transfer
+    let raw_tx = TransactionTestContext::transfer_tx_bytes(8453, DEV_SIGNER.parse().unwrap()).await;
```
Dzejkop: Can we extract 8453 to a const? And also use it in the second call to `transfer_tx_bytes`?

world-chain-builder/src/e2e_tests/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    // Re-encode the envolope
+    let mut buff = Vec::<u8>::new();
+    world_chain_pooled_tx_element.encode_enveloped(&mut buff);
+
+    // Pre-validate the decoding
+    let pooled_transaction_element = recover_raw_transaction(buff.clone().into());
+    assert!(
+        pooled_transaction_element.is_ok_and(|e| e.0.semaphore_proof.is_some_and(|p| p == proof))
+    );
```
Dzejkop: We could buble the error here and assert directly on the value, no?

Dzejkop: Also this is just checking our encoding/decoding? Feels a bit redundant

world-chain-builder/src/e2e_tests/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    let (mut node, mut dev_wallets, id, _tasks) = setup(chain_spec).await?;
+    let std_wallet = dev_wallets.pop().unwrap();
+    // inject normal tx
+    let raw_tx = TransactionTestContext::transfer_tx_bytes(id, std_wallet.clone()).await;
+    // Call eth_sendRawTransaction with the enveloped semaphore tx.
+    let semaphore_res = node.rpc.inject_tx(buff.into()).await;
+    let res = node.rpc.inject_tx(raw_tx).await;
+
+    assert!(res.is_ok());
+    assert!(semaphore_res.is_ok());
```
Dzejkop: Let's just buble the errors with `?`, there's really no difference, and bubbling gives better error visiblity

world-chain-builder/src/e2e_tests/mod.rs
0xForerunner: Damn that is a serious type

world-chain-builder/src/e2e_tests/mod.rs
0xOsiris: Yeah I added this to isolate a bug early on. Will remove!

PR #24, opened 2024-10-07
Update Reth

0xForerunner:
This was a pain haha. Everything seems to be working after the update though so we should be good to go.

### DIFF
--- a/world-chain-builder/Cargo.toml
+++ b/world-chain-builder/Cargo.toml
@@ -10,55 +10,54 @@ members = ["crates/*"]
 
 [dependencies]
 # reth
-reth = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
-reth-revm = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
-reth-cli-util = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
-reth-cli = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
-reth-node-builder = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
-reth-chainspec = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
-reth-evm = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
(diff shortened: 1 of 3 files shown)


world-chain-builder/Cargo.toml
```diff
@@ -10,72 +10,76 @@ members = ["crates/*"]
...
+reth-chainspec = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-evm = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-db = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-provider = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-network-api = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-node-api = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+reth-rpc = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
+# reth-rpc-types = { git = "https://github.com/paradigmxyz/reth", rev = "093b1a1" }
```
0xOsiris: Can we remove this if we aren't using it?

world-chain-builder/src/pool/validator.rs
```diff
@@ -706,47 +702,10 @@ pub mod tests {
...
-        let inner = EthPooledTransaction::new(tx, 0);
-        let tx = WorldChainPooledTransaction {
-            inner,
-            semaphore_proof: Some(semaphore_proof.clone()),
-        };
-
-        validator.set_validated(&tx, &semaphore_proof).unwrap();
-    }
 
-    #[test]
```
0xOsiris, line 719: Just curious the reason why we need to remove the deposit tx validation test?

0xForerunner: It wasn't a very useful test. I can add it back if you think we want it!

world-chain-builder/Cargo.toml
0xForerunner: Oh yeah I'll get rid of that.

PR #27, opened 2024-10-10
Store pbh nullifiers in database after block is sealed

0xForerunner:
Open to comments on the approach used here. I think this is probably the path of least resistance. Alternatively we could cache the nullifiers when the block is sealed and finalize them on confirmation, however I'm not sure it's worth the trouble.

### DIFF
--- a/world-chain-builder/src/external_nullifier.rs
+++ b/world-chain-builder/src/external_nullifier.rs
@@ -49,8 +49,8 @@ pub enum ExternalNullifierParsingError {
     #[error("error parsing prefix - {0}")]
     InvalidPrefix(strum::ParseError),
 
-    #[error("error parsing month - {0}")]
-    InvalidMonth(DateMarkerParsingError),
+    #[error("error parsing date marker - {0}")]
+    InvalidDateMarker(#[from] DateMarkerParsingError),
 
     #[error("error parsing nonce - {0}")]
     InvalidNonce(std::num::ParseIntError),
@@ -70,14 +70,11 @@ impl FromStr for ExternalNullifier {
             });
         }
 
-        // no need to check the exact value since there's only one variant
         let prefix: Prefix = parts[0]
             .parse()
             .map_err(ExternalNullifierParsingError::InvalidPrefix)?;
 
-        let date_marker = parts[1]
-            .parse()
-            .map_err(ExternalNullifierParsingError::InvalidMonth)?;
+        let date_marker = parts[1].parse()?;
 
         let nonce_str = parts[2];
         let nonce_str_trimmed = nonce_str.trim_start_matches('0');


world-chain-builder/src/pbh/semaphore.rs
```diff
@@ -42,13 +42,27 @@ impl Encodable for Proof {
...
+    /// A string containing a prefix, the date marker, and the pbh nonce
     pub external_nullifier: String,
+    /// The hash of the external nullifier
     pub external_nullifier_hash: Field,
+    /// A nullifier hash used to keep track of
+    /// previously used pbh transactions
     pub nullifier_hash: Field,
+    /// This is the transaction hash
+    /// which associates this proof with a specific transaction
     pub signal_hash: Field,
```
0xForerunner: Do we want to remove the signal_hash from the `PbhPayload`? It's not technically needed here. However I'm not sure how friendly the error message generated from `semaphore-rs` will be be if the signal hash is incorrect.

0xForerunner: I removed it! We can add if back later if we decide we want it for some reason.

world-chain-builder/src/payload/builder.rs
```diff
@@ -538,6 +520,25 @@ where
             // append sender and transaction to the respective lists
             executed_senders.push(tx.signer());
             executed_txs.push(tx.into_signed());
+
+            // add the nullifier to the db
+            if let Some(pbh_payload) = pool_tx.transaction.pbh_payload() {
+                match set_pbh_nullifier(&db_tx, pbh_payload.nullifier_hash) {
```
0xOsiris, line 526: Should we be checking existence of the key here first before insertion? And marking invalid based on that rather than reinserting a duplicate key.

I just don't understand why this `set_pbh_nullifier` invocation would fail in the case of existence

0xForerunner: I'm not 100% sure what you mean.

If we get this condition it mean the nullifier has already been inserted so we can mark as invalid.
`if write.operation == DatabaseWriteOperation::CursorInsert`

My thought here was that it would be slightly faster to do it this way than to first check if the key exists, then if not, insert.

0xOsiris: I'm just wondering why this branch case hits

```rs
Err(DatabaseError::Write(write))
```

0xForerunner: > I'm just wondering why this branch case hits

That case is for when a nullifier already exists in the db.

world-chain-builder/src/external_nullifier.rs
```diff
@@ -58,11 +71,11 @@ impl FromStr for ExternalNullifier {
         }
 
         // no need to check the exact value since there's only one variant
-        let Prefix::V1 = parts[0]
+        let prefix: Prefix = parts[0]
```
Dzejkop, line 73: Can you remove the comment above this line?

PR #37, opened 2024-10-16
PBH ExEx

0xKitsune:
This PR introduces an ExEx that commits nullifier hashes to the `ExecutedPbhNullifierTable` after the transaction has landed onchain.

### DIFF
--- a/world-chain-builder/bin/world-chain-builder.rs
+++ b/world-chain-builder/bin/world-chain-builder.rs
@@ -32,7 +32,7 @@ fn main() {
             let db =
                 load_world_chain_db(data_dir.data_dir(), ext_args.builder_args.clear_nullifiers)?;
 
-            let handle: reth::builder::NodeHandle<reth::builder::NodeAdapter<reth::api::FullNodeTypesAdapter<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>, reth_provider::providers::BlockchainProvider<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>>>, reth::builder::components::Components<reth::api::FullNodeTypesAdapter<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>, reth_provider::providers::BlockchainProvider<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>>>, reth::transaction_pool::Pool<reth::transaction_pool::TransactionValidationTaskExecutor<world_chain_builder::pool::validator::WorldChainTransactionValidator<reth_provider::providers::BlockchainProvider<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>>, world_chain_builder::pool::tx::WorldChainPooledTransaction>>, world_chain_builder::pool::ordering::WorldChainOrdering<world_chain_builder::pool::tx::WorldChainPooledTransaction>, reth::transaction_pool::blobstore::DiskFileBlobStore>, reth_optimism_evm::OptimismEvmConfig, reth_optimism_evm::OpExecutorProvider, std::sync::Arc<dyn Consensus>, reth_optimism_node::engine::OptimismEngineValidator>>, world_chain_builder::node::builder::WorldChainAddOns> = builder
+            let handle = builder
                 .node(WorldChainBuilder::new(ext_args.clone(), db.clone())?)
                 .install_exex("PBHExEx", move |ctx| pbh_exex(ctx, db))
                 .launch()


world-chain-builder/bin/world-chain-builder.rs
```diff
@@ -25,13 +27,14 @@ fn main() {
...
...nv>>, reth_provider::providers::BlockchainProvider<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>>>, reth::builder::components::Components<reth::api::FullNodeTypesAdapter<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>, reth_provider::providers::BlockchainProvider<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>>>, reth::transaction_pool::Pool<reth::transaction_pool::TransactionValidationTaskExecutor<world_chain_builder::pool::validator::WorldChainTransactionValidator<reth_provider::providers::BlockchainProvider<reth::api::NodeTypesWithDBAdapter<WorldChainBuilder, std::sync::Arc<reth_db::DatabaseEnv>>>, world_chain_builder::pool::tx::WorldChainPooledTransaction>>, world_chain_builder::pool::ordering::WorldChainOrdering<world_chain_builder::pool::tx::WorldChainPooledTransaction>, reth::transaction_pool::blobstore::DiskFileBlobStore>, reth_optimism_evm::OptimismEvmConfig, reth_optimism_evm::OpExecutorProvider, std::sync::Arc<dyn Consensus>, reth_optimism_node::engine::OptimismEngineValidator>>, world_chain_builder::node::builder::WorldChainAddOns> = builder
```
0xOsiris: Is this necessary?😅

0xKitsune: Definitely not, this must have auto filled. Just pushed a patch to remove this.

world-chain-builder/src/exex/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+                            if let Some(nullifier) = get_validated_nullifier(&db_tx, tx.hash())? {
+                                set_executed_nullifier(&db_tx, nullifier)?;
+                            }
+                        }
+                    }
+                }
+                ExExNotification::ChainReorged { old, new } => {
+                    // Remove old nullifiers from reorged chain
+                    for (_, sealed_block) in old.blocks() {
+                        for tx in sealed_block.transactions() {
```
0xOsiris, line 37: Not necessary to change now, but we could consider slotting the nullifiers in the database as a `Vec<<Nullifier Hash>>` key'd by the block hash. Then we could have a single database read per block instead of one per pbh tx in the block.

0xOsiris: But I think this makes sense regardless

world-chain-builder/src/payload/builder.rs
```diff
@@ -665,11 +640,6 @@ where
     let sealed_block = block.seal_slow();
     debug!(target: "payload_builder", ?sealed_block, "sealed built block");
 
-    // commit the pbh nullifiers to the db
```
0xOsiris, line 668: Where are we committing the nullifier's to be fetched in the ExEx?

world-chain-builder/src/exex/mod.rs
0xForerunner: yeah this is a decent amount of reads. I think it should probably be fine though.

0xKitsune: Agreed, happy to look further into optimizing this.

world-chain-builder/src/exex/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        while let Some(notification) = ctx.notifications.try_next().await? {
+            let db_tx = pbh_db.tx_mut()?;
+
+            match &notification {
+                ExExNotification::ChainCommitted { new } => {
+                    // Insert executed nullifiers for the new block
+                    for (_, sealed_block) in new.blocks() {
+                        for tx in sealed_block.transactions() {
+                            if let Some(nullifier) = get_validated_nullifier(&db_tx, tx.hash())? {
+                                set_executed_nullifier(&db_tx, nullifier)?;
```
0xForerunner, line 29: There is a potential problem here.
- We validate a pbh tx A with nullifier X
- We validate a different pbh tx B with the same nullifier X
- A is included in block n
- B is included in block n+1 as a non-pbh tx
- We attempt to insert the same nullifier twice here.
- Unhandled Error.

0xForerunner: @0xKitsune

0xKitsune: Ahh I was under the impression that we check if the nullifier has been seen upon validation when the tx is inserted into the pool. After review I see that we check the `ExecutedPbhNullifierTable` during validation. Is there a reason that we do not check the `ValidatedPbhNullifierTable` instead? This would solve the issue you mentioned above.

0xForerunner: Solving this robustly is actually kinda hard. Checking the `ValidatedPbhNullifierTable` is also insufficient since it is just a map from tx_hash -> nullifier_hash. Nullifiers are not guaranteed to be unique here.

0xForerunner: You _could_ create another table which guarantees validated nullifier uniqueness, but there are still additional problems. Specifically, just because a tx was included in a block, doesn't mean that it was included as pbh. I think you would need yet again another table which tracks pbh block hashes.

0xKitsune: I think we could consider simply dropping the `ExecutedPbhNullifierTable`, and only using the `ValidatedPbhNullifierTable` instead. We can validate nullifier hash uniqueness upon initial validation. This would imply that once a PBH transaction is sent, it will eventually be included onchain (as long as it clears validation). 

If we go this route, no need for this ExEx.

PR #46, opened 2024-10-25
Osiris/better tx validation

0xOsiris:


world-chain-builder/src/pool/validator.rs
```diff
@@ -131,14 +131,17 @@ where
         transaction: &Tx,
         payload: &PbhPayload,
     ) -> Result<(), TransactionValidationError> {
+        // Validate the root first to prevent invalidating a valid transaction with a root that has
+        // not landed on chain yet
+        self.validate_root(payload)?;
+
```
0xKitsune, line 137: Can we also move `validate_external_nullifier()` here to validate the nullifier before inserting into the db?

PR #62, opened 2024-11-04
Osiris/kub backend

0xOsiris:
This was a pain😅 
- Run's devnet with a Kubernetes backend to patch resource limit issues
- Upgrades test fixtures to the current month
- Changes WorldId contract address from Optimism -> WorldChain deployed address.
- Cleans up devnet `.github` folder

### DIFF
--- a/.github/workflows/run-k8s.yml
+++ b/.github/workflows/run-k8s.yml
@@ -1,10 +1,8 @@
 name: K8s CI
 on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
   workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 1'
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-k8s.yml
```diff
@@ -0,0 +1,91 @@
+name: K8s CI
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
```
0xKitsune: This seems like a heavy workflow to run on every pull/push. We should consider running this on a schedule.

```
name: K8s CI
on:
  workflow_dispatch:
  schedule:
    - cron: '0 0 * * 1' # Runs every Monday at midnight UTC

```

Issue #72, opened 2024-12-04
Extend RPC modules to implement `eth_sendRawTransactionConditional`


0xOsiris:
Extend the `WorldChainEthApi` to support `eth_sendRawTransactionConditional`. 

A strict implementation will require adding [ConditionalOptions](https://github.com/alloy-rs/alloy/blob/2dea91e21b83f8902a9b0033a623393d33c78155/crates/rpc-types-eth/src/eip4337.rs#L10) to the `WorldChainPooledTransactionElement` that such that the bundle tx can be fill or kill when ordering the block.

References:
- https://notes.ethereum.org/@yoav/SkaX2lS9j
- https://docs.optimism.io/stack/features/send-raw-transaction-conditional

PR #73, opened 2024-12-09
Osiris/ext ethapi

0xOsiris:
Addresses #72

This PR creates an RPC server in the builder that Extends the EthApi with `eth_sendRawTransactionConditional`. This RPC method validates the `ConditionalOptions` at the RPC layer prior to inserting the transaction into the transaction pool. 

I have excluded Transaction forwarding to the `sequencer_http` from the implementation as this will likely be used by Pbh bundles, but open to any feedback. 

Integration CI Test: https://github.com/worldcoin/world-chain/actions/runs/12244869012


world-chain-builder/crates/assertor/src/main.rs
```diff
@@ -128,11 +128,28 @@ where
...
+            .await?;
+        queue.push(PendingTransactionBuilder::new(
+            builder_provider.root(),
+            tx_hash,
+        ));
+    }
+
     let futs = queue.into_iter().map(|builder| async {
         let hash = builder.tx_hash().clone();
         let receipt = builder.get_receipt().await;
```
0xForerunner:
```suggestion
    let num_transactions = fixture.fixture.len();
    let half = num_transactions / 2;

    let builder_provider_clone = builder_provider.clone();

    stream::iter(fixture.fixture.iter().enumerate())
        .map(Ok)
        .try_for_each_concurrent(CONCURRENCY_LIMIT, move |(index, transaction)| {
            let builder_provider = builder_provider_clone.clone();
            async move {
                let tx = if index < half {
                    // First half, use send_raw_transaction
                    builder_provider.send_raw_transaction(transaction).await?
                } else {
                    // Second half, use eth_sendRawTransactionConditional
                    let rlp_hex = hex::encode_prefixed(transaction);
                    let tx_hash = builder_provider
                        .client()
                        .request(
                            "eth_sendRawTransactionConditional",
                            (rlp_hex, ConditionalOptions::default()),
                        )
                        .await?;
                    PendingTransactionBuilder::new(builder_provider.root(), tx_hash)
                };
                let hash = *tx.tx_hash();
                let receipt = tx.get_receipt().await;
                assert!(receipt.is_ok());
                debug!(
                    receipt = ?receipt.unwrap(),
                    hash = ?hash,
                    index = index,
                    "Transaction Receipt Received"
                );
                Ok::<(), eyre::Report>(())
            }
        })
        .await?;
```

0xForerunner (approved):
Looks good man!

world-chain-builder/src/rpc/bundle.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        if let Some(min_timestamp) = options.timestamp_min {
+            if min_timestamp > latest.timestamp {
+                return Err(ErrorCode::from(-32003).into());
+            }
+        }
+
+        if let Some(max_timestamp) = options.timestamp_max {
+            if max_timestamp <= latest.timestamp {
```
0xKitsune, line 113: Should this be `if max_timestamp < latest.timestamp`? If the timestamp is equal to the max timestamp I am under the impression that the bundle is still valid?

world-chain-builder/src/rpc/bundle.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.validate_known_accounts(options.known_accounts, latest.header.number.into())?;
+
+        if let Some(min_block) = options.block_number_min {
+            if min_block > latest.number {
+                return Err(ErrorCode::from(-32003).into());
+            }
+        }
+
+        if let Some(max_block) = options.block_number_max {
+            if max_block <= latest.number {
```
0xKitsune, line 101: Should this be `if max_block < latest.number`?

world-chain-builder/src/rpc/bundle.rs
0xOsiris: If `max_timestamp == latest.timestamp` then the next block will always have the property `max_timestamp < next_block.timestamp` which would be invalid. Was the thought

0xKitsune: Sgtm makes sense! Consider adding a comment here as well.

0xOsiris: I believe we are picking up the last confirmed block here, not the latest pending. But let me double check.

> Yeah this is the latest block, not the latest pending block.

world-chain-builder/src/rpc/bundle.rs
0xOsiris: I think it should be `<=` as the pending block will always be > then `latest.number`

world-chain-builder/src/rpc/bundle.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        for (address, storage) in known_accounts.into_iter() {
+            match storage {
+                AccountStorage::Slots(slots) => {
+                    for (slot, value) in slots.into_iter() {
+                        let current = state.storage(address, slot.into()).map_err(|e| {
+                            ErrorObject::owned(
+                                ErrorCode::InternalError.code(),
+                                e.to_string(),
+                                Some(""),
```
0xKitsune, line 141: In cases where we are returning an empty string here, can we just return `None` instead?

0xKitsune (approved):

PR #74, opened 2024-12-11
PBHVerifier

0xForerunner:
Smart contract that verifies PBH Proofs on chain.


pbh-verifier/src/PBHVerifier.sol
```diff
@@ -0,0 +1,165 @@
...
+    * @param callData              - The method call to execute on this account.
+    * @param accountGasLimits      - Packed gas limits for validateUserOp and gas limit passed to the callData method call.
+    * @param preVerificationGas    - Gas not calculated by the handleOps method, but added to the gas paid.
+    *                                Covers batch overhead.
+    * @param gasFees               - packed gas fields maxPriorityFeePerGas and maxFeePerGas - Same as EIP-1559 gas parameters.
+    * @param paymasterAndData      - If set, this field holds the paymaster address, verification gas limit, postOp gas limit and paymaster-specific extra data
+    *                                The paymaster will pay for the transaction instead of the sender.
+    * @param signature             - Sender-verified signature over the entire request, the EntryPoint address and the chain ID.
+    */
+    struct PackedUserOperation {
```
0xOsiris: Can we inherit this type from the `account-abstraction` contracts?

0xForerunner: Haha yeah for sure :) sorry this isn't even close to being ready for review haha. I just opened the PR to show the external nullifier implementation!

pbh-verifier/src/PBHVerifier.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                                  Vars                                  ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @dev The World ID instance that will be used for verifying proofs
+    IWorldIDGroups internal immutable worldId;
+
+    /// @dev The World ID group ID (always 1)
+    uint256 internal immutable groupId = 1;
```
0xKitsune: Nit: Consider `SCREAMING_SNAKE_CASE` for immutables, [similar to constants](https://docs.soliditylang.org/en/latest/style-guide.html?utm_source=chatgpt.com#constants). There is no hard and fast rule for this but this is a common pattern.

0xKitsune:
Looks great, very clean!

pbh-verifier/src/PBHVerifier.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    /// @param userOp A packed user operation, used to generate the signal hash
+    /// @param root The root of the Merkle tree (returned by the JS widget).
+    /// @param nullifierHash The nullifier hash for this proof, preventing double signaling (returned by the JS widget).
+    /// @param proof The zero-knowledge proof that demonstrates the claimer is registered with World ID (returned by the JS widget).
+    function verifyPbhProof(
+        PackedUserOperation memory userOp,
+        uint256 root,
+        uint256 nullifierHash,
+        ExternalNullifier memory externalNullifier,
```
karankurbur: how are we getting the externalNullifier data here - do we need to add this with the proof in the signature

or is this determined by the builder?

0xOsiris: @karankurbur Yes, this should be decoded from the `signature` on the `UserOperation`

0xOsiris (approved):
This looks good as a first pass, very clean! One small nit would be to rename the top level directory to `contracts` since the `PbhAggregator` I'm assuming will live in the foundry workspace + different contracts, and utils. 

Additionally can we simply emit the proof instead of verifying on chain after checking the uniqueness of the nullifier hash for the builder to verify during transaction validation.

pbh-verifier/.github/workflows/test.yml
0xOsiris, line 1: Can we also move this workflow to the root `.github/workflows`

0xKitsune (approved):

pbh-verifier/src/helpers/PBHExternalNullifier.sol
```diff
@@ -0,0 +1,62 @@
...
+    error InvalidPbhNonce();
+
+    /// @notice Encodes a PBH external nullifier using the provided year, month, and nonce.
+    /// @param pbhNonce An 8-bit nonce value (0-255) used to uniquely identify the nullifier within a month.
+    /// @param month An 8-bit 1-indexed value representing the month (1-12).
+    /// @param year A 16-bit value representing the year (e.g., 2024).
+    /// @return The encoded PBHExternalNullifier.
+    function encode(uint8 pbhNonce, uint8 month, uint16 year) internal pure returns (uint256) {
+        require(month > 0 && month < 13, InvalidExternalNullifierMonth());
+        require(year <= 9999, InvalidExternalNullifierYear());
```
0xKitsune: Nit: You can use `type(uint16).max` here.

`require(year <= type(uint16).max, InvalidExternalNullifierYear());`

0xForerunner: isn't this a tautology ahah? a uint16 has to be <= type(uint16).max :)

pbh-verifier/.github/workflows/test.yml
0xForerunner: > Additionally can we simply emit the proof instead of verifying on chain after checking the uniqueness of the nullifier hash for the builder to verify during transaction validation.

@0xOsiris could you explain your thoughts here?

0xOsiris: > > Additionally can we simply emit the proof instead of verifying on chain after checking the uniqueness of the nullifier hash for the builder to verify during transaction validation.
> 
> @0xOsiris could you explain your thoughts here?

I mean this is probably up for a wider discussion, but instead of verifying the proof in the `PBHVerifier` we can instead just emit the proof. The builder will pick up the proof(s) when validating the bundle transaction, and verify them off chain. If any proof in the transaction is invalid the bundle will be dropped from the mempool. 

In this case the `PBHVerifier` is only in charge of storing the external nullifier hashes, and reverting the whole bundle on a rate limit. The benefit would be saving ~200,000 gas / UO, downside is higher trust assumptions in the builder

pbh-verifier/src/helpers/PBHExternalNullifier.sol
0xOsiris: Why does it need to be `<=` 9999? 🕐 We have the bits available

pbh-verifier/.github/workflows/test.yml
0xForerunner: Yeah okay I'm into that! I think this could work. Let's chat tomorrow at standup.

contracts/src/PBHVerifier.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {WorldIDProxy} from "@world-id-contracts/abstract/WorldIDProxy.sol";
+
+/// @title PBH Verifier
+/// @author Worldcoin
+/// @notice An implementation of an on chain PBH verifier.
+contract PBHVerifier is WorldIDProxy {
```
0xOsiris, line 10: Small nit, but maybe we just inherit `ERC1967Proxy` here instead of `WorldIDProxy`.  Should be the same thing, just less gas

0xKitsune: +1 on this, here is a link to the [proxy implementation we can inherit](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Proxy.sol).

contracts/src/PBHVerifierImplV1.sol
```diff
@@ -0,0 +1,205 @@
...
+    /// @param nonce Transaction/UserOp nonce.
+    /// @param callData Transaction/UserOp call data.
+    /// @param pbhExternalNullifier External nullifier encoding month, year, and a pbhNonce.
+    /// @param nullifierHash Nullifier hash for this semaphore proof.
+    /// @param proof The zero-knowledge proof that demonstrates the claimer is registered with World ID.
+    event PBH(
+        uint256 indexed root,
+        address indexed sender,
+        uint256  nonce,
+        bytes callData,
```
0xKitsune: Nit: Order `indexed` fields together.

contracts/src/PBHVerifierImplV1.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                                  Vars                                  ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @dev The World ID group ID (always 1)
+    uint256 internal immutable GROUP_ID = 1;
+
+    /// @dev The World ID instance that will be used for verifying proofs
+    IWorldIDGroups internal worldId;
```
0xKitsune: Nit: Lets set this visibility to `external`. It might be useful to be able to query this value.

contracts/src/PBHVerifierImplV1.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    /// @dev Make this configurable
+    uint8 internal numPbhPerMonth;
```
0xKitsune: Lets set this as external. This seems like a useful value for third parties to query.

contracts/src/PBHVerifierImplV1.sol
```diff
@@ -0,0 +1,205 @@
...
+        
+        // If worldId address is set, proceed with on chain verification,
+        // otherwise assume verification has been done off chain by the builder.
+        if (address(worldId) != address(0)) {
+            // We now generate the signal hash from the sender, nonce, and calldata
+            uint256 signalHash = abi.encodePacked(sender, nonce, callData).hashToField();
+
+            // We now verify the provided proof is valid and the user is verified by World ID
+            worldId.verifyProof(root, GROUP_ID, signalHash, nullifierHash, pbhExternalNullifier, proof);
+        }
```
0xKitsune, line 190: Do we anticipate verifying onchain? If not we should remove this logic.

0xForerunner: I'm not sure to be honest. Do you think that's ever something we'll need?

PR #76, opened 2024-12-11
feat(4337): Enforce `ConditionalOptions` check during payload building

0xKitsune:
This PR updates the payload builder to enforce any pooled transaction with `ConditionalOptions` clear validation before including the tx into the payload. In the current implementation, all transactions that fail conditional options checks will be removed from the mempool.

```rust
// --snip--
  while let Some(pool_tx) = best_txs.next() {
        if let Some(conditional_options) = pool_tx.transaction.conditional_options() {
            if let Err(_) = validate_conditional_options(conditional_options, &client) {
                best_txs.mark_invalid(&pool_tx);
                invalid_txs.push(pool_tx.hash().clone());
                continue;
            }
        }

// --snip--
 if !invalid_txs.is_empty() {
   pool.remove_transactions(invalid_txs);
}
```

### DIFF
--- a/world-chain-builder/src/payload/builder.rs
+++ b/world-chain-builder/src/payload/builder.rs
@@ -408,11 +408,13 @@ where
     }
 
     if !attributes.no_tx_pool {
+        let mut invalid_txs = vec![];
         let verified_gas_limit = (verified_blockspace_capacity as u64 * block_gas_limit) / 100;
         while let Some(pool_tx) = best_txs.next() {
             if let Some(conditional_options) = pool_tx.transaction.conditional_options() {
                 if let Err(_) = validate_conditional_options(conditional_options, &client) {
                     best_txs.mark_invalid(&pool_tx);
+                    invalid_txs.push(pool_tx.hash().clone());
                     continue;
                 }
             }
@@ -511,6 +513,10 @@ where
             // append sender and transaction to the respective lists
             executed_senders.push(tx.signer());
             executed_txs.push(tx.into_signed());
+        }
+
+        if !invalid_txs.is_empty() {
+            pool.remove_transactions(invalid_txs);
         }
     }
 
--- a/world-chain-builder/src/payload/builder.rs
+++ b/world-chain-builder/src/payload/builder.rs
@@ -45,6 +45,7 @@ use tracing::{debug, trace, warn};
 
 use crate::pool::noop::NoopWorldChainTransactionPool;
 use crate::pool::tx::WorldChainPoolTransaction;
+use crate::rpc::bundle::validate_conditional_options;
 
 /// Priority blockspace for humans builder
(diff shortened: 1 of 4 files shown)


world-chain-builder/src/payload/builder.rs
```diff
@@ -406,8 +408,17 @@ where
     }
 
     if !attributes.no_tx_pool {
+        let mut invalid_txs = vec![];
         let verified_gas_limit = (verified_blockspace_capacity as u64 * block_gas_limit) / 100;
         while let Some(pool_tx) = best_txs.next() {
+            if let Some(conditional_options) = pool_tx.transaction.conditional_options() {
```
0xOsiris, line 414: I think this should be fine, but instead of computing the state on `BlockId::latest()` we should be computing it on `BlockId::pending()`

world-chain-builder/src/rpc/bundle.rs
```diff
@@ -75,97 +77,106 @@ where
...
+/// See also <https://pkg.go.dev/github.com/aK0nshin/go-ethereum/arbitrum_types#ConditionalOptions>
+pub fn validate_conditional_options<Client>(
+    options: &ConditionalOptions,
+    provider: &Client,
+) -> RpcResult<()>
+where
+    Client: BlockReaderIdExt + StateProviderFactory,
+{
+    let latest = provider
+        .block_by_id(BlockId::latest())
```
0xOsiris: Here let's grab the state on `BlockId::pending()`

world-chain-builder/src/rpc/bundle.rs
```diff
@@ -75,97 +77,106 @@ where
...
+            return Err(ErrorCode::from(-32003).into());
         }
+    }
 
-        if let Some(max_block) = options.block_number_max {
-            if max_block <= latest.number {
-                return Err(ErrorCode::from(-32003).into());
-            }
+    if let Some(max_block) = options.block_number_max {
+        if max_block <= latest.number {
```
0xOsiris: Then change this to a `<`

world-chain-builder/src/rpc/bundle.rs
```diff
@@ -75,97 +77,106 @@ where
...
+            return Err(ErrorCode::from(-32003).into());
         }
+    }
 
-        if let Some(max_timestamp) = options.timestamp_max {
-            if max_timestamp <= latest.timestamp {
-                return Err(ErrorCode::from(-32003).into());
-            }
+    if let Some(max_timestamp) = options.timestamp_max {
+        if max_timestamp <= latest.timestamp {
```
0xOsiris: And this to a `<`

0xOsiris (approved):
Looks good, just one improvement I think we can make!

PR #80, opened 2024-12-13
feat: override

0xOsiris:
Overrides `eth_sendRawTransaction` on the EthApi

### DIFF
--- a/world-chain-builder/crates/world/rpc/src/eth.rs
+++ b/world-chain-builder/crates/world/rpc/src/eth.rs
@@ -4,7 +4,6 @@ use alloy_primitives::{map::HashMap, StorageKey};
 use alloy_rpc_types::erc4337::{AccountStorage, ConditionalOptions};
 use jsonrpsee::{
     core::{async_trait, RpcResult},
-    proc_macros::rpc,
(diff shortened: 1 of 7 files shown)


world-chain-builder/crates/world/rpc/src/eth.rs
```diff
@@ -83,6 +84,20 @@ where
...
+
+    async fn maybe_forward_raw_transaction(&self, tx: Bytes, hash: B256) -> RpcResult<()> {
+        if let Some(client) = self.raw_tx_forwarder().as_ref() {
+            tracing::debug!( target: "rpc::eth",  "forwarding raw transaction to");
+            let _ = client.forward_raw_transaction(&tx).await.inspect_err(|err| {
+                    tracing::debug!(target: "rpc::eth", %err, hash=?*hash, "failed to forward raw transaction");
+                });
+        }
+        Ok(())
+    }
```
0xKitsune: Do we need a separate function for this? it seems minimal enough to put directly in `send_raw_transaction()`.

0xOsiris: It was just duplicate code in `send_raw_transaction `, and `send_raw_transaction_conditional` is why I separated it

PR #83, opened 2024-12-17
Osiris/signature aggregation

0xOsiris:
This PR:
- Adds a `PBHSignatureAggregator` called by the `EntryPoint` as well as the Bundler to aggregate the `PBHPayload`s onto a unified signature.
- Adds a `PBHEntryPoint` contract that implements `handleAggregatedOps` interface

I'm going to separate integration, and further unit testing into a separate PR as this PR is getting fairly large. And will be useful to have in the base branch with all the contracts for continued development.


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -124,22 +86,25 @@ contract PBHVerifierImplV1 is WorldIDImpl {
...
     ///      upgrading. Create a separate initializer function instead.
     ///
-    /// @param _worldId The World ID instance that will be used for verifying proofs. If set to the
+    /// @param __worldId The World ID instance that will be used for verifying proofs. If set to the
     ///        0 addess, then it will be assumed that verification will take place off chain.
     /// @param _numPbhPerMonth The number of allowed PBH transactions per month.
     ///
     /// @custom:reverts string If called more than once at the same initialisation number.
-    function initialize(IWorldIDGroups _worldId, uint8 _numPbhPerMonth) public reinitializer(1) {
+    function initialize(IWorldIDGroups __worldId, IEntryPoint __entryPoint, uint8 _numPbhPerMonth)
```
Dzejkop: Maybe?

```suggestion
    function initialize(IWorldIDGroups worldId, IEntryPoint entryPoint, uint8 numPbhPerMonth)
```

I feel like it'd be in line with the [suggested naming convention](https://docs.soliditylang.org/en/latest/style-guide.html#local-and-state-variable-names)

contracts/src/PBHEntryPoint.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+
+/// @title PBH Verifier
+/// @author Worldcoin
+/// @notice An implementation of an on chain PBH verifier.
+contract PBHEntryPoint is ERC1967Proxy {
```
Dzejkop, line 9: Can we make it a generic proxy?

Dzejkop: Actually, can't we remove this contract and use `ERC1967Proxy` directly?

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -156,49 +121,40 @@ contract PBHVerifierImplV1 is WorldIDImpl {
...
-    /// @notice Sets a new value for numPbhPerMonth.
-    /// @param _newValue The new value set.
-    /// @dev Can only be set by the owner.
-    function setNumPbhPerMonth(uint8 _newValue) external virtual onlyOwner onlyProxy onlyInitialized {
-        numPbhPerMonth = _newValue;
+    /// @notice Validates the hashed operations is the same as the hash transiently stored.
+    /// @param hashedOps The hashed operations to validate.
+    function validateSignaturesCallback(bytes32 hashedOps) external view virtual onlyProxy onlyInitialized {
+        assembly ("memory-safe") {
+            if iszero(eq(tload(_HASHED_OPS_SLOT), hashedOps)) { revert(0, 0) }
```
Dzejkop: Can we use `transient` keyword instead of assembly?

https://docs.soliditylang.org/en/v0.8.28/contracts.html#transient-storage

Dzejkop: seems that VSCode's solidity support doesn't recognize it but seems to compile

contracts/src/PBHEntryPointImplV1.sol
0xKitsune: Yeah agreed, lets update to this imo.

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,176 @@
...
+        uint256 nullifierHash,
+        uint256[8] proof
+    );
+
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                                  Vars                                  ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @dev Gap for transient storage.
+    uint256 private __gap;
```
0xKitsune: Lets avoid double underscores. It is not easy to distinguish between `_var` and `__var`.

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,176 @@
...
+
+    /// @dev The World ID group ID (always 1)
+    uint256 internal immutable _GROUP_ID = 1;
```
0xKitsune: Can we set this as a constant if we are assigning the value here? Or is there value as keeping it as immutable?

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,176 @@
...
 
-/// @title PBH Verifier
+/// @title PBH Verifier Implementation Version 1
 /// @author Worldcoin
-/// @notice An implementation of an on chain PBH verifier.
-/// TODO: Inherit directly from ERC1967Proxy
-contract PBHVerifier is WorldIDProxy {
+/// @notice An implementation of a priority blockspace for humans (PBH) transaction verifier.
+/// @dev This is the implementation delegated to by a proxy.
+contract PBHVerifier is IPBHVerifier, WorldIDImpl {
```
0xKitsune, line 16: Does this need to be a proxy implementation? I think we can deploy this without a proxy.

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,176 @@
...
+    /// @param nullifierHash Nullifier hash for this semaphore proof.
+    /// @param proof The zero-knowledge proof that demonstrates the claimer is registered with World ID.
+    event PBH(
+        uint256 indexed root,
+        address indexed sender,
+        uint256 nonce,
+        bytes callData,
+        uint256 indexed pbhExternalNullifier,
+        uint256 nullifierHash,
+        uint256[8] proof
```
0xKitsune: nit: group indexed fields together

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,176 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                                  Functions                             ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @param sender The sender of this particular transaction or UserOp.
+    /// @param nonce Transaction/UserOp nonce.
+    /// @param callData Transaction/UserOp call data.
+    /// @param pbhPayload The PBH payload containing the proof data.
+    function verifyPbhProof(address sender, uint256 nonce, bytes memory callData, PBHPayload memory pbhPayload)
```
0xKitsune: nit: can we set `pbhPayload` as `calldata` instead of memory? Also maybe we update this function to `verifyPbh` for clarity as its not always verifying the proof.

contracts/test/PBHSignatureAggregator.t.sol
```diff
@@ -0,0 +1,78 @@
...
+        bytes memory sigBuffer = new bytes(65);
+        bytes memory signature = new bytes(65 + proofLength);
+        console.log("Proof Length: ", proofLength);
+        assembly {
+            mstore(add(signature, 65), sigBuffer)
+            mstore(add(signature, add(65, proofLength)), proofData)
+        }
+
+        console.logBytes(signature);
+        uOps[0] = PackedUserOperation({
```
Dzejkop: Let's add one more user op & proof? To ensure the proofs are being aggregated together

contracts/src/PBHEntryPoint.sol
0xOsiris: Yeah thats fine with me! Down for either way here. Really only helpful for differentiating namespace while testing/deploying.

contracts/src/PBHVerifier.sol
0xOsiris: I'll just add transient storage here instead of gapping the slot.

contracts/src/PBHVerifier.sol
0xOsiris: @0xKitsune Can you clarify what you mean here? Do you mean remove the `ERC1967Proxy` altogether?

contracts/test/PBHSignatureAggregator.t.sol
0xOsiris: Yeah sounds good, I'm going to separate this out into a utils contract today and spit out valid `PackedUserOperation[]` that we can actually submit to the `EntryPoint` to test the full flow with a no-op account

contracts/src/PBHVerifier.sol
0xOsiris: I think so

0xOsiris: It cannot be calldata

contracts/src/PBHEntryPointImplV1.sol
0xOsiris: This doesn't compile on my end for some reason on 0.8.28 solc 😅 , evm version is `cancun` in the `foundry.toml` as well.

contracts/src/interfaces/IPBHEntryPoint.sol
```diff
@@ -0,0 +1,15 @@
...
+import {IWorldIDGroups} from "@world-id-contracts/interfaces/IWorldIDGroups.sol";
+import {IEntryPoint} from "@account-abstraction/contracts/interfaces/IEntryPoint.sol";
+import {IPBHVerifier} from "./IPBHVerifier.sol";
+
+interface IPBHEntryPoint is IPBHVerifier {
+    function handleAggregatedOps(
+        IEntryPoint.UserOpsPerAggregator[] calldata opsPerAggregator,
+        address payable beneficiary
+    ) external;
+    function initialize(IWorldIDGroups __worldId, IEntryPoint __entryPoint, uint8 _numPbhPerMonth) external;
```
0xKitsune: nit: can we remove the underscores from the args?

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -156,49 +122,40 @@ contract PBHVerifierImplV1 is WorldIDImpl {
...
-        numPbhPerMonth = _newValue;
+    /// @notice Validates the hashed operations is the same as the hash transiently stored.
+    /// @param hashedOps The hashed operations to validate.
+    function validateSignaturesCallback(bytes32 hashedOps) external view virtual onlyProxy onlyInitialized {
+        assembly ("memory-safe") {
+            if iszero(eq(tload(_hashedOps.slot), hashedOps)) { revert(0, 0) }
+        }
     }
+
+    // TODO: PBH Multicall Entry Point
```
0xKitsune: Lets remove this TODO

0xOsiris: This is a TODO though right?

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,154 @@
...
+    /// @param _numPbhPerMonth The number of allowed PBH transactions per month.
+    function setNumPbhPerMonth(uint8 _numPbhPerMonth) external virtual onlyOwner onlyProxy onlyInitialized {
+        numPbhPerMonth = _numPbhPerMonth;
+    }
+
+    /// @dev If the World ID address is set to 0, then it is assumed that verification will take place off chain.
+    /// @notice Sets the World ID instance that will be used for verifying proofs.
+    /// @param __worldId The World ID instance that will be used for verifying proofs.
+    function setWorldId(address __worldId) external virtual onlyOwner onlyProxy onlyInitialized {
+        _worldId = IWorldIDGroups(__worldId);
```
0xKitsune: We shold emit an event here if we are updating a state variable. Also lets remove the double underscore fro mthe arg.

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,154 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                                  Functions                             ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @param sender The sender of this particular transaction or UserOp.
+    /// @param nonce Transaction/UserOp nonce.
+    /// @param callData Transaction/UserOp call data.
+    /// @param pbhPayload The PBH payload containing the proof data.
+    function verifyPbhProof(address sender, uint256 nonce, bytes calldata callData, PBHPayload memory pbhPayload)
```
0xKitsune: nit: consider changing this function name to `verifyPbh` as this is not always verifying the proof. This might be confusing to anyone that only sees this function called.

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,154 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                                  Vars                                  ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @dev The World ID group ID (always 1)
+    uint256 internal constant _GROUP_ID = 1;
+
+    /// @dev The World ID instance that will be used for verifying proofs
+    IWorldIDGroups internal _worldId;
```
0xKitsune: Lets make this public.

contracts/src/PBHVerifier.sol
```diff
@@ -1,32 +1,154 @@
...
-    ///        This is usually an encoded function call, and allows for initialising the storage of
-    ///        the proxy in a way similar to a traditional solidity constructor.
-    constructor(address _logic, bytes memory _data) payable WorldIDProxy(_logic, _data) {
-        // !!!! DO NOT PUT PROGRAM LOGIC HERE !!!!
-        // It should go in the `initialize` function of the delegate instead.
+    /// @param sender The sender of this particular transaction or UserOp.
+    /// @param nonce Transaction/UserOp nonce.
+    /// @param callData Transaction/UserOp call data.
+    /// @param payload The zero-knowledge proof that demonstrates the claimer is registered with World ID.
+    event PBH(address indexed sender, uint256 indexed nonce, bytes callData, PBHPayload payload);
```
0xKitsune: Should we emit the `calldata`? It seems like it would be more efficient to emit the `signalHash` derived from the UserOp calldata?

contracts/src/PBHSignatureAggregator.sol
```diff
@@ -0,0 +1,76 @@
...
+///         will be considered as Priority User Operations, and will need to pack a World ID proof in the signature field.
+contract PBHSignatureAggregator is IAggregator {
+    /// @notice Thrown when the Hash of the UserOperations is not
+    ///         in transient storage of the `PBHVerifier`.
+    error InvalidUserOperations();
+
+    /// @notice The PBHVerifier contract.
+    IPBHEntryPoint internal immutable _pbhEntryPoint;
+
+    constructor(address __pbhEntryPoint) {
```
0xKitsune: Lets remove `__` and just use `pbhEntryPoint`.

contracts/src/PBHEntryPoint.sol
karankurbur: Agree with @Dzejkop, no need for additional contract

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -126,20 +88,24 @@ contract PBHVerifierImplV1 is WorldIDImpl {
...
     /// @param _worldId The World ID instance that will be used for verifying proofs. If set to the
     ///        0 addess, then it will be assumed that verification will take place off chain.
+    /// @param _entryPoint The ERC-4337 Entry Point.
     /// @param _numPbhPerMonth The number of allowed PBH transactions per month.
     ///
     /// @custom:reverts string If called more than once at the same initialisation number.
-    function initialize(IWorldIDGroups _worldId, uint8 _numPbhPerMonth) public reinitializer(1) {
+    function initialize(IWorldIDGroups _worldId, IEntryPoint _entryPoint, uint8 _numPbhPerMonth)
+        external
+        reinitializer(1)
```
karankurbur, line 97: is the best practice to use `initializer` or `reinitializer(1)` for the first init?

contracts/src/PBHEntryPointImplV1.sol
Dzejkop: odd, let's leave it like that

Dzejkop (approved):
Looks great!

contracts/src/PBHEntryPointImplV1.sol
0xOsiris: I think you use reinitializer to be able to upgrade the initialization steps itself

PR #84, opened 2024-12-18
Dzejkop/testing-and-compilation

Dzejkop:


0xForerunner (approved):
LGTM

world-chain-builder/crates/world/node/benches/validate_transaction.rs
```diff
@@ -1,192 +0,0 @@
-use std::future::Future;
```
0xOsiris, line 1: It might be nice to keep this, so we can get an estimate on how validation time scales with large bundles. But we can add it back when we have more infra setup

0xOsiris (approved):

world-chain-builder/crates/world/node/benches/validate_transaction.rs
Dzejkop: yeah, I only removed it because it seemed to deal with the previous PBH setup. I def want to have benches but let's re-add them once we're done with the builder code imo

PR #85, opened 2024-12-18
Packed External Nullifier repr

Dzejkop:


world-chain-builder/crates/world/pbh/src/external_nullifier.rs
```diff
@@ -1,121 +1,155 @@
...
-    InvaldFormat { actual: String },
-
-    #[error("error parsing prefix - {0}")]
-    InvalidPrefix(strum::ParseError),
-
-    #[error("error parsing date marker - {0}")]
-    InvalidDateMarker(#[from] DateMarkerParsingError),
-
-    #[error("error parsing nonce - {0}")]
-    InvalidNonce(std::num::ParseIntError),
```
0xForerunner: LGTM! Only suggestion would be to leave in the granular errors! Might be nice to have.

0xForerunner (approved):
Looks good!

0xOsiris (approved):
LGTM, nice work!

contracts/src/helpers/PBHExternalNullifier.sol
```diff
@@ -25,25 +30,33 @@ library PBHExternalNullifier {
...
 
+    uint8 public constant V1 = 1;
+
     /// @notice Encodes a PBH external nullifier using the provided year, month, and nonce.
     /// @param pbhNonce An 8-bit nonce value (0-255) used to uniquely identify the nullifier within a month.
     /// @param month An 8-bit 1-indexed value representing the month (1-12).
     /// @param year A 16-bit value representing the year (e.g., 2024).
     /// @return The encoded PBHExternalNullifier.
-    function encode(uint8 pbhNonce, uint8 month, uint16 year) internal pure returns (uint256) {
+    function encode(uint8 version, uint8 pbhNonce, uint8 month, uint16 year) internal pure returns (uint256) {
```
0xKitsune, line 44: If `V1` is constant, we can probably omit `version` from the args and encode the external nullifier with `V1` directly.

contracts/src/helpers/PBHExternalNullifier.sol
```diff
@@ -52,7 +65,8 @@ library PBHExternalNullifier {
     /// @dev This function ensures the external nullifier matches the current year and month,
     ///      and that the nonce does not exceed `numPbhPerMonth`.
     function verify(uint256 externalNullifier, uint8 numPbhPerMonth) public view {
-        (uint8 pbhNonce, uint8 month, uint16 year) = PBHExternalNullifier.decode(externalNullifier);
+        (uint8 version, uint8 pbhNonce, uint8 month, uint16 year) = PBHExternalNullifier.decode(externalNullifier);
```
0xForerunner, line 68: Found an issue here! We're actually under constrained. We need to make sure that the leading zeros here are as expected.
```suggestion
        require(externalNullifier <= type(uint40).max, InvalidExternalNullifierLeadingZeros());
```

contracts/src/helpers/PBHExternalNullifier.sol
```diff
@@ -52,7 +65,8 @@ library PBHExternalNullifier {
     /// @dev This function ensures the external nullifier matches the current year and month,
     ///      and that the nonce does not exceed `numPbhPerMonth`.
     function verify(uint256 externalNullifier, uint8 numPbhPerMonth) public view {
```
0xForerunner, line 73: Found an issue here! We're actually under constrained. We need to make sure that the leading zeros here are as expected.
```suggestion
    function verify(uint256 externalNullifier, uint8 numPbhPerMonth) public view {
     require(externalNullifier <= type(uint40).max, InvalidExternalNullifierLeadingZeros());
```

contracts/src/helpers/PBHExternalNullifier.sol
```diff
@@ -8,11 +8,16 @@ import "@BokkyPooBahsDateTimeLibrary/BokkyPooBahsDateTimeLibrary.sol";
...
-///      - Bits 32-255: Empty
-///      - Bits 16-31: Year
-///      - Bits 8-15: Month
-///      - Bits 0-7: Nonce
+///      - Bits:40-255: Empty
+///      - Bits 32-39: Year
+///      - Bits 16-31: Month
+///      - Bits 8-15: Nonce
+///      - Bits 0-7: Version
 library PBHExternalNullifier {
```
0xForerunner, line 16:
```suggestion
library PBHExternalNullifier {
    /// @notice Thrown when the provided external nullifier doesn't
    /// contain the correct leading zeros
    error InvalidExternalNullifierLeadingZeros();
    
```

contracts/src/helpers/PBHExternalNullifier.sol
Dzejkop: Thanks!

world-chain-builder/crates/world/pbh/src/external_nullifier.rs
Dzejkop: Good point I'll re-add month validation

Dzejkop: Other errors don't really make sense at "parse time"

contracts/src/helpers/PBHExternalNullifier.sol
Dzejkop: it's true but it makes the API asymmetric and makes testing a little uglier. I could add an `encodeV1` function though?

Dzejkop: I'll leave it as is. I'm open to changing it later

PR #86, opened 2024-12-18
setNumPbhPerMonth and associated tests

0xForerunner:


0xOsiris (approved):

contracts/src/helpers/PBHExternalNullifier.sol
```diff
@@ -55,6 +55,6 @@ library PBHExternalNullifier {
         (uint8 pbhNonce, uint8 month, uint16 year) = PBHExternalNullifier.decode(externalNullifier);
         require(year == BokkyPooBahsDateTimeLibrary.getYear(block.timestamp), InvalidExternalNullifierYear());
         require(month == BokkyPooBahsDateTimeLibrary.getMonth(block.timestamp), InvalidExternalNullifierMonth());
-        require(pbhNonce <= numPbhPerMonth, InvalidPbhNonce());
+        require(pbhNonce < numPbhPerMonth, InvalidPbhNonce());
```
0xKitsune, line 58: I might be out of the loop, can you clarify why this shouldnt be `<=`?

contracts/test/PBHVerifierVerify.sol
```diff
@@ -75,4 +69,37 @@ contract PBHVerifierVerify is PBHVerifierTest {
...
+    /// @notice Test that setNumPBHPerMonth works as expected
+    function testSetNumPBHPerMonth() public {
+        MockWorldIDGroups(address(worldID)).setVerifyProofSuccess(true);
+        uint8 month = uint8(BokkyPooBahsDateTimeLibrary.getMonth(block.timestamp));
+        uint16 year = uint16(BokkyPooBahsDateTimeLibrary.getYear(block.timestamp));
+
+        // Value starts at 30, make sure 30 reverts.
+        uint256 pbhExternalNullifier = PBHExternalNullifier.encode(30, month, year);
+        uint256 nullifierHash = 0;
+        vm.expectRevert();
```
0xKitsune: Can we update to expect the specific error?

contracts/test/PBHVerifierVerify.sol
```diff
@@ -75,4 +69,37 @@ contract PBHVerifierVerify is PBHVerifierTest {
...
+        PBHVerifierImpl(address(pbhVerifier)).verifyPbhProof(
+            root, sender, nonce, testCallData, pbhExternalNullifier, nullifierHash, proof
+        );
+
+        // Increase numPbhPerMonth from non owner, expect revert
+        vm.prank(address(123));
+        vm.expectRevert();
```
0xKitsune: Same here, can we update to expect the specific error?

contracts/src/helpers/PBHExternalNullifier.sol
0xForerunner, line 60: because the (0..=29) is equal to 30 entries. the pbhNonces are 0 indexed.

PR #87, opened 2024-12-18
PBH 4337 Module

karankurbur:


0xKitsune (approved):

contracts/src/helpers/PBHExternalNullifier.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.28;
```
Dzejkop, line 2: let's bump in all files?

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,92 @@
...
+import {PackedUserOperation} from "@account-abstraction/contracts/interfaces/PackedUserOperation.sol";
+import {ValidationData} from "@account-abstraction/contracts/core/Helpers.sol";
+import {_packValidationData} from "@account-abstraction/contracts/core/Helpers.sol";
+import {ISafe} from "@4337/interfaces/Safe.sol";
+
+contract PBHSafe4337Module is Safe4337Module {
+    uint256 constant ECDSA_SIGNATURE_LENGTH = 65;
+    uint256 constant TIMESTAMP_BYTES = 12; // 6 bytes each for validAfter and validUntil
+
+    address public immutable PBH_SIGNATURE_AGGREGATOR;
```
Dzejkop, line 14: non-constants [should be lowerCamelCase](https://docs.soliditylang.org/en/latest/style-guide.html#local-and-state-variable-names)

0xKitsune: Immutables aren't specifically mentioned in the style guide but its typical to see immutables in caps case. IMO its a nice way to know that a value will never change after deployment.

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,92 @@
...
+        // operationData is not determined by the signature
+        (bytes memory operationData, uint48 validAfter, uint48 validUntil, bytes calldata signatures) =
+            _getSafeOp(userOp);
+
+        // If it is a PBH transaction, we need to handle two cases with the signature:
+        // 1. The bundler simulates the call with the proof appended
+        // 2. UserOp execution without proof appended
+        if (key == PBH_NONCE_KEY) {
+            isPBH = true;
+        }
```
0xKitsune: This can be `(key == PBH_NONCE_KEY) ? true : false;`

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,92 @@
...
+
+        // Base signature length calculation:
+        // TIMESTAMP_BYTES (12) + (threshold * ECDSA_SIGNATURE_LENGTH)
+        uint256 expectedLength =
+            TIMESTAMP_BYTES + (ISafe(payable(userOp.sender)).getThreshold() * ECDSA_SIGNATURE_LENGTH);
+
+        // If the signature length is greater than the expected length, then we know that the bundler appended the proof
+        // We need to remove the proof from the signature before validation
+        if (key == PBH_NONCE_KEY && userOp.signature.length > expectedLength) {
```
0xKitsune: Instead of checking `key == PBH_NONCE_KEY` here can we use `isPBH`?

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,92 @@
...
+            // TODO: Check for valid proof size
```
0xOsiris: FYI: Signature size with proof is `417 Bytes`, Encoded Proof is `352 Bytes`

karankurbur:
Note: I need to disable `--via-ir` else i get stack trace too deep. Removing this from foundry causes the PBHVerifier tests to fail. Any suggestions?

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,91 @@
...
+        // This does NOT validate the proof
+        // It removes the first 12 bytes from the signature as it represents the validAfter and validUntil values
+        // operationData is not determined by the signature
+        (bytes memory operationData, uint48 validAfter, uint48 validUntil, bytes calldata signatures) =
+            _getSafeOp(userOp);
+
+        // If it is a PBH transaction, we need to handle two cases with the signature:
+        // 1. The bundler simulates the call with the proof appended
+        // 2. UserOp execution without proof appended
+        bool isPBH = (key == PBH_NONCE_KEY) ? true : false;
```
0xKitsune:
```suggestion
        bool isPBH = (key == PBH_NONCE_KEY);
```

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,91 @@
...
+        // Base signature length calculation:
+        // TIMESTAMP_BYTES (12) + (threshold * ECDSA_SIGNATURE_LENGTH)
+        uint256 expectedLength =
+            TIMESTAMP_BYTES + (ISafe(payable(userOp.sender)).getThreshold() * ECDSA_SIGNATURE_LENGTH);
+
+        // If the signature length is greater than the expected length, then we know that the bundler appended the proof
+        // We need to remove the proof from the signature before validation
+        if (isPBH && userOp.signature.length > expectedLength) {
+            // 352 bytes is the size of the encoded proof
+            require(userOp.signature.length - expectedLength == 352, "Invalid proof size in signature");
```
0xKitsune: Can we use a custom error here?

contracts/src/PBH4337Module.sol
```diff
@@ -0,0 +1,91 @@
...
+
+        // If it is a PBH transaction, we need to handle two cases with the signature:
+        // 1. The bundler simulates the call with the proof appended
+        // 2. UserOp execution without proof appended
+        bool isPBH = (key == PBH_NONCE_KEY);
+
+        // Base signature length calculation:
+        // TIMESTAMP_BYTES (12) + (threshold * ECDSA_SIGNATURE_LENGTH)
+        uint256 expectedLength =
+            TIMESTAMP_BYTES + (ISafe(payable(userOp.sender)).getThreshold() * ECDSA_SIGNATURE_LENGTH);
```
0xOsiris, line 58: Will our logic break here if the threshold is > 1? Should we require the threshold is 1?

0xOsiris: Our logic in the `PBHSignatureAggregator` would for sure break, but maybe we should update the signature aggregator to determine the signature length based on the threshold as well. 

Referring to: https://github.com/worldcoin/world-chain/blob/268f6a9f2629a623e1ed2c5e40894bf6cfb10ac2/contracts/src/PBHSignatureAggregator.sol#L62

Let me know what you think

0xOsiris (approved):
Nice work, LGTM

contracts/src/PBH4337Module.sol
karankurbur: i intentionally left the logic here to support more than 1 signer, we should update the aggregator to do the same. 

There are a few projects in the works on the app team to add recovery/passkey signers so we should keep this in mind in general

PR #89, opened 2024-12-21
fix(payload-builder): Fix payload builder tests

0xKitsune:
This PR fixes various payload  builder tests. Note that `test_transaction_pool_ordering()` is still failing.

### DIFF
--- a/world-chain-builder/crates/world/node/tests/e2e.rs
+++ b/world-chain-builder/crates/world/node/tests/e2e.rs
@@ -312,7 +312,7 @@ async fn test_dup_pbh_nonce() -> eyre::Result<()> {
 
     let raw_tx_0 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 0).await;
     ctx.node.rpc.inject_tx(raw_tx_0.clone()).await?;
-    let raw_tx_1 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 1).await;
+    let raw_tx_1 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 0).await;
 
     // Now that the nullifier has successfully been stored in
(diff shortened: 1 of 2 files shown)


world-chain-builder/crates/world/node/tests/e2e.rs
```diff
@@ -312,7 +312,7 @@ async fn test_dup_pbh_nonce() -> eyre::Result<()> {
 
     let raw_tx_0 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 0).await;
     ctx.node.rpc.inject_tx(raw_tx_0.clone()).await?;
-    let raw_tx_1 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 1).await;
+    let raw_tx_1 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 0).await;
```
0xOsiris, line 315:
```suggestion
    let raw_tx_1 = ctx.raw_pbh_tx_bytes(signer.clone(), 0, 1).await;
```

This is the nonce, and should be 1 I believe as it's on the same signer

0xOsiris (approved):
Just one nit, looks good!

world-chain-builder/crates/world/node/tests/e2e.rs
0xKitsune: Yes, the nonce is being set to a nonce that has already been used which causes the second tx to fail when being sent to the node. This should be the expected behavior, lmk if I am overlooking something here.

`assert!(ctx.node.rpc.inject_tx(raw_tx_1.clone()).await.is_err())`

0xOsiris: Ahh my bad, sounds good!

PR #95, opened 2024-12-26
feat(contracts): PBHMulticall

0xKitsune:
This PR introduces the `pbhMulticall()` function which enables arbitrary PBH transactions outside of a 4337 bundle. Tests still being written.


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -107,6 +108,9 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl {
     ///         World ID in a given month.
     uint8 public numPbhPerMonth;
 
+    /// @notice Address of the Multicall3 implementation.
+    address multicall3;
```
0xOsiris: Can we explicitly specify the visibility here as `internal`

0xOsiris (approved):
LGTM nice work, would be good to get a test on `pbhMulticall` in this PR, or another. Also just needs a `forge fmt`

Issue #97, opened 2024-12-27
chore(world-chain-builder): remove primitives crate and any references


0xOsiris:
The `world-chain-builder-primitives` crate is no longer required on the current architecture, and should be removed.

varun-doshi:
I can take this one

PR #98, opened 2024-12-29
feat: Removed world-chain-builder-primitives 

varun-doshi:
Ref https://github.com/worldcoin/world-chain/issues/97


0xKitsune (approved):
Lgtm thanks for the PR!

PR #101, opened 2024-12-31
PBHEntryPoint Uninit Tests

0xForerunner:


contracts/test/TestSetup.sol
```diff
@@ -61,8 +61,8 @@ contract TestSetup is Test {
         vm.deal(address(this), type(uint128).max);
         vm.deal(address(safe), type(uint256).max);
 
-        // Deposit some funds into the Entry Point from the Mock Account.
-        entryPoint.depositTo{value: 10 ether}(address(safe));
+        // TODO: Deposit some funds into the Entry Point from the Mock Account.
+        // entryPoint.depositTo{value: 10 ether}(address(safe));
```
0xOsiris: Can we uncomment this? This is necessary for `testFullFlow` Setup in the Signature aggregation tests.

0xOsiris (approved):
Looks good, just one change!

contracts/test/TestSetup.sol
0xForerunner: It wasn't working when I ran it, unless you fixed it on main. Lemme check.

0xOsiris: You need to fork mainnet when running the tests. `forge test -f $ETHEREUM_PROVIDER`. As we are just wrapping the `entryPoint` around the mainnet address. It's passing in CI on main I believe

PR #104, opened 2024-12-31
chore(pbh-contracts): custom revert messages

0xOsiris:


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -213,7 +213,10 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
...
             bytes32 hashedOps = keccak256(abi.encode(opsPerAggregator[i].userOps));
             assembly ("memory-safe") {
-                if gt(tload(hashedOps), 0) { revert(0, 0) }
+                if gt(tload(hashedOps), 0) {
+                    mstore(0x00, 0x5e75ad06) // StorageCollision()
+                    revert(0x00, 0x04)
+                }
 
                 tstore(hashedOps, hashedOps)
             }
```
0xKitsune, line 223: Can we add the custom errors in the interface or contract? This will make it easier to assert during testing/debugging.

0xKitsune (approved):

PR #109, opened 2025-01-02
Dzejkop/support multicall in validator

Dzejkop:


world-chain-builder/crates/world/pool/src/bindings.rs
```diff
@@ -24,9 +33,22 @@ sol! {
...
+
         function handleAggregatedOps(
             IEntryPoint.UserOpsPerAggregator[] calldata,
             address payable
-        ) public;
+        ) external;
+
+        function pbhMulticall(
+            IMulticall3.Call3[] calls,
+            PBHPayload paylaod,
```
0xKitsune:
```suggestion
            PBHPayload payload,
```

world-chain-builder/crates/world/pool/src/eip4337.rs
```diff
@@ -13,3 +13,9 @@ pub fn hash_user_op(user_op: &PackedUserOperation) -> Field {
 
     hash_to_field(keccak_hash.as_slice())
 }
+
+pub fn hash_pbh_multicall(msg_sender: Address, calls: Vec<IMulticall3::Call3>) -> Field {
+    let keccak_hash = keccak256(SolValue::abi_encode_packed(&(&msg_sender, calls)));
```
0xOsiris, line 18: This hash is not necessarily unique across transactions. I think we should hash the tx nonce as well at minimum right?

0xOsiris (changes requested):
I think the signal hash here needs to be including the tx nonce, to prevent double spend on the proofs

world-chain-builder/crates/world/pool/src/eip4337.rs
0xOsiris: `keccak256(sender, tx nonce)` would be sufficient, could hash the calls as well

world-chain-builder/crates/world/pool/src/bindings.rs
```diff
@@ -24,9 +33,22 @@ sol! {
     }
 
     contract IPBHValidator {
```
0xOsiris: We should update this contract name to `IPBHEntryPoint`

world-chain-builder/crates/world/pool/src/validator.rs
```diff
@@ -215,9 +295,15 @@ where
         mut transaction: Self::Transaction,
     ) -> TransactionValidationOutcome<Self::Transaction> {
         if transaction.to().unwrap_or_default() == self.pbh_validator {
+            // Try and validate a PBH bundle tx
```
0xOsiris, line 298: We need to match the function selector here, otherwise we are never going to hit `validate_pbh_multicall` because of the early return on `validate_pbh_bundle`

world-chain-builder/crates/world/pool/src/eip4337.rs
0xKitsune: Including the tx nonce is not possible here since the tx nonce is not available onchain. The `signalHash` is not provided as an input but is rather derived onchain.

```solidity

    function pbhMulticall(IMulticall3.Call3[] calldata calls, PBHPayload calldata pbhPayload)
        external
        virtual
        onlyInitialized
        onlyProxy
        nonReentrant
        returns (IMulticall3.Result[] memory returnData)
    {
        uint256 signalHash = abi.encode(msg.sender, calls).hashToField();

        verifyPbh(signalHash, pbhPayload);
        nullifierHashes[pbhPayload.nullifierHash] = true;

        returnData = IMulticall3(multicall3).aggregate3(calls);
        emit PBH(msg.sender, pbhPayload);

        return returnData;
    }
```

It would be possible to create proofs ahead of time for a given action where the calls are the same. However users can not double spend nullifier hashes as this is enforced onchain.

0xOsiris: Well then shouldn't we be hashing the `pbhPayload` as the `date - pbhNonce` will guarantee uniqueness across transactions? Or does it not matter that the signal hash can be duplicated across proofs

world-chain-builder/crates/world/pool/src/validator.rs
0xOsiris: Edit: See that we are only returning `Err()` on proof verification

0xOsiris (approved):

0xKitsune (approved):

world-chain-builder/crates/world/pool/src/eip4337.rs
0xKitsune: We can always guarantee uniqueness of the proof provided with the tx because the external nullifier includes the PBH nonce and we enforce that all nullifier hashes are unique.

With the PBH multicall, the signal hash can be duplicated but the nullifier hash is still enforcing that proofs are unique. Note that it is possible to create proofs ahead of time if the calldata being executed does not change. This is possible via 4337 bundles as well, but the smart account nonce must also be encoded in the signal hash ahead of time making it slightly more constrained.

world-chain-builder/crates/world/pool/src/validator.rs
Dzejkop: yeah the flow is a bit confusing but I didn't wanna change it in this PR - I'll make a refactor at some point in the future

Dzejkop:
> I think the signal hash here needs to be including the tx nonce, to prevent double spend on the proofs

This matches the smart contract logic. And we can't access the nonce in the smart contract

PR #113, opened 2025-01-06
PBH multicall gas limit

karankurbur:
Add pbhGasLimit that is checked inside pbhMulticall

### DIFF
--- a/contracts/src/PBHEntryPointImplV1.sol
+++ b/contracts/src/PBHEntryPointImplV1.sol
@@ -290,17 +290,17 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
         nonReentrant
         returns (IMulticall3.Result[] memory returnData)
     {
-        // Check if pbh gas limit is exceeded
-        if (gasleft() > pbhGasLimit) {
-            revert GasLimitExceeded(gasleft());
-        }
-
         uint256 signalHash = abi.encode(msg.sender, calls).hashToField();
         verifyPbh(signalHash, pbhPayload);
         nullifierHashes[pbhPayload.nullifierHash] = true;
 
         returnData = IMulticall3(multicall3).aggregate3{gas: pbhGasLimit}(calls);
         emit PBH(msg.sender, pbhPayload);
+
+        // Check if pbh gas limit is exceeded
+        if (gasleft() > pbhGasLimit) {
+            revert GasLimitExceeded(gasleft());
+        }
 
         return returnData;
     }
--- a/contracts/src/PBHEntryPointImplV1.sol
+++ b/contracts/src/PBHEntryPointImplV1.sol
@@ -75,10 +75,10 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
     error InvalidHashedOps();
 
     /// @notice Thrown when the gas limit for a PBH multicall transaction is exceeded
-    error GasLimitExceeded();
+    error GasLimitExceeded(uint256 gasLimit);
 
     /// @notice Thrown when setting the gas limit for a PBH multicall to 0
-    error InvalidMulticallGasLimit(uint256 gasLimit);
+    error InvalidPBHGasLimit(uint256 gasLimit);
 
     ///////////////////////////////////////////////////////////////////////////////
     ///                                  Events                                ///
@@ -182,7 +182,7 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
         multicall3 = _multicall3;
 
         if (_pbhGasLimit == 0 || _pbhGasLimit > block.gaslimit) {
-            revert InvalidMulticallGasLimit(_pbhGasLimit);
+            revert InvalidPBHGasLimit(_pbhGasLimit);
         }
 
         pbhGasLimit = _pbhGasLimit;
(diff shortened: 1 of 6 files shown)


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -118,6 +128,9 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
     /// @dev Whether a nullifier hash has been used already. Used to guarantee an action is only performed once by a single person
     mapping(uint256 => bool) public nullifierHashes;
 
+    /// @notice The max gas limit for a PBH multicall transaction
+    uint256 public multicallGasLimit;
```
0xKitsune: What do you think about updating this to `pbhGasLimit`? We can have the builder read this value and use this to enforce gas limit per PBH UserOp as well.

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -74,12 +74,22 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
     /// @notice Thrown when the hash of the user operations is invalid
     error InvalidHashedOps();
 
+    /// @notice Thrown when the gas limit for a PBH multicall transaction is exceeded
+    error GasLimitExceeded();
```
0xKitsune: Since the max gas limit can be updated, lets emit the current gas limit with the error.

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -74,12 +74,22 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
...
     ///                                  Events                                ///
     //////////////////////////////////////////////////////////////////////////////
 
     event PBHEntryPointImplInitialized(
-        IWorldID indexed worldId, IEntryPoint indexed entryPoint, uint8 indexed numPbhPerMonth, address multicall3
+        IWorldID indexed worldId,
+        IEntryPoint indexed entryPoint,
+        uint8 indexed numPbhPerMonth,
+        address multicall3,
+        uint256 multicallGasLimit
```
0xKitsune: nit: Since this has been updated to `pbhGasLimit`, can we update this as well?

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -297,4 +322,14 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
         worldId = IWorldID(_worldId);
         emit WorldIdSet(_worldId);
     }
+
+    /// @notice Sets the max gas limit for a PBH multicall transaction.
+    /// @param _pbhGasLimit The max gas limit for a PBH multicall transaction.
+    function setPBHGasLimit(uint256 _pbhGasLimit) external virtual onlyProxy onlyInitialized onlyOwner {
```
0xKitsune, line 343: Lets emit an event here whenever the gas limit is updated.

contracts/src/interfaces/IPBHEntryPoint.sol
```diff
@@ -27,8 +27,13 @@ interface IPBHEntryPoint {
...
-        external;
+    function initialize(
+        IWorldID worldId,
+        IEntryPoint entryPoint,
+        uint8 _numPbhPerMonth,
+        address _multicall3,
+        uint256 _multicallGasLimit
+    ) external;
 
     function validateSignaturesCallback(bytes32 hashedOps) external view;
```
0xKitsune, line 39: Can we also add `function pbhGasLimit() external view;` to the interface?

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -145,7 +146,38 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+            // Extract error selector
+            bytes4 selector = bytes4(err);
+            assertEq(selector, PBHEntryPointImplV1.GasLimitExceeded.selector);
+
+            // Extract value from error data and verify it's non-zero
+            uint256 gasLimit;
+            assembly {
+                gasLimit := mload(add(err, 36)) // 4 bytes selector + 32 bytes offset
+            }
+            assertTrue(gasLimit > 0, "Gas limit should be non-zero");
```
0xKitsune: Can we assert the expected gas limit against `pbhGasLimit()`?

0xOsiris (approved):
LGTM, Can you update the `DeployDevnet.s.sol` script with the new initialization calldata, so this doesn't break CI.

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -264,8 +290,12 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
         nonReentrant
         returns (IMulticall3.Result[] memory returnData)
     {
-        uint256 signalHash = abi.encode(msg.sender, calls).hashToField();
+        // Check if pbh gas limit is exceeded
+        if (gasleft() > pbhGasLimit) {
```
0xOsiris: Would it be better to check this after using up the gas. That way if the actual gas used is < `pbhGasLimit` the transaction will always succeed even if the `gasleft()` is initially > `pbhGasLimit`

Dzejkop: Can we not attach a gas limit to the `aggregate3` call below?

Something like

```
returnData = IMulticall3(multicall3).aggregate3{gas: pbhGasLimit}(calls);
```

0xOsiris: > Can we not attach a gas limit to the aggregate3 call below?

Yeah that's a good approach

PR #115, opened 2025-01-06
Osiris/audit review

0xOsiris:
Addresses Internal Audit Review nits

### DIFF
--- a/contracts/src/PBH4337Module.sol
+++ b/contracts/src/PBH4337Module.sol
@@ -26,12 +26,13 @@ contract PBHSafe4337Module is Safe4337Module {
     /// @notice Thrown when the proof size is invalid.
     error InvalidProofSize();
 
-    /// @notice Thrown when a null address is passed in the constructor.
-    error ZeroAddress();
+    /// @notice Thrown when a null data is passed in the constructor.
+    error ZeroInitializedData();
 
     constructor(address entryPoint, address _pbhSignatureAggregator, uint192 _pbhNonceKey) Safe4337Module(entryPoint) {
-        require(_pbhSignatureAggregator != address(0), ZeroAddress());
-        require(entryPoint != address(0), ZeroAddress());
+        require(_pbhSignatureAggregator != address(0), ZeroInitializedData());
+        require(entryPoint != address(0), ZeroInitializedData());
+        require(_pbhNonceKey != 0, ZeroInitializedData());
         PBH_SIGNATURE_AGGREGATOR = _pbhSignatureAggregator;
         PBH_NONCE_KEY = _pbhNonceKey;
     }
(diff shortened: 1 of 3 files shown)


contracts/src/PBHSignatureAggregator.sol
```diff
@@ -82,7 +87,7 @@ contract PBHSignatureAggregator is IAggregator {
                 userOps[i].signature.length == expectedLength + PROOF_DATA_LENGTH,
                 InvalidSignatureLength(expectedLength + PROOF_DATA_LENGTH, userOps[i].signature.length)
             );
-            bytes memory proofData = userOps[i].signature[expectedLength:];
+            bytes memory proofData = userOps[i].signature[expectedLength:expectedLength + PROOF_DATA_LENGTH];
```
Dzejkop, line 90: Do we have a test for this edge case?

0xOsiris: Yes, we do [here](https://github.com/worldcoin/world-chain/blob/1cbf420f5b23067e3fd6b9c9a4f6d4f1172c184a/contracts/test/PBHSignatureAggregator.t.sol#L131). The require statement above will always revert before we hit this line if the end of the byte array isn't `expectedLength + PROOF_DATA_LENGTH`

0xOsiris: It's identical logic. Thought it was best to be explicit though

contracts/src/PBH4337Module.sol
```diff
@@ -8,21 +8,35 @@ import {_packValidationData} from "@account-abstraction/contracts/core/Helpers.s
...
+
+    /// @notice The PBH Nonce Key.
+    /// @dev This key is used to identify a PBH user operation.
     uint192 public immutable PBH_NONCE_KEY;
 
+    /// @notice Thrown when the proof size is invalid.
     error InvalidProofSize();
 
+    /// @notice Thrown when a null data is passed in the constructor.
+    error ZeroInitializedData();
```
0xKitsune: Nit: Can we update this to `AddressZero()` to be consistent with the `PBHEntryPointImplV1` contract?

contracts/src/PBHSignatureAggregator.sol
```diff
@@ -27,10 +28,14 @@ contract PBHSignatureAggregator is IAggregator {
     /// @notice Thrown when the length of the signature is invalid.
     error InvalidSignatureLength(uint256 expected, uint256 actual);
 
+    /// @notice Thrown when a zero address is passed as the PBHEntryPoint.
+    error ZeroAddress();
```
0xKitsune: Can we either update this to `AddressZero()` or update the `PBHEntryPointImplV1` to use `ZeroAddress()` for consistency?

contracts/src/PBH4337Module.sol
0xOsiris: Do we want another error for a zero `PBH_NONCE_KEY`?

0xKitsune (approved):

PR #116, opened 2025-01-06
minor audit changes

Dzejkop:


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -200,14 +205,10 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuard {
         // Verify the external nullifier
         PBHExternalNullifier.verify(pbhPayload.pbhExternalNullifier, numPbhPerMonth);
 
-        // If worldId address is set, proceed with on chain verification,
-        // otherwise assume verification has been done off chain by the builder.
-        if (address(worldId) != address(0)) {
```
0xOsiris: Why do we want to remove the option to have only off chain verification?

0xKitsune: Agreed, I think it might be useful to keep this. If gas costs for verification become feasible to verify onchain, we can simply set the address.

0xKitsune (approved):

contracts/src/PBHEntryPointImplV1.sol
Dzejkop: sounds good, I'll leave it in

0xKitsune (approved):

PR #118, opened 2025-01-07
fix: Account for Dynamic Data size in 4337 Module signatures

0xOsiris:
**Overview**
- Adds compatibility in the `PBHSignatureAggregator` & `PBH4337Module` for [EIP-1271 Signatures](https://eips.ethereum.org/EIPS/eip-1271).
- Adds Tests with a Mock Signature Validator Contract


contracts/src/PBH4337Module.sol
```diff
@@ -5,6 +5,7 @@ import {Safe4337Module} from "@4337/Safe4337Module.sol";
 import {PackedUserOperation} from "@account-abstraction/contracts/interfaces/PackedUserOperation.sol";
 import {ValidationData} from "@account-abstraction/contracts/core/Helpers.sol";
 import {_packValidationData} from "@account-abstraction/contracts/core/Helpers.sol";
+import {SafeModuleSignatures} from "./helpers/SafeModuleSignatures.sol";
```
0xKitsune, line 8: If this lib consists o one function, maybe we should add it directly to the `PBH4337Module`. Im fine with keeping it as a lib as well if you think its applicable for other contracts though.

contracts/src/helpers/SafeModuleSignatures.sol
```diff
@@ -0,0 +1,37 @@
...
+        internal
+        pure
+        returns (uint256 expectedLength)
+    {
+        expectedLength = 0x41 * threshold;
+        if (signatures.length < expectedLength) {
+            revert InvalidSignatureLength(expectedLength, signatures.length);
+        }
+
+        for (uint256 i = 0; i < threshold; i++) {
```
0xKitsune: Nit: `++i` for very small gas savings. This is very minor but low hanging.

0xKitsune (approved):

PR #125, opened 2025-01-09
Feat/validate user op signature

0xOsiris:
Implements `validateUserOpSignature` on the `PBHSignatureAggregator` along with tests.

### DIFF
--- a/contracts/scripts/DeployDevnet.s.sol
+++ b/contracts/scripts/DeployDevnet.s.sol
@@ -84,7 +84,7 @@ contract DeployDevnet is Script {
     }
 
     function deployPBHSignatureAggregator() public {
-        pbhSignatureAggregator = address(new PBHSignatureAggregator(pbhEntryPoint));
+        pbhSignatureAggregator = address(new PBHSignatureAggregator(pbhEntryPoint, worldIdGroups));
         console.log("PBHSignatureAggregator Deployed at: ", pbhSignatureAggregator);
     }
 
--- a/contracts/scripts/DeployDevnet.s.sol
+++ b/contracts/scripts/DeployDevnet.s.sol
@@ -29,6 +29,8 @@ import {Verifier as DeletionB10} from "@world-id-contracts/verifiers/deletion/b1
 import {Verifier as DeletionB100} from "@world-id-contracts/verifiers/deletion/b100.sol";
 
 contract DeployDevnet is Script {
+    WorldIDIdentityManager worldIDOrb;
+
     address public entryPoint;
     address public worldIdGroups;
     address public pbhEntryPoint;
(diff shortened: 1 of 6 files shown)


contracts/src/helpers/SafeModuleSignatures.sol
```diff
@@ -29,9 +37,35 @@ library SafeModuleSignatures {
...
+
+    /// @notice Utility function to extract the encoded proof data from the signature.
+    /// @param signatures Signature data.
+    /// @param threshold The Signer threshold.
+    /// @return length The length of the signatures.
+    /// @return proofData The encoded proof data.
+    function extractProof(bytes calldata signatures, uint256 threshold)
+        internal
+        pure
+        returns (uint256 length, bytes memory proofData)
```
0xKitsune: Is there a reason that someone would need the length returned? It seems they could calculate this just by `signatures.len() - proofData.len()`? If its useful to have this returned, happy to keep it here!

contracts/src/PBHSignatureAggregator.sol
```diff
@@ -47,26 +48,35 @@ contract PBHSignatureAggregator is IAggregator {
...
      * @param userOp        - The userOperation received from the user.
-     * @return sigForUserOp - The value to put into the signature field of the userOp when calling handleOps.
-     *                        (usually empty, unless account and aggregator support some kind of "multisig".
+     * @return sigForUserOp - The new userOperation signature.
      */
     function validateUserOpSignature(PackedUserOperation calldata userOp)
         external
-        pure
+        view
         returns (bytes memory sigForUserOp)
```
0xKitsune, line 65: Nit: Since this is just the UserOp signature, consider updating to `signature` to keep consistent to 4337 naming conventions.

contracts/scripts/DeployDevnet.s.sol
```diff
@@ -84,7 +84,7 @@ contract DeployDevnet is Script {
     }
 
     function deployPBHSignatureAggregator() public {
-        pbhSignatureAggregator = address(new PBHSignatureAggregator(pbhEntryPoint));
+        pbhSignatureAggregator = address(new PBHSignatureAggregator(pbhEntryPoint, worldIdGroups));
```
0xKitsune: `WorldIdGroups` is for the router contract. Lets use `WorldId` instead to avoid confusion.

0xKitsune (approved):

contracts/src/PBHSignatureAggregator.sol
0xOsiris: This naming is inherited from the `IAggregator` interface

contracts/src/helpers/SafeModuleSignatures.sol
0xOsiris: Yeah well you need to know the length in the SafeModule as it doesn't know if there's a proof attached or not. But you're right in the Aggregator we can just index `signatures[0:signatures.length - PROOF_DATA_LENGTH]`

contracts/src/PBHSignatureAggregator.sol
0xKitsune: Sound good, disregard!

0xKitsune (approved):

PR #127, opened 2025-01-09
Osiris/alchemy review

0xOsiris:


contracts/src/helpers/ByteHasher.sol
```diff
@@ -7,6 +7,6 @@ library ByteHasher {
     /// @return The hash of the specified value
     /// @dev `>> 8` makes sure that the result is included in our field
     function hashToField(bytes memory value) internal pure returns (uint256) {
-        return uint256(keccak256(abi.encodePacked(value))) >> 8;
+        return uint256(keccak256(value)) >> 8;
```
0xKitsune, line 10: This has to be `abi.encodePacked()`

https://docs.world.org/world-id/id/on-chain#hash-to-field-helper-function

We should prefer using the `ByteHasher` lib directly rather than reimplementing it here.

0xKitsune (changes requested):

0xKitsune (approved):

PR #136, opened 2025-01-16
fix(contracts): Fix order of arguments when initializing `ValidationData`

0xKitsune:


0xOsiris (approved):
For clarity these are aliased, so this doesn't affect behavior. The original implementation also has parity to the 4337 module: https://github.com/worldcoin/safe-modules/blob/9abf69ea1df673c1010aeb9bbbc6aa14124ba425/modules/4337/contracts/Safe4337Module.sol#L298

PR #137, opened 2025-01-23
Update PBH Validation

0xKitsune:


world-chain-builder/crates/world/pbh/Cargo.toml
```diff
@@ -13,6 +13,7 @@ semaphore.workspace = true
 strum.workspace = true
 serde.workspace = true
 bon.workspace = true
+eyre.workspace = true
```
0xOsiris: I think this is unused based on the compilation logs

world-chain-builder/crates/world/pbh/src/payload.rs
```diff
@@ -63,14 +80,80 @@ pub struct PbhPayload {
...
+    pub fn validate_external_nullifier(
+        &self,
+        date: chrono::DateTime<chrono::Utc>,
+        pbh_nonce_limit: u8,
+    ) -> Result<(), PbhValidationError> {
+        // In most cases these will be the same value, but at the month boundary
+        // we'll still accept the previous month if the transaction is at most a minute late
+        // or the next month if the transaction is at most a minute early
+        let valid_dates = [
+            DateMarker::from(date - chrono::Duration::minutes(1)),
```
0xOsiris: This relaxed constraint doesn't exist in the contracts. I think we should either remove this, or update the `PBHExternalNullifier` library to reflect the 1 minute relaxation if possible.

world-chain-builder/crates/world/pbh/src/payload.rs
```diff
@@ -63,14 +80,80 @@ pub struct PbhPayload {
...
+            DateMarker::from(date + chrono::Duration::minutes(1)),
+        ];
+        if valid_dates
+            .iter()
+            .all(|d| self.external_nullifier.date_marker() != *d)
+        {
+            return Err(PbhValidationError::InvalidExternalNullifierPeriod);
+        }
+
+        if self.external_nullifier.nonce >= pbh_nonce_limit {
```
0xOsiris, line 126: Not in the scope of this PR, but really the `pbh_nonce_limit` should be read from storage of the `PBHEntryPoint` contract rather than configured through CLI. If we update this value in the contracts we won't have a dynamic update of this value in the builder

world-chain-builder/crates/world/pool/src/validator.rs
```diff
@@ -295,21 +190,27 @@ where
...
 
-        self.inner.validate_one(origin, transaction.clone())
+        match function_signature {
+            IPBHEntryPoint::handleAggregatedOpsCall::SELECTOR => {
+                self.validate_pbh_bundle(origin, transaction)
+            }
+            IPBHEntryPoint::pbhMulticallCall::SELECTOR => {
+                self.validate_pbh_multicall(origin, transaction)
+            }
+            _ => WorldChainPoolTransactionError::InvalidCalldata.to_outcome(transaction),
```
0xOsiris: Won't this not allow us to interact with the Admin functions on the contract? I think we should filter out all selectors that we could interact with, and not mark pbh

0xOsiris (changes requested):
Looks great, just a couple changes!

world-chain-builder/crates/world/pbh/Cargo.toml
0xKitsune: This is used but only for tests, we can move to dev dependencies.

world-chain-builder/crates/world/pbh/src/payload.rs
0xKitsune: Lets remove this

world-chain-builder/crates/world/pool/src/validator.rs
0xKitsune: Good catch, Ill update this.

world-chain-builder/crates/world/pbh/src/payload.rs
0xKitsune: Agreed, Ill update this in a separate PR.

world-chain-builder/crates/world/pool/src/validator.rs
```diff
@@ -295,21 +190,27 @@ where
...
 
-        self.inner.validate_one(origin, transaction.clone())
+        match function_signature {
+            IPBHEntryPoint::handleAggregatedOpsCall::SELECTOR => {
+                self.validate_pbh_bundle(origin, transaction)
+            }
+            IPBHEntryPoint::pbhMulticallCall::SELECTOR => {
+                self.validate_pbh_multicall(origin, transaction)
+            }
+            _ => return self.inner.validate_one(origin, transaction.clone()),
```
0xOsiris, line 212: Ah yeah this is a good approach!

PR #141, opened 2025-01-24
L-04

0xOsiris:
L-04 Misleading Comments

### DIFF
--- a/contracts/src/PBHEntryPointImplV1.sol
+++ b/contracts/src/PBHEntryPointImplV1.sol
@@ -108,6 +108,9 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
     /// @notice Thrown when setting the gas limit for a PBH multicall to 0
     error InvalidPBHGasLimit(uint256 gasLimit);
 
+    /// @notice Thrown when the length of PBHPayloads on the aggregated signature is not equivalent to the amount of UserOperations.
+    error InvalidAggregatedSignature(uint256 payloadsLength, uint256 userOpsLength);
+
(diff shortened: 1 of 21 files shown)


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -30,7 +30,7 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
     /// @dev The EntryPoint where Aggregated PBH Bundles will be proxied to.
     IEntryPoint public entryPoint;
 
-    /// @notice The number of PBH transactions that may be used by a single
+    /// @notice The upper bound on the number of PBH transactions that may be used by a single
     ///         World ID in a given month.
     uint8 public numPbhPerMonth;
```
0xKitsune: Wdyt about adjusting the name to `pbhNonceLimit` and updating the comment to be something like "The number of PBH transactions alloted to each World ID per month, 0 indexed. For example, if the `pbhNonceLimit` is 29, a user can submit 30 PBH txs"

Note that we would also need to change the arg in `PBHEntryPointImplInitialized` and related comments.

contracts/src/PBHSignatureAggregator.sol
```diff
@@ -32,13 +32,6 @@ contract PBHSignatureAggregator is IAggregator {
     ///                                  ERRORS                                ///
     //////////////////////////////////////////////////////////////////////////////
```
0xKitsune: Looks like these changes are related to `N-06`. Can we remove these changes in favor of [#145?](https://github.com/worldcoin/world-chain/pull/145)

Dzejkop (approved):

PR #144, opened 2025-01-24
N-03

0xOsiris:
N-03 Code Simplifications

ACK
> The handleAggregatedOps function [calls verifyPbh](https://github.com/worldcoin/world-chain/blob/cc112c40157eca93e998649459340f4c5f3580c0/contracts/src/PBHEntryPointImplV1.sol#L222) which redundantly invokes the [onlyProxy and onlyInitialized modifiers](https://github.com/worldcoin/world-chain/blob/cc112c40157eca93e998649459340f4c5f3580c0/contracts/src/PBHEntryPointImplV1.sol#L175-L176). The function body could be moved to its own internal function so that it can be invoked without the modifiers.

I believe we want this to be public for use by the bundler off chain. cc @0xKitsune

### DIFF
--- a/contracts/src/PBH4337Module.sol
+++ b/contracts/src/PBH4337Module.sol
@@ -106,7 +106,7 @@ contract PBHSafe4337Module is Safe4337Module {
         // A malicious bundler can pad the Safe operation `signatures` with additional bytes, causing the account to pay
         // more gas than needed for user operation validation (capped by `verificationGasLimit`).
         // `_checkSignaturesLength` ensures that there are no additional bytes in the `signature` than are required.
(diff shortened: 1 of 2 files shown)


0xKitsune:
@0xOsiris I believe this is saying that we can create an additional, internal `_verifyPbh()` function that is called within the context of `handleAggregatedOps()` to avoid unncessary gas overhead of checking `onlyProxy` and `onlyInitialized` twice.

```solidity

function handleAggregatedOps() onlyProxy onlyInitialized {
    // --snip--
    _verifyPbh();
    // --snip--


}

function verifyPbh() onlyProxy onlyInitialized {
     _verifyPbh();
}

function _verifyPbh() internal {
    // --snip--
}
```

Dzejkop (approved):

PR #154, opened 2025-02-03
World ID Sponsoring Paymaster

karankurbur:


contracts/src/WorldIdPaymster.sol
```diff
@@ -0,0 +1,96 @@
...
+    ) internal view override returns (bytes memory context, uint256 validationData) {
+        // Check if sender is verified in address book
+        uint256 verifiedUntil = addressBook.addressVerifiedUntil(userOp.sender);
+        require(
+            verifiedUntil > block.timestamp,
+            "AddressBookPaymaster: Sender not verified"
+        );
+
+        uint256 today = getDate();
+        uint256 txCount = userDailyTxs[userOp.sender][today];
```
sevazhidkov, line 58: maybe total gas is actually a better metric

sevazhidkov: otherwise it can be abused by huge transactions

PR #155, opened 2025-02-03
Osiris/better validation

0xOsiris:
Related to `6.1 [Medium] Nullifier hashes may be consumed without a valid proof`
Introduces a Revm inspector to inspect the call trace of a transaction in the transaction validator. If any target address along the call stack is the `PBHEntryPoint` the transaction is considered "Invalid" and dropped from the Mempool. 

This protects the `PBHEntryPoint`s `nullifierHashes` mapping from ever being populated without proof verification.

### DIFF
--- a/world-chain-builder/crates/world/pool/src/inspector.rs
+++ b/world-chain-builder/crates/world/pool/src/inspector.rs
@@ -1,0 +1,34 @@
+use reth::revm::{Database, Inspector};
+use revm_primitives::Address;
+
+/// Simple inspector that keeps track of the call stack
+pub struct CallTracer {
+    /// A vector of addresses across the call stack.
+    pub stack: Vec<Address>,
+}
+
+impl CallTracer {
+    /// Creates a new instance [`CallTracer`]
+    pub fn new() -> Self {
+        Self { stack: Vec::new() }
+    }
+
+    /// Checks whether the `pbh_entrypoint` exists within the call stack.
+    /// 
+    /// Note: This is excluding the target of the transaction from an EOA as the first external call interpreted will be
+    ///       within the call context of a contract. 
+    pub fn is_valid(&self, pbh_entrypoint: Address) -> bool {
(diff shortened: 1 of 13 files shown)


world-chain-builder/crates/world/pool/src/inspector.rs
```diff
@@ -0,0 +1,32 @@
...
+/// Simple inspector that keeps track of the call stack
+pub struct CallTracer {
+    /// A vector of addresses across the call stack.
+    pub stack: Vec<Address>,
+}
+
+impl CallTracer {
+    /// Creates a new instance [`CallTracer`]
+    pub fn new() -> Self {
+        Self { stack: Vec::new() }
```
0xForerunner, line 13: Have you considered using a hash set for this?

0xForerunner (changes requested):

world-chain-builder/crates/world/pool/src/inspector.rs
0xOsiris: Well there's a few different errors with this PR, the original idea was `PBHEntryPoint` address should not exist on the call stack in anywhere except the first index. But this will not be true since the `PBHSignatureAggregator` calls back into it. 

We could use a `HashSet` here, and maintain a counter on the amount of calls to differentiate the first call. Then check the caller before inserting the target onto the stack.

0xOsiris: This logic also eliminates the validity of a contract calling into the `PBHEntryPoint` which is actually a very large downside imo

0xForerunner: > and maintain a counter

Yeah that's a decent way of doing it! I'll let you decide. TBH I think a vec probably works just fine as well.

0xForerunner (approved):

0xOsiris:
Closing in favor of #158

PR #157, opened 2025-02-04
Osiris/contract deployment

0xOsiris:


contracts/scripts/Deploy.s.sol
```diff
@@ -10,13 +10,12 @@ import {IWorldID} from "@world-id-contracts/interfaces/IWorldID.sol";
 import {IPBHEntryPoint} from "../src/interfaces/IPBHEntryPoint.sol";
 import {IEntryPoint} from "@account-abstraction/contracts/interfaces/IEntryPoint.sol";
 
-contract DeployDevnet is Script {
+contract Deploy is Script {
```
0xKitsune, line 13: Can we specify the network here? `WORLD_ID` is not deterministic and will need to change between networks.

contracts/scripts/Deploy.s.sol
```diff
@@ -10,13 +10,12 @@ import {IWorldID} from "@world-id-contracts/interfaces/IWorldID.sol";
...
     address public pbhEntryPoint;
     address public pbhEntryPointImpl;
     address public pbhSignatureAggregator;
 
-    address internal constant WORLD_ID = address(0);
-    address internal constant MULTICALL3_ADDRESS = 0xcA11bde05977b3631167028862bE2a173976CA11;
+    address internal constant WORLD_ID = 0x047eE5313F98E26Cc8177fA38877cB36292D2364;
```
0xKitsune, line 18: Do we need this here if we are not using it?

0xOsiris: We are using it

0xOsiris: For the `PBHSignatureAggregator`

PR #158, opened 2025-02-05
feat(payload-builder): Implement `PBHCallTracer` 

0xKitsune:
Replaces #155, which addresses the same issue with a revised approach.

Related to `6.1 [Medium] Nullifier hashes may be consumed without a valid proof`
Introduces a Revm inspector to inspect the call trace of a transaction in the transaction validator. If any target address along the call stack is the PBHEntryPoint the transaction is considered "Invalid" and dropped from the Mempool.

This protects the PBHEntryPoints nullifierHashes mapping from ever being populated without proof verification.


world-chain-builder/crates/world/payload/src/inspector.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl<DB: Database> Inspector<DB> for PBHCallTracer {
+    fn call(
+        &mut self,
+        context: &mut reth::revm::EvmContext<DB>,
+        inputs: &mut reth::revm::interpreter::CallInputs,
+    ) -> Option<reth::revm::interpreter::CallOutcome> {
+        // Check if the target address is the PBHEntryPoint. If the caller is not the tx origin, mark the tx as invalid.
+        if inputs.target_address == self.pbh_entry_point {
+            if inputs.caller != context.env.tx.caller {
```
0xOsiris: This will trigger on the callback from the `PBHSignatureAggregator`

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -827,20 +859,15 @@ where
                 Err(err) => {
                     match err {
                         EVMError::Transaction(err) => {
-                            if matches!(err, InvalidTransaction::NonceTooLow { .. }) {
```
0xOsiris: We should follow `reth`s implementation on the `NonceTooLow` case https://github.com/paradigmxyz/reth/blob/8f59efb96a67240f5fe76956cfadf2209393f725/crates/optimism/payload/src/builder.rs#L987

0xOsiris (approved):

0xOsiris (approved):
Looks good, nice work!

PR #165, opened 2025-02-12
feat(pbh): Implement `WorldChainTxEnvelope` enabling offchain PBH

0xKitsune:
The Current PBH design allows a user to submit a tx with a World ID proof encoded in the calldata. Encoding the proof in calldata allows transactions to be sent to any execution client without the need to support a new transaction envelope. The tx will be eventually be picked up by the builder, which will decode/validate the proof and include the tx with priority when building a block. While encoding the proof in calldata maximizes out of the box compatibility and enables peering of PBH  transactions, it comes at the cost of additional onchain data.

This PR introduces the `WorldChainTxEnvelope`, a new EIP-2718 TxEnvelope containing a `PBHSidecar`, allowing users to submit a PBH payload directly to the builder, reducing the calldata necessary for PBH.


world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -838,6 +859,8 @@ where
             info.executed_transactions.push(tx.into_tx());
         }
 
+        // TODO: backrun the block and post the nullifier hashes contained in the PBH Sidecars
```
0xOsiris, line 862: I think we would need to front run the block with the first tx filling all nullifier hashes consumed from all sidecars in the payload. Backrunning the block will leave any on chain PBH transactions in the current payload unaware of the state changes from the sidecars.

world-chain-builder/crates/world/pool/src/ordering.rs
```diff
@@ -35,7 +35,7 @@ where
         let effective_tip_per_gas = transaction.effective_tip_per_gas(base_fee).map(U256::from);
 
         Some(WorldChainPriority {
-            is_pbh: transaction.valid_pbh(),
+            is_pbh: transaction.pbh_sidecar().is_some(),
```
0xOsiris, line 38: This will return `None` for on Chain PBH transactions, and `is_pbh` will always be marked `false`

world-chain-builder/crates/world/pool/src/validator.rs
```diff
@@ -115,14 +116,16 @@ where
...
+            pbh_bundle.extend(pbh_payloads);
         }
 
         if let TransactionValidationOutcome::Valid {
             transaction: ValidTransaction::Valid(tx),
             ..
         } = &mut tx_outcome
         {
-            tx.set_valid_pbh();
+            tx.set_pbh_sidecar(PBHSidecar::PBHBundle(pbh_bundle));
```
0xOsiris, line 188: Ah I see, disregard the comment above

world-chain-builder/crates/world/payload/src/builder.rs
0xOsiris: Or is the idea for the builder to be the only entity with write access to the `nullifierHashes` mapping, and then back run every block with insertions (for on or off chain PBH transactions).

world-chain-builder/crates/world/rpc/src/api_ext.rs
```diff
@@ -79,14 +251,16 @@ where
     }
 
     async fn send_raw_transaction(&self, tx: Bytes) -> Result<B256, Self::Error> {
-        let recovered = recover_raw_transaction(&tx)?;
-        let pool_transaction: WorldChainPooledTransaction =
-            OpPooledTransaction::from_pooled(recovered).into();
+        let mut data: &[u8] = &tx;
+        let tx_envelope = WorldChainTxEnvelope::decode_2718(&mut data)
+            .map_err(|_| EthApiError::FailedToDecodeSignedTransaction)?;
+        let pooled_tx = WorldChainPooledTransaction::try_from(tx_envelope)
```
0xOsiris, line 257: Why not just decode the `tx` into `WorldChainPooledTransaction` instead of decoding into the `WorldChainTxEnvelope` first?

PR #171, opened 2025-02-13
Osiris/4337 patches

0xOsiris:
Changes Requested for ease of interaction with the `PBHSignatureAggregator` from the bundler. 

Adding the `signalHash` to the revert data allows the bundler to map the revert reason to specific user operations.


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -90,7 +90,8 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
     //////////////////////////////////////////////////////////////////////////////
 
     /// @notice Thrown when attempting to reuse a nullifier
-    error InvalidNullifier();
+    /// @param signalHash The signal hash associated with the PBH payload.
+    error InvalidNullifier(uint256 signalHash);
```
0xKitsune: Should we also include `externalNullifier` here?

contracts/src/PBHSignatureAggregator.sol
```diff
@@ -79,9 +79,12 @@ contract PBHSignatureAggregator is IAggregator {
         // We now generate the signal hash from the sender, nonce, and calldata
         uint256 signalHash = abi.encodePacked(userOp.sender, userOp.nonce, userOp.callData).hashToField();
 
-        worldID.verifyProof(
-            pbhPayload.root, signalHash, pbhPayload.nullifierHash, pbhPayload.pbhExternalNullifier, pbhPayload.proof
-        );
+        pbhEntryPoint.verifyPbh(signalHash, pbhPayload);
+        if (address(pbhEntryPoint.worldId()) == address(0)) {
```
0xKitsune, line 85: Can we add a comment to explain why we verify the proof here if we do not verify the proof in the pbh entry point?

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -193,11 +199,14 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
...
-            revert InvalidNullifier();
+            revert InvalidNullifier(signalHash);
         }
 
         // Verify the external nullifier
-        PBHExternalNullifier.verify(pbhPayload.pbhExternalNullifier, numPbhPerMonth);
+        try PBHExternalNullifier.verify(pbhPayload.pbhExternalNullifier, numPbhPerMonth) {}
+        catch (bytes memory reason) {
+            revert InvalidExternalNullifier(signalHash, reason);
+        }
```
0xKitsune: Instead of try catching here, would it be possible to simply return the `signalHash` and `externalNullifier` with each error returned from the `verify` function? 


```solidity
 function verify(uint256 externalNullifier, uint8 numPbhPerMonth, uint256 signalHash) public view {
        require(externalNullifier <= type(uint40).max, InvalidExternalNullifierLeadingZeros(externalNullifier, signalHash));
        (uint8 version, uint8 pbhNonce, uint8 month, uint16 year) = PBHExternalNullifier.decode(externalNullifier);
        require(version == V1, InvalidExternalNullifierVersion(externalNullifier, signalHash));
        require(year == BokkyPooBahsDateTimeLibrary.getYear(block.timestamp), InvalidExternalNullifierYear(externalNullifier, signalHash));
        require(month == BokkyPooBahsDateTimeLibrary.getMonth(block.timestamp), InvalidExternalNullifierMonth(externalNullifier, signalHash));
        require(pbhNonce < numPbhPerMonth, InvalidPbhNonce(externalNullifier, signalHash));
    }

```

Alternatively, we could have a single `InvalidExternalNullifier()` error within the `PBHExtgernalNullifier` lib that contains a `reason` enum or string.

We should also update `PBHExternalNullifier.verify` to be internal to avoid an external call.

contracts/src/PBHSignatureAggregator.sol
0xOsiris: Yeah, sounds good. Reason is if `address(pbhEntryPoint.worldId()) != address(0)` then we would be duping the semaphore verification from `pbhEntryPoint.verifyPbh(signalHash, pbhPayload);`

contracts/src/PBHEntryPointImplV1.sol
0xOsiris: Yeah that SGTM. But we still want a unified `InvalidExternalNullifier` error so parsing the revert selector in the bundler is much simpler. We can just do this by passing a string message in addition to `signalHash`, and `externalNullifier`

contracts/src/PBHSignatureAggregator.sol
0xKitsune: Yeah makes sense, just would be helpful to add a comment for other devs to know how this works. Primarily, we should add documentation as to why we do onchain proof aggregation here if we do not verify onchain within the entrypoint.

contracts/src/PBHEntryPointImplV1.sol
0xKitsune: Yeah a unified `InvalidExternalNullifier` sgtm, we can either pass a string as the reason or use an enum. An enum would be nice so that Rundler can use pattern matching since the `sol!()` macro from `alloy` will translate the solidity types into Rust types.

0xKitsune (approved):

PR #175, opened 2025-02-20
chore(rundler): update rundler to v0.6.0-alpha.3

dancoombs:
- Update Rundler to v0.6.0-alpha.3
- This changed one of the config files to "builder config file"
- I felt this was confusing naming as the worldchain builder is also in the setup - so I added a `rundler_` qualifier in a few places where it wasn't obviously a rundler file.


0xOsiris (approved):
Thanks for the PR!

PR #180, opened 2025-02-25
feat: switch pbh nonce to u16

0xOsiris:


0xForerunner (approved):

world-chain-builder/crates/world/node/src/node.rs
```diff
@@ -243,7 +243,7 @@ where
                 WorldChainTransactionValidator::new(
                     op_tx_validator,
                     root_validator,
-                    num_pbh_txs,
+                    num_pbh_txs as u16,
```
0xKitsune: This should be a `u16` already, we should update the field's type from `u8` to `u16`.

world-chain-builder/crates/world/node/src/test_utils.rs
```diff
@@ -70,7 +70,7 @@ pub async fn raw_pbh_bundle_bytes(
     let month = dt.month() as u8;
     let year = dt.year() as u16;
 
-    let ext_nullifier = ExternalNullifier::v1(month, year, pbh_nonce);
+    let ext_nullifier = ExternalNullifier::v1(month, year, pbh_nonce as u16);
```
0xKitsune: Lets update the arg to be a `u16`.

world-chain-builder/crates/world/pbh/src/external_nullifier.rs
```diff
@@ -24,25 +24,25 @@ pub struct ExternalNullifier {
...
+    #[builder(default = 0)]
+    pub nonce: u16,
 }
 
 /// The encoding format is as follows:
-///      - Bits:40-255: Empty
-///      - Bits 32-39: Year
-///      - Bits 16-31: Month
-///      - Bits 8-15: Nonce
+///      - Bits:48-263: Empty
```
0xKitsune: There are only 256 bits available.

0xKitsune (approved):

PR #183, opened 2025-02-27
feat: `WorldChainBlockRegistry` 

0xKitsune:
This PR introduces the `WorldChainBlockRegistry` which enables the World Chain builder to record which blocks were produced by the builder. This allows third parties to have visibility into which blocks should be constructed with the PBH ordering policy. 

```solidity
contract WorldChainBlockRegistry is Ownable2Step {
   // --snip--

    /// @notice Record the current block as being built by the World Chain Builder
    function stampBlock() public onlyBuilder {
        require(builtBlocks[block.number] == address(0), BlockAlreadyRegistered());
        builtBlocks[block.number] = msg.sender;
        emit BuiltBlock(msg.sender, block.number);
    }

```

### DIFF
--- a/contracts/src/WorldChainBlockRegistry.sol
+++ b/contracts/src/WorldChainBlockRegistry.sol
@@ -25,7 +25,7 @@ contract WorldChainBlockRegistry is Ownable2Step {
     ///                                  ERRORS                                ///
     //////////////////////////////////////////////////////////////////////////////
     error AddressZero();
-    error BlockAlreadyStamped();
+    error BlockAlreadyRegistered();
     error Unauthorized();
 
     ///////////////////////////////////////////////////////////////////////////////
@@ -45,7 +45,7 @@ contract WorldChainBlockRegistry is Ownable2Step {
     }
 
     function stampBlock() public onlyBuilder {
-        require(builtBlocks[block.number] == address(0), BlockAlreadyStamped());
+        require(builtBlocks[block.number] == address(0), BlockAlreadyRegistered());
         builtBlocks[block.number] = msg.sender;
         emit BuiltBlock(msg.sender, block.number);
     }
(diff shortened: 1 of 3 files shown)


contracts/lib/account-abstraction
0xOsiris, line 1: Can we revert this

contracts/lib/forge-std
0xOsiris, line 1: Let's revert this

contracts/lib/openzeppelin-contracts-upgradeable
0xOsiris, line 1: Let's revert this as well

contracts/lib/safe-contracts
0xOsiris, line 1: Let's revert this as well

contracts/lib/safe-modules
0xOsiris, line 1: Let's revert this as well

contracts/src/WorldChainBlockRegistry.sol
```diff
@@ -0,0 +1,67 @@
...
+/// @notice This contract records which blocks were built by the World Chain Builder.
+/// @dev Each block, the builder will insert a transaction calling stampBlock() to indiciate which
+/// blocks should enforce PBH priority ordering.
+contract WorldChainBlockRegistry is Ownable2Step {
+    ///////////////////////////////////////////////////////////////////////////////
+    ///                             STATE VARIABLES                             ///
+    //////////////////////////////////////////////////////////////////////////////
+
+    /// @notice Address of the World Chain Builder
+    address worldChainBuilder;
```
0xOsiris, line 16: We may consider making this 
```solidity
address[] authorizedBuilders
``` 
or 

```solidity
mapping(address builder => bool authorized) authorizedBuilders
``` 
this way we can easily extend to multiple authorized parties

PR #188, opened 2025-03-07
Osiris/reth v1.2.2

0xOsiris:
Bump Reth to `v1.2.2`


Dzejkop (approved):

world-chain-builder/crates/world/node/tests/mod.rs
```diff
@@ -174,7 +174,7 @@ async fn test_can_build_pbh_payload() -> eyre::Result<()> {
 
     assert_eq!(
         payload.block().body().transactions.len(),
-        pbh_tx_hashes.len()
+        pbh_tx_hashes.len() + 1
```
Dzejkop, line 177: that's the block stamping tx, right?

0xOsiris: Yes, these tests are failing on main now

PR #190, opened 2025-03-12
feat: bump reth 1.3.0

klkvr:
Reth v1.3.0 introduces new EVM API and involves breaking changes to `OpPayloadBuilder` and EVMs so we've decided to use world-chain-builder to highlight unexpected breakages and share the updated version with you.

The main change is that payload builder now operates on a `BlockBuilder` trait which is exposed directly from `EvmConfig` and re-uses the EVM.

This implies that payload builder needs to re-use the same EVM during entire duration of block building, so with this PR `PBHCallTracer` is now always configured, but disabled by default via a new `active` flag which is only flipped when executing transactions from the pool.

`BlockBuilder` also encapsulates the logic for constructing headers so this part was removed from the builder and is now simply inherited from OP implementation.

### DIFF
--- a/world-chain-builder/Cargo.lock
+++ b/world-chain-builder/Cargo.lock
@@ -14316,6 +14316,7 @@ dependencies = [
  "op-alloy-consensus",
  "op-alloy-network",
  "op-alloy-rpc-types",
+ "op-revm",
  "reth",
  "reth-basic-payload-builder",
  "reth-chain-state",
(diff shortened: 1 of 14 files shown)


klkvr:
Updated the tests as well, ci should pass now

0xOsiris (approved):
LGTM, thanks for the PR! 

cc @0xKitsune

0xKitsune (approved):
Looks great, thanks for the PR!

PR #192, opened 2025-03-17
support compressed proofs in backwards compat way

Dzejkop:


world-chain-builder/crates/world/pbh/src/payload.rs
```diff
@@ -97,12 +97,25 @@ impl PBHPayload {
         let date = chrono::Utc::now();
         self.validate_external_nullifier(date, pbh_nonce_limit)?;
 
+        let flat = self.proof.0.flatten();
+        let proof =
+            if flat[4].is_zero() && flat[5].is_zero() && flat[6].is_zero() && flat[7].is_zero() {
```
piohei: You can extract this if statement to some method like `is_compressed` or similar. :)

piohei (approved):

world-chain-builder/crates/world/pbh/src/payload.rs
Dzejkop: good point, done!

0xOsiris (approved):

PR #194, opened 2025-03-18
feat: cleanup contracts + builder patches

0xOsiris:
**Overview**:
- Removes the `pbhMulticall` from the `PBHEntryPoint`
- Removes spending nullifier hashes from the contract flow
- Aggregates nullifier hashes in the payload builder, and back runs the block to write the hashes to the `PBHEntryPoint`
- Removes `PBHCallTracer`
- Allows for deterministic deployments of the PBHEntryPoint and PBHSignatureAggregator across Sepolia & Mainnet
- Simplifies (and makes Implementation more gas efficient)


contracts/scripts/Deploy.s.sol
```diff
@@ -19,6 +19,7 @@ contract Deploy is Script {
     address internal constant ENTRY_POINT = 0x0000000071727De22E5E9d8BAf0edAc6f37da032;
     uint256 internal constant MAX_PBH_GAS_LIMIT = 15000000; // 15M
     uint16 internal constant PBH_NONCE_LIMIT = type(uint16).max;
+    address[] internal authorizedBuilders = [address(0)]; // TODO:
```
0xKitsune: Can we update this to address the TODO?

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -139,31 +160,40 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
...
 
         if (_numPbhPerMonth == 0) {
             revert InvalidNumPbhPerMonth();
         }
 
+        if (_authorizedBuilders.length == 0) {
+            revert InvalidAuthorizedBuilders();
+        }
+
+        for (uint256 i = 0; i < _authorizedBuilders.length; i++) {
```
0xKitsune: nit: ++i for good measure

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -309,4 +318,36 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
...
     }
+
+    /// @notice Adds a builder to the list of authorized builders.
+    /// @param builder The address of the builder to authorize.
+    function addBuilder(address builder) external virtual onlyProxy onlyInitialized onlyOwner {
+        if (builder == address(0)) {
+            revert AddressZero();
+        }
+
+        authorizedBuilder[builder] = true;
```
0xKitsune, line 338: Can we add an event when a builder is added?

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -309,4 +318,36 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
...
+    }
+
+    /// @notice Removes a builder from the list of authorized builders.
+    /// @param builder The address of the builder to deauthorize.
+    function removeBuilder(address builder) external virtual onlyProxy onlyInitialized onlyOwner {
+        delete authorizedBuilder[builder];
```
0xKitsune, line 345: Can we add an event for when builders are removed as well?

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -277,5 +239,53 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+        vm.expectRevert("Ownable: caller is not the owner");
+        pbhEntryPoint.removeBuilder(addr);
+    }
+
+    function test_spendNullifierHashes() public {
+        uint256[] memory nullfierHashes = new uint256[](3);
+        nullfierHashes[0] = uint256(0);
+        nullfierHashes[1] = uint256(1);
+        nullfierHashes[2] = uint256(1);
+        vm.prank(BLOCK_BUILDER);
```
0xKitsune, line 277: Can we update this to check event emission?

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -277,5 +239,53 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+        vm.prank(addr);
+        vm.expectRevert("Ownable: caller is not the owner");
+        pbhEntryPoint.removeBuilder(addr);
+    }
+
+    function test_spendNullifierHashes() public {
+        uint256[] memory nullfierHashes = new uint256[](3);
+        nullfierHashes[0] = uint256(0);
+        nullfierHashes[1] = uint256(1);
+        nullfierHashes[2] = uint256(1);
```
0xKitsune: Is this purposefully duplicated or is this meant to be `uint256(2)`?

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -277,5 +239,53 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+
+    function test_removeBuilder_RevertIf_NotOwner(address addr) public {
+        vm.assume(addr != OWNER);
+        vm.prank(addr);
+        vm.expectRevert("Ownable: caller is not the owner");
+        pbhEntryPoint.removeBuilder(addr);
+    }
+
+    function test_spendNullifierHashes() public {
+        uint256[] memory nullfierHashes = new uint256[](3);
```
0xKitsune: Consider fuzzing this input instead

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -277,5 +239,53 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+
+    function test_spendNullifierHashes_RevertIf_NotBlockBuilder(address builder) public {
+        uint256[] memory nullfierHashes = new uint256[](3);
+        nullfierHashes[0] = uint256(0);
+        nullfierHashes[1] = uint256(1);
+        nullfierHashes[2] = uint256(1);
+        vm.assume(builder != BLOCK_BUILDER);
+        vm.prank(builder);
+        vm.expectRevert(PBHEntryPointImplV1.UnauthorizedBuilder.selector);
+        pbhEntryPoint.spendNullifierHashes(nullfierHashes);
```
0xKitsune: Can we assert state has not changed after this?

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -583,19 +587,14 @@ where
...
-
         let mut invalid_txs = vec![];
         let verified_gas_limit = (self.verified_blockspace_capacity as u64 * block_gas_limit) / 100;
 
+        let mut spent_nullifier_hashes = HashSet::new();
+
         // Subtract stamp_block_tx gas limit from the overall block gas limit.
         // This gas will be saved to execute stampBlock at the end of the block.
-        block_gas_limit -= stamp_block_tx.gas_limit();
+        block_gas_limit -= SPEND_NULLIFIER_HASHES_GAS;
```
0xKitsune: `SPEND_NULLIFIER_HASHES_GAS` will need to be a dynamic value since there will be `n` `SSTORE`s depending on how many PBH transactions are in the block.

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -616,11 +615,23 @@ where
...
-                continue;
+            if pooled_tx.valid_pbh() {
+                if info.cumulative_gas_used + tx.gas_limit() > verified_gas_limit {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    continue;
+                }
+
+                let Some(payloads) = pooled_tx.payload() else {
+                    unimplemented!()
+                };
```
0xKitsune: It looks like there will never be a case where `pooled_tx.valid_pbh()` will be true but `pooled_tx.payload()` will be `None`.

Since the implementation of `valid_pbh()` is:
```rust
  fn valid_pbh(&self) -> bool {
        self.payload.is_some()
    }

```

I think we can consolidate this to `if let Some(payload) = pooled_tx.pbh_payload()` to check if the tx is PBH.

world-chain-builder/crates/world/payload/src/stamp.rs
```diff
@@ -1,58 +1,60 @@
...
+use reth_optimism_primitives::OpTransactionSigned;
+use reth_primitives::Recovered;
+use reth_primitives_traits::SignedTransaction;
+use revm::Inspector;
+use semaphore_rs::Field;
+use WorldChainBlockRegistry::spendNullifierHashesCall;
 
 sol! {
     #[sol(rpc)]
     interface WorldChainBlockRegistry {
```
0xKitsune: Can we update this interface name since the block registry will not be used anymore?

world-chain-builder/crates/world/payload/src/stamp.rs
```diff
@@ -1,58 +1,60 @@
...
-        function stampBlock();
+        function spendNullifierHashes(uint256[] calldata _nullifierHashes) external;
     }
 }
 
-pub fn stamp_block_tx<DB, Client>(
+pub const SPEND_NULLIFIER_HASHES_GAS: u64 = 100_000;
+
+pub fn spend_nullifiers_tx<DB, I, Client>(
```
0xKitsune: It might make sense to move this and relevant logic into `builder.rs`, wdyt?

world-chain-builder/crates/world/pool/src/bindings.rs
```diff
@@ -51,11 +51,6 @@ sol! {
             IEntryPoint.UserOpsPerAggregator[] calldata,
             address payable
         ) external;
```
0xKitsune, line 53: Can we add `spendNullifierHashes` here?

world-chain-builder/crates/world/pool/src/tx.rs
```diff
@@ -14,31 +14,36 @@ use reth_primitives::{kzg::KzgSettings, Recovered};
...
-    pub valid_pbh: bool,
+    pub payload: Option<Vec<PBHPayload>>,
 }
 
 pub trait WorldChainPoolTransaction: EthPoolTransaction {
     fn valid_pbh(&self) -> bool;
-    fn set_valid_pbh(&mut self);
+    fn set_pbh_payloads(&mut self, payload: Vec<PBHPayload>);
     fn conditional_options(&self) -> Option<&TransactionConditional>;
+    fn payload(&self) -> Option<&Vec<PBHPayload>>;
```
0xKitsune: Wdyt about updating to `pbh_payload`? Within the builder logic, this makes it a little more clear that `tx.pbh_payload()` is checking for the PBH data compared to `tx.payload()`.

contracts/scripts/Deploy.s.sol
0xOsiris: Yes, just wasn't sure which wallet we wanted to use for the builder. Do you have one in mind?

0xKitsune: Can we just pass this as an arg?

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -616,11 +617,26 @@ where
...
+                if info.cumulative_gas_used + tx.gas_limit() > verified_gas_limit {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    continue;
+                }
+
+                if payloads
+                    .iter()
+                    .any(|payload| spent_nullifier_hashes.contains(&payload.nullifier_hash))
+                {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
```
0xKitsune: Do we need to mark invalid and add the tx to `invalid_txs` or can we simply add to `invalid_txs` since these transactions will be dropped from the mempool after payload building?

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -616,11 +617,26 @@ where
...
+                    invalid_txs.push(*pooled_tx.hash());
+                    continue;
+                }
+
+                block_gas_limit -= COLD_SSTORE_GAS * payloads.len() as u64;
+                if !applied_fixed_gas {
+                    block_gas_limit -= FIXED_GAS;
+                    applied_fixed_gas = true;
+                }
```
0xKitsune: Can be simplified to something like the following

```rust
if spent_nullifier_hashes.is_empty() { 
    block_gas_limit -= FIXED_GAS
}

block_gas_limit -= COLD_SSTORE_GAS * payloads.len() as u64;
           
```

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -616,11 +617,26 @@ where
...
+                if payloads
+                    .iter()
+                    .any(|payload| spent_nullifier_hashes.contains(&payload.nullifier_hash))
+                {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    invalid_txs.push(*pooled_tx.hash());
+                    continue;
+                }
+
+                block_gas_limit -= COLD_SSTORE_GAS * payloads.len() as u64;
```
0xKitsune: Should this be applied after execution? Otherwise invalid transactions will reduce the block gas limit.

world-chain-builder/crates/world/pool/src/tx.rs
```diff
@@ -14,31 +14,31 @@ use reth_primitives::{kzg::KzgSettings, Recovered};
...
+    pub payload: Option<Vec<PBHPayload>>,
 }
 
 pub trait WorldChainPoolTransaction: EthPoolTransaction {
-    fn valid_pbh(&self) -> bool;
-    fn set_valid_pbh(&mut self);
+    fn set_pbh_payloads(&mut self, payload: Vec<PBHPayload>);
     fn conditional_options(&self) -> Option<&TransactionConditional>;
+    fn pbh_payload(&self) -> Option<&Vec<PBHPayload>>;
 }
```
0xKitsune, line 30: Consider updating to use a slice `fn pbh_payload(&self) -> Option<&[PBHPayload]>;`

0xKitsune (approved):

contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -294,19 +312,47 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, WorldIDImpl, ReentrancyGuardTran
...
+    /// @notice Removes a builder from the list of authorized builders.
+    /// @param builder The address of the builder to deauthorize.
+    function removeBuilder(address builder) external virtual onlyProxy onlyOwner {
+        delete authorizedBuilder[builder];
+        emit BuilderDeauthorized(builder);
+    }
+
+    /// @notice Allows a builder to spend all nullifiers within PBH blockspace.
+    /// @param _nullifierHashes The nullifier hashes to spend.
+    function spendNullifierHashes(uint256[] memory _nullifierHashes) external virtual onlyProxy onlyBuilder {
```
0xKitsune:
```suggestion
    function spendNullifierHashes(uint256[] calldata _nullifierHashes) external virtual onlyProxy onlyBuilder {
```

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -273,9 +238,61 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+        vm.assume(addr != OWNER);
+        vm.prank(addr);
+        vm.expectRevert(abi.encodeWithSelector(OwnableUpgradeable.OwnableUnauthorizedAccount.selector, addr));
+
+        pbhEntryPoint.removeBuilder(addr);
+    }
+
+    function test_spendNullifierHashes(uint256[] memory nullifierHashes) public {
+        vm.prank(BLOCK_BUILDER);
+        vm.expectEmit(true, true, false, false);
```
0xKitsune:
```suggestion
        vm.expectEmit(true, true, true, true);
```

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -273,9 +238,61 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+        emit PBHEntryPointImplV1.NullifierHashesSpent(BLOCK_BUILDER, nullifierHashes);
+        pbhEntryPoint.spendNullifierHashes(nullifierHashes);
```
0xKitsune, line 280: Can we assert state changes after this call?

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -273,9 +238,61 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+        emit PBHEntryPointImplV1.BuilderDeauthorized(addr);
+        pbhEntryPoint.removeBuilder(addr);
+    }
+
+    function test_removeBuilder_RevertIf_NotOwner(address addr) public {
+        vm.assume(addr != OWNER);
+        vm.prank(addr);
+        vm.expectRevert(abi.encodeWithSelector(OwnableUpgradeable.OwnableUnauthorizedAccount.selector, addr));
+
+        pbhEntryPoint.removeBuilder(addr);
```
0xKitsune, line 273: Can we assert that no state changes occured after this call?

contracts/test/PBHEntryPointImplV1.t.sol
```diff
@@ -273,9 +238,61 @@ contract PBHEntryPointImplV1Test is TestSetup {
...
+    function test_addBuilder_RevertIf_NotOwner(address addr) public {
+        vm.assume(addr != OWNER);
+        vm.prank(addr);
+        vm.expectRevert(abi.encodeWithSelector(OwnableUpgradeable.OwnableUnauthorizedAccount.selector, addr));
+        pbhEntryPoint.addBuilder(addr);
+    }
+
+    function test_removeBuilder(address addr) public {
+        vm.prank(OWNER);
+        vm.expectEmit(true, false, false, false);
```
0xKitsune: When evaluating event emission, [it is best practice to set all args to true](https://book.getfoundry.sh/guides/best-practices?highlight=expectEmit#general-test-guidance).

This ensures:
- You always test everything in your event and removes the chance of accidentally missing something.
- If you add a topic (i.e. a new indexed parameter), it’s now tested by default.

Even if you only have 1 topic, the extra true arguments don’t hurt.


```suggestion
        vm.expectEmit(true, true, true, true);
```

world-chain-builder/crates/world/pool/src/validator.rs
```diff
@@ -22,15 +22,14 @@ use reth_optimism_primitives::OpTransactionSigned;
...
 /// The slot of the `pbh_gas_limit` in the PBHEntryPoint contract.
-pub const PBH_GAS_LIMIT_SLOT: U256 = U256::from_limbs([305, 0, 0, 0]);
+pub const PBH_GAS_LIMIT_SLOT: U256 = U256::from_limbs([53, 0, 0, 0]);
 
 /// The slot of the `pbh_nonce_limit` in the PBHEntryPoint contract.
-pub const PBH_NONCE_LIMIT_SLOT: U256 = U256::from_limbs([302, 0, 0, 0]);
+pub const PBH_NONCE_LIMIT_SLOT: U256 = U256::from_limbs([50, 0, 0, 0]);
 
 /// The offset in bits of the `PBH_NONCE_LIMIT_SLOT` containing the u16 nonce limit.
 pub const PBH_NONCE_LIMIT_OFFSET: u32 = 160;
```
0xKitsune, line 35: Lets plan to add tests to assert that these are reading from the correct slot. We can plan to add this in a subsequent PR since this is already evaluated indirectly through other tests.

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -616,11 +606,20 @@ where
...
-                continue;
+            if let Some(payloads) = pooled_tx.pbh_payload() {
+                if info.cumulative_gas_used + tx.gas_limit() > verified_gas_limit {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    continue;
+                }
+
+                if payloads
+                    .iter()
+                    .any(|payload| spent_nullifier_hashes.contains(&payload.nullifier_hash))
```
0xKitsune: I think this may allow an edge case where a bundler provides `n` PBH UserOps all with the same PBH Payload/nullifier. This may already be accounted for if we are ensuring there are no duplicates somewhere else in the validation logic.

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -644,7 +643,18 @@ where
             }
 
             let gas_used = match builder.execute_transaction(tx.clone()) {
-                Ok(res) => res,
+                Ok(res) => {
+                    if let Some(payloads) = pooled_tx.pbh_payload() {
+                        if spent_nullifier_hashes.is_empty() {
+                            block_gas_limit -= FIXED_GAS
+                        }
```
0xKitsune: An alternative approach is to optimistically reduce the `block_gas_limit` by `FIXED_GAS`. This assumes that the block will always have at least one PBH tx and in the case where it does not, the block will have `21000` less gas capacity.

0xKitsune (approved):

contracts/test/PBHEntryPointImplV1.t.sol
0xOsiris: If we only have a single indexed topic to test wouldn't this only be asserting the first indexed topic matches?

world-chain-builder/crates/world/payload/src/builder.rs
0xOsiris: I don't think they could all have the same PBH Payload because the `signal` hash to the proof for all of them would invariantly be different. However they could potentially reuse the same nullifier hashes while still recomputing the proof. Assuming all re-used proof's are on the same identity

contracts/test/PBHEntryPointImplV1.t.sol
0xOsiris: Not sure if that's really feasible since we are fuzzing `nullifierHashes`, this is an array of arbitrarily large size that we would have to index elements of individually

world-chain-builder/crates/world/pool/src/validator.rs
0xOsiris: Yeah it's definitely evaluated correctly through Kurtosis tests, but E2E tests in the builder are kind of fake inserting values in the correct slots.

world-chain-builder/crates/world/payload/src/builder.rs
```diff
@@ -643,8 +625,35 @@ where
...
+
+                if payloads.iter().any(|payload| {
+                    let dup = spent_nullifier_hashes.contains(&payload.nullifier_hash);
+                    spent_nullifier_hashes.insert(payload.nullifier_hash);
+                    dup
+                }) {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    invalid_txs.push(*pooled_tx.hash());
+                    continue;
+                }
```
0xKitsune: Since we are using a `HashSet` this can be further simplified to the following:


```rust
if payloads.iter().any(|payload| !spent_nullifier_hashes.insert(payload.nullifier_hash)) {
    best_txs.mark_invalid(tx.signer(), tx.nonce());
    invalid_txs.push(*pooled_tx.hash());
    continue;
}
```

PR #199, opened 2025-03-25
Osiris/refactor devnet

0xOsiris:
This PR refactor's the devnet to spawn the kurtosis package standalone, and hook in containerized builder op-node, and world-chain-builder instances into the rollup-boost service running in Kurtosis.

Major Benefits:
- We never have to worry about maintaining the full op-stack. As upgrades come through things invariantly break
- Fast iteration on redeployment of the builder without redeploying the entire stack
- Better visibility, and customizability of observability in the world-chain-builder

Downsides:
- Currently I'm binding the builders ip to the host ip which isn't ideal (had to do this so the Kurtosis services could reach the builder)
  - happy to hear any thoughts if there's a better way to do this
- Setup is a bit hacky but it works

There is another way to do this which is to create a starlark package that wraps around the `ethereum-package`, and `optimism-package` but this is a bit tricky as well


devnet/kurtosis/network_params.yaml
```diff
@@ -45,12 +44,11 @@ optimism_package:
         extra_params: []
       mev_params:
         rollup_boost_image: ""
-        builder_host: ""
-        builder_port: ""
+        builder_host: "192.168.0.91"
```
0xOsiris, line 47: This value is automatically configured with `just devnet-up`

PR #204, opened 2025-04-01
feat(flashblocks): Initial `FlashblocksPayloadBuilder` implementation

0xKitsune:
This PR introduces a `FlashBlocksPayloadBuilder` that extends Reth’s payload building logic to support streaming flashblocks during block construction. This PR also defines a new generic `PayloadBuilderCtx` trait, allowing generic payload building logic to be compatible with Flashblocks.

### Summary of Changes

- Adds `FlashBlocksPayloadBuilder`
- Introduces `PayloadBuilderCtx` trait for generic payload building within the `FlashblocksPayloadBuilder`.
- Introduces a `Flashblock` trait and `finish_flashblock()` method which is intended to be used to seal the flashblock before it is streamed. This logic is based on the existing `builder.finish()` implementation that seals the finalized block.

### Remaining Work
- Upstream the `PayloadBuilderCtx` trait to Reth
- Implement the `Flashblock` trait on `BasicBlockBuilder` to support `finish_flashblock()`.
  - The current `BasicBlockBuilder` from Reth is private. This should be updated to be made public which will enable us to implement the `Flashblock` trait.

This PR serves as the initial groundwork for flashblocks to support generic payload building logic. Further integration and testing will follow once the upstream dependencies are in place.

### DIFF
--- a/world-chain-builder/crates/flashblocks/payload/src/builder.rs
+++ b/world-chain-builder/crates/flashblocks/payload/src/builder.rs
@@ -1,8 +1,3 @@
-use std::{
-    collections::HashMap,
-    sync::{Arc, Mutex},
-};
-
 use alloy_consensus::{Header, EMPTY_OMMER_ROOT_HASH};
 use alloy_eips::merge::BEACON_NONCE;
 use alloy_primitives::{Address, B256, U256};
@@ -14,6 +9,8 @@ use reth::{
 };
 use reth_basic_payload_builder::{BuildArguments, BuildOutcome, BuildOutcomeKind};
 use reth_basic_payload_builder::{MissingPayloadBehaviour, PayloadBuilder, PayloadConfig};
+use reth_chain_state::{ExecutedBlock, ExecutedBlockWithTrieUpdates};
+use reth_evm::block::BlockExecutorFactory;
 use reth_evm::{
     execute::{BlockBuilder, BlockBuilderOutcome},
     ConfigureEvm, Evm,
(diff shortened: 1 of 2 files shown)


world-chain-builder/crates/flashblocks/payload/src/builder.rs
```diff
@@ -142,31 +124,37 @@ impl<Pool, Client, Evm, Txs> FlashBlocksPayloadBuilder<Pool, Client, Evm, Txs> {
     }
 }
 
-impl<Pool, Client, Evm, N, Txs> PayloadBuilder for FlashBlocksPayloadBuilder<Pool, Client, Evm, Txs>
+impl<Pool, Client, Evm, N, T> FlashBlocksPayloadBuilder<Pool, Client, Evm, T>
```
0xOsiris, line 127: Not super important, but does this `T` need to be here if we isolate the generic below to `Txs`?

Maybe we could just bind `Txs` to `PayloadTransactions<Transaction: PoolTransaction<Consensus = N::SignedTx>>` on the struct/trait definition so we don't have a floating generic

world-chain-builder/crates/flashblocks/payload/src/builder.rs
```diff
@@ -176,270 +164,348 @@ where
...
+        let state_provider = self.client.state_by_block_hash(ctx.parent().hash())?;
+        let state = StateProviderDatabase::new(&state_provider);
+
+        if ctx.attributes().no_tx_pool {
+            builder.build(state, &state_provider, ctx)
+        } else {
+            // sequencer mode we can reuse cachedreads from previous runs
+            builder.build(cached_reads.as_db_mut(state), &state_provider, ctx)
+        }
+        .map(|out| out.with_cached_reads(cached_reads))
```
0xOsiris, line 182: What is this doing aha

0xKitsune: This is the same logic [used in the `OpPayloadBuilder`](https://github.com/paradigmxyz/reth/blob/main/crates/optimism/payload/src/builder.rs#L167-L173). I reimplemented it here for parity.

world-chain-builder/crates/flashblocks/payload/src/builder.rs
```diff
@@ -176,270 +164,348 @@ where
...
         })?;
 
-        // execute sequencer transactions
+        // 2. execute sequencer transactions
         let mut info = ctx.execute_sequencer_transactions(&mut builder)?;
 
-        let mut next_block = None;
         if !ctx.attributes().no_tx_pool {
-            let tx_attrs = ctx.best_transaction_attributes(builder.evm_mut().block());
+            let flashblock_gas_limit = ctx.attributes().gas_limit.unwrap_or(ctx.parent().gas_limit)
```
0xOsiris: This could also be a moving value based on the total amount of gas consumed. If you just subtract `gas_limit - total_gas_used` then you have maximum flexibility to fill up the full gas limit. 

Doing it this way will potentially restrict individual flash blocks from filling the gas limit if previous flash blocks used a lower amount of gas.

0xOsiris (approved):

world-chain-builder/crates/flashblocks/payload/src/builder.rs
0xOsiris: Oh I just read it wrong my b

world-chain-builder/crates/flashblocks/payload/src/builder.rs
0xKitsune: Great idea, Ill update this now.

PR #205, opened 2025-04-02
fix: sleep between flashblocks + minor cleanup

Dzejkop:


world-chain-builder/crates/flashblocks/payload/src/builder.rs
```diff
@@ -338,6 +337,8 @@ impl<Txs> FlashblockBuilder<'_, Txs> {
 
             // tx.send(serde_json::to_string(&fb_payload).unwrap_or_default())
             //     .expect("TODO: handle error");
+
+            std::thread::sleep(Duration::from_secs(self.flashblock_interval));
```
0xKitsune: I think this would result in the builder sleeping for an `flashblock_interval` + `time_elapsed` while executing the txs. We should update the payload builder logic to keep track of how long it takes to execute the best txs and then sleep for `flashblock_interval - time_elapsed`.

0xKitsune (approved):

PR #212, opened 2025-04-08
(flashblocks) Make things compile

Dzejkop:
Makes things compile again after the restructure & reth update. Also contains some drive-by changes I'm too lazy to extract


Cargo.toml
```diff
@@ -153,3 +158,48 @@ rand = { version = "0.8", features = ["small_rng"] }
 reqwest = { version = "0.12", default-features = false }
 bon = "3.3.0"
 rayon = "1.10.0"
+
```
0xForerunner, line 161: Probably wanna remove this!

PR #220, opened 2025-04-23
Flashblocks: Connect most components

Dzejkop:
Most components are connected and (in-theory) working.

What's missing is the WS service to stream the flashblocks & e2e testing


crates/world/flashblocks/src/builder/retaining_payload_txs.rs
```diff
@@ -1,6 +1,8 @@
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 
+use alloy_primitives::TxHash;
 use reth_payload_util::PayloadTransactions;
+use reth_transaction_pool::PoolTransaction;
 
 /// This type exists to yield bes transactions from the tx pool
```
0xOsiris:
```suggestion
/// This type exists to yield best transactions from the tx pool
```

crates/world/flashblocks/src/builder/retaining_payload_txs.rs
```diff
@@ -48,22 +57,41 @@ where
...
     type Transaction = I::Transaction;
 
     fn next(&mut self, ctx: ()) -> Option<Self::Transaction> {
-        // TODO: Implement
-        self.retaining.inner.next(ctx)
+        if let Some(n) = self.inner.prev.pop_front() {
+            self.inner.observed_hashes.insert(*n.hash());
+            self.inner.observed.push(n.clone());
+
+            return Some(n);
```
0xOsiris, line 75: Why do we yield previously observed transactions prior to the current transactions in the `best_txs` iterator even if the priority fee might be lower. Would this potentially break priority ordering invariants on all subsequent flash blocks except the first?

0xOsiris: I think we want to check if the priority fee on `self.inner.inner.next(ctx)` is higher than `self.inner.prev.pop_front()` prior to yielding the value

0xOsiris (approved):
LGTM - just one question/suggestion. Nice work!

crates/world/flashblocks/src/builder/retaining_payload_txs.rs
0xOsiris: Alternatively we could not keep track of `prev`

crates/world/flashblocks/src/builder/retaining_payload_txs.rs
```diff
@@ -1,6 +1,8 @@
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 
+use alloy_primitives::TxHash;
 use reth_payload_util::PayloadTransactions;
+use reth_transaction_pool::PoolTransaction;
 
 /// This type exists to yield bes transactions from the tx pool
 /// while doing bookkeping so we can deterministically replay them on the following
```
0xOsiris:
```suggestion
/// while doing bookkeeping so we can deterministically replay them on the following
```

crates/world/flashblocks/src/builder/retaining_payload_txs.rs
Dzejkop: this is so that we're iteratively building upon the same block every 250ms.

This guarantess that flashblock n+1 builds on top of flashblock n

0xOsiris: Maybe I'm misunderstanding, but all of the transactions of `prev` should be a part of [this iterator](https://github.com/worldcoin/world-chain/blob/f69b4f3479f0d359e9c056cf9985c442916aeefa/crates/world/flashblocks/src/builder/mod.rs#L491) they just may not be at the top of the iterator as order may have changed due to transactions with higher priority fees. Right now we are treating `prev` with highest priority in `n + 1` when maybe they shouldn't be included in that block.

Dzejkop: it's true, but prev just contains transactions included in the previous flashblock(s). The current implementation is not the best since we're rebuilding a block from scratch every time. This mechanism ensures that if txs were included in prior flashblocks then they will also be included, in the same order, in this new flashblock we're building here

PR #223, opened 2025-04-25
World Chain Snapshotter

0xForerunner:
Wraps reth dockerfile with bash script. Script forwards all args to reth, periodically stops reth to take snapshot, restarts, and uploads to s3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ WORKDIR /app
 
 # Install wget in the final image
 RUN apt-get update && \
-    apt-get install -y wget && \
+    apt-get install -y wget netcat-traditional && \
     rm -rf /var/lib/apt/lists/*
 
 


snapshotter/snapshot.sh
```diff
@@ -0,0 +1,118 @@
...
+  if [[ "$hex_block" == "null" || -z "$hex_block" ]]; then
+    echo "Error: Could not retrieve block number" >&2
+    return 2
+  fi
+
+  # Convert hex to decimal
+  echo $((hex_block))
+}
+
+is_synced() {
```
0xOsiris, line 57: I think we can use `eth_syncing` here instead of hooking into an external provider, but this works as well! Reth should [return a null response when it is done syncing](https://github.com/paradigmxyz/reth/blob/59b717e8f455fb13e8d80e479b43d8d01a224b39/crates/rpc/rpc-eth-api/src/core.rs#L385), and we could use that as a signal.

0xOsiris (approved):

PR #235, opened 2025-05-15
fix(validator): Check duplicate nullifier hashes in PBH bundle

karankurbur:
Check duplicate nullifier hashes in pool validation

### DIFF
--- a/crates/world/payload/src/builder.rs
+++ b/crates/world/payload/src/builder.rs
@@ -620,11 +620,20 @@ where
 
             // If the transaction is verified, check if it can be added within the verified gas limit
             if let Some(payloads) = pooled_tx.pbh_payload() {
-                if info.cumulative_gas_used + tx.gas_limit() > verified_gas_limit {
+                // Step 1: Collect all nullifier_hashes for this transaction
+                let mut same_tx_nullifiers = std::collections::HashSet::new();
+                let has_duplicate_nullifier_in_same_tx = payloads
+                    .iter()
+                    .any(|payload| !same_tx_nullifiers.insert(payload.nullifier_hash));
+
+                // Step 2: If there are duplicates within the transaction, mark as invalid and don't add to spent_nullifier_hashes
+                if has_duplicate_nullifier_in_same_tx {
                     best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    invalid_txs.push(*pooled_tx.hash());
                     continue;
                 }
 
+                // Step 3: Check for duplicates across all transactions.
                 if payloads
                     .iter()
                     .any(|payload| !spent_nullifier_hashes.insert(payload.nullifier_hash))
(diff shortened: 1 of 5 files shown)


crates/world/payload/src/builder.rs
```diff
@@ -624,7 +624,20 @@ where
                     best_txs.mark_invalid(tx.signer(), tx.nonce());
                     continue;
                 }
+                // Step 1: Collect all nullifier_hashes for this transaction
+                let mut same_tx_nullifiers = std::collections::HashSet::new();
```
0xOsiris:
```suggestion
                let mut same_tx_nullifiers = HashSet::new();
```

0xOsiris (approved):

crates/world/payload/src/builder.rs
```diff
@@ -624,7 +624,20 @@ where
...
+                    .any(|payload| !same_tx_nullifiers.insert(payload.nullifier_hash));
+
+                // Step 2: If there are duplicates within the transaction, mark as invalid and don't add to spent_nullifier_hashes
+                if has_duplicate_nullifier_in_same_tx {
+                    best_txs.mark_invalid(tx.signer(), tx.nonce());
+                    invalid_txs.push(*pooled_tx.hash());
+                    continue;
+                }
 
+                // Step 3: Check for duplicates across all transactions.
```
0xKitsune: Wdyt about updating the PBH validation logic to reject duplicate nullifier hashes in the same bundle? That way, we can avoid making any changes to the payload building logic to address the issue.

crates/world/pool/src/validator.rs
```diff
@@ -179,16 +182,25 @@ where
...
             {
-                Ok(payloads) => {
-                    aggregated_payloads.extend(payloads);
-                }
+                Ok(payloads) => payloads,
                 Err(err) => return err.to_outcome(tx),
+            };
+
+            // Now check for duplicate nullifier_hashes
+            for payload in &payloads {
```
0xOsiris, line 194: One nit - this isn't checking cross `aggregated_ops` duplicates, so it's still possible to have a duplicate nullifier hash in the same bundle in a separate `AggregatedOps`.

Just move this logic to iterate over `aggregated_payloads` after they have been accumulated to fix.

0xOsiris (approved):
LGTM 🚢

crates/world/pool/src/validator.rs
```diff
@@ -179,16 +182,25 @@ where
...
                 .collect::<Result<Vec<PbhPayload>, WorldChainPoolTransactionError>>()
             {
-                Ok(payloads) => {
-                    aggregated_payloads.extend(payloads);
-                }
+                Ok(payloads) => payloads,
                 Err(err) => return err.to_outcome(tx),
+            };
+
+            // Now check for duplicate nullifier_hashes
```
0xKitsune, line 193: Nit: Since `seen_nullifier_hashes` is only used here can we initialize the hashset right before the dedup logic rather than before parsing the PBH payloads?

0xKitsune (approved):

PR #249, opened 2025-05-27
fix: update `MaintainPoolConfig`

0xKitsune:


crates/world/node/src/node.rs
```diff
@@ -277,10 +288,17 @@ where
...
                     chain_events,
                     ctx.task_executor().clone(),
-                    Default::default(),
+                    reth_transaction_pool::maintain::MaintainPoolConfig {
+                        max_tx_lifetime: pool.config().max_queued_lifetime,
+                        no_local_exemptions: transaction_pool
+                            .config()
+                            .local_transactions_config
+                            .no_exemptions,
+                        ..Default::default()
```
0xOsiris, line 300: Are the other config items not configurable over CLI?

0xKitsune: This is the same configuration as [reth](https://github.com/paradigmxyz/reth/blob/01befb2415c9426d25161126e57a3b2ce0285d28/crates/optimism/node/src/node.rs#L690-L697) as well as [op-reth](https://github.com/paradigmxyz/reth/blob/01befb2415c9426d25161126e57a3b2ce0285d28/crates/ethereum/node/src/node.rs#L396-L403). It looks like `max_update_depth` and `max_reload_accounts` are not configurable over CLI.

PR #252, opened 2025-06-11
Osiris/flashblocks devnet

0xOsiris:
Adds Custom rollup-boost launcher to devnet for testing Flashblocks

Stacked on #251


crates/world/bin/Cargo.toml
```diff
@@ -48,8 +48,8 @@ reth-optimism-consensus.workspace = true
 reth-optimism-forks.workspace = true
 reth-optimism-primitives = { workspace = true, features = ["serde"] }
 
-# revm with required optimism features
-revm = { workspace = true, features = ["secp256k1", "blst", "c-kzg"] }
+# revm 
+revm = { workspace = true }
```
0xKitsune:
```suggestion
revm.workspace = true
```

crates/world/pool/src/tx.rs
```diff
@@ -44,6 +48,18 @@ impl WorldChainPoolTransaction for WorldChainPooledTransaction {
     }
 }
 
+impl OpPooledTx for WorldChainPooledTransaction {
+    fn encoded_2718(&self) -> std::borrow::Cow<'_, Bytes> {
+        Cow::Borrowed(self.inner.encoded_2718())
```
0xKitsune: Is `Cow::Borrowed()` necessary here or can we just return `self.inner.encoded_2718()`?

0xKitsune (approved):

crates/world/pool/src/tx.rs
0xOsiris: I think because of the return type right? `encoded_2718()` just returns `Bytes`

PR #258, opened 2025-07-04
feat: add UserOp hash to PBH event

0xOsiris:
Adds the UserOp hash as an indexed topic to the `PBH` event. This is necessary for ease of indexing PBH transactions from the UserOp hash stored by app backend.


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -71,8 +74,9 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, Base, ReentrancyGuardTransient {
     ///
     /// @param sender The sender of this particular transaction or UserOp.
     /// @param signalHash Signal hash associated with the PBHPayload.
+    /// @param userOpHash The hash of the UserOperation that contains the PBHPayload.
     /// @param payload The zero-knowledge proof that demonstrates the claimer is registered with World ID.
-    event PBH(address indexed sender, uint256 indexed signalHash, PBHPayload payload);
+    event PBH(address indexed sender, uint256 indexed signalHash, bytes32 indexed userOpHash, PBHPayload payload);
```
0xOsiris: Should we remove `signalHash` to save some gas here if it's not going to be used for indexing? @aurel-fr @0xKitsune @karankurbur

0xOsiris (commented):
Not sure what's going on with CI. Will fix in a different PR. Locally e2e tests are passing

PR #260, opened 2025-07-08
Add helper function to return index of unused nullifierHash

karankurbur:
The client must calculate, given the semaphore private key + nonce, the list of nullifierHashes for the month. This function will return the index of the first unused nullifierHash or -1 if all have been consumed.

### DIFF
--- a/contracts/src/PBHEntryPointImplV1.sol
+++ b/contracts/src/PBHEntryPointImplV1.sol
@@ -367,15 +367,39 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, Base, ReentrancyGuardTransient {
     }
 
     /// @notice Returns the index of the first unspent nullifier hash in the given list.
+    /// @notice This function assumes the array input represents the nullifier hashes for a given sempahore key and monotonically increasing nonces for a given month.
     /// @param hashes The list of nullifier hashes to search through.
     /// @return The index of the first unspent nullifier hash in the given list.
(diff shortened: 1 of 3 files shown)


contracts/src/PBHEntryPointImplV1.sol
```diff
@@ -365,4 +365,17 @@ contract PBHEntryPointImplV1 is IPBHEntryPoint, Base, ReentrancyGuardTransient {
     function getUserOpHash(PackedUserOperation calldata userOp) public view virtual returns (bytes32 hash) {
         hash = keccak256(abi.encode(userOp.hash(), address(entryPoint), block.chainid));
     }
+
+    /// @notice Returns the index of the first unspent nullifier hash in the given list.
+    /// @param hashes The list of nullifier hashes to search through.
+    /// @return The index of the first unspent nullifier hash in the given list.
+    /// @dev Returns -1 if no unspent nullifier hash is found.
+    function getUnspentNullifierHash(uint256[] calldata hashes) public view virtual returns (int256) {
```
0xForerunner: It's possible that nullifiers have been used out of order. It might be more flexible if instead we returned each nullifier that has not yet been used (or maybe its index?)

0xOsiris: Updated

0xForerunner (approved):

0xOsiris (approved):
LGTM

PR #261, opened 2025-07-16
Osiris/patch transition state

0xOsiris:


crates/world/payload/src/builder.rs
```diff
@@ -411,7 +411,7 @@ impl<Txs> WorldChainBuilder<'_, Txs> {
             // TODO: Validate gas limit
             if ctx
                 .execute_best_transactions(&mut info, &mut builder, best_txs, 0)?
-                .is_some()
+                .is_none()
```
0xOsiris, line 416: Note I changed the `None` `Some` match here to be able to pass `BundleState` through the `Some` variant. This changes the logic in a very minor way for our canonical block builder.

crates/world/flashblocks/src/builder/mod.rs
```diff
@@ -252,135 +253,191 @@ where
...
+            id = %ctx.payload_id(),
+        );
 
-        // NOTE: Do we need to init this every time?
-        let mut db = State::builder()
-            .with_database(db)
-            .with_bundle_update()
-            .build();
+        let _enter = span.enter();
+        debug!(target: "payload_builder", id=%ctx.payload_id(), "building new payload");
```
0xForerunner: The event labels are automatically inherited from the enclosing span, so there's no reason to repeat "id" here.

crates/world/flashblocks/src/builder/mod.rs
```diff
@@ -389,146 +446,230 @@ where
...
+                            debug!(target: "payload_builder", "parent cancel was dropped, cancelling flashblock job");
+                            // parent cancel was dropped by the payload jobs generator
+                            // cancel the child job, and drop the sender
+                            break;
+                        }
+
+                        tokio::select! {
+                            _ = flashblock_interval.tick() => {
+                                let elapsed = instant.elapsed();
+                                debug!(target: "payload_builder", ?elapsed, "flashblock interval exceeded, cancelling current job");
```
0xForerunner: Log message is inconsistent with behaviour

crates/world/flashblocks/src/builder/mod.rs
```diff
@@ -252,135 +253,191 @@ where
...
+            debug!(target: "payload_builder", id=%ctx.attributes().payload_id(), "building flashblock {flashblock_idx}");
+            let state = StateProviderDatabase::new(&state_provider);
+
+            let mut db = State::builder()
+                .with_database(state)
+                .with_bundle_update()
+                .with_bundle_prestate(bundle_state)
+                .build();
+
+            let notify = tokio::task::block_in_place(|| rx.blocking_recv());
```
0xForerunner, line 346: Are we sure block in place is necessary here? `fn build` is already a long running computation heavy synchronous function. I assume reth internal already expects this function block significantly.

crates/world/flashblocks/src/builder/mod.rs
```diff
@@ -252,135 +253,191 @@ where
...
+
+            match notify {
+                Some(()) => {
```
0xForerunner, line 356: Would be nice to create a new span here when actually building the flashback.

crates/world/flashblocks/src/builder/payload_txns.rs
```diff
@@ -0,0 +1,81 @@
...
+            observed: Vec::new(),
+        }
+    }
+
+    pub fn with_prev(mut self, prev: Vec<TxHash>) -> Self {
+        self.prev.extend(prev);
+        self
+    }
+
+    pub fn take_observed(self) -> (Vec<TxHash>, Vec<TxHash>) {
```
0xForerunner: More efficient to return impl Iterator here.

crates/world/flashblocks/src/builder/payload_txns.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+impl<'a, I> PayloadTransactions for BestPayloadTxnsGuard<'a, I>
+where
+    I: PayloadTransactions<Transaction: PoolTransaction + Clone>,
+{
+    type Transaction = I::Transaction;
+
+    fn next(&mut self, ctx: ()) -> Option<Self::Transaction> {
+        loop {
```
0xForerunner: prefer `while let Some(n)`

0xForerunner (commented):
Looks great man! Very well done.

crates/world/flashblocks/src/builder/mod.rs
0xOsiris: Yeah it's necessary - I get this without it:
```Critical task `payload builder service` panicked: `Cannot block the current thread from within a runtime. This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.```

PR #277, opened 2025-08-14
feat(flashblocks): payload building orchestration & engine api extension

0xOsiris:
### Overview
*1. P2P Start/Stop Signal Refactor*
- The logic for handling start and stop signals has been moved out of the `FlashblocksPayloadBuilder`.
- This decouples signal processing from the `FlashblocksHandle` & payload construction

*2. Pre-confirmation Caching*
- Pre-confirmations are now cached as fully built blocks in the `PayloadStore` of the `PayloadBuilderHandle`
- This allows the node to serve blocks at any point after the first build job has resolved

*3. Engine API Extension: `flashblocks_engineForkChoiceUpdatedV3`*
- Introduces the `flashblocks_engineForkChoiceUpdatedV3` method to the Engine API.
- Integrates `Authorization`  management from rollup-boost into the payload building lifecycle

*4. Makes `FlashblockState` aware of it's confirmation status & removes redundant execution when we have a failover scenario*


.github/workflows/rust-ci.yml
```diff
@@ -80,6 +80,6 @@ jobs:
             key: cargo-test-${{ hashFiles('**/Cargo.lock') }}
             restore-keys: cargo-test-   
         - name: clippy
-          run: cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings
```
0xForerunner: why remove --workspace?

crates/world/bin/src/main.rs
```diff
@@ -51,8 +53,20 @@ fn main() {
                     flashblocks_tx.clone(),
                 );
 
-                let node =
-                    WorldChainFlashblocksNode::new(args.clone(), flashblocks_handle, state.clone());
+                let authorization = Authorization::new(
```
0xForerunner: Should we prefer using option here?

crates/world/flashblocks/src/payload/job.rs
```diff
@@ -0,0 +1,394 @@
...
+            BuiltPayload = OpBuiltPayload,
+            Attributes = OpPayloadBuilderAttributes<OpTxEnvelope>,
+        > + Unpin
+        + 'static,
+    Builder::Attributes: Unpin + Clone,
+    Builder::BuiltPayload: Unpin + Clone,
+{
+    type Output = Result<(), PayloadBuilderError>;
+
+    /// Polls the payload builder job to drive payload construction and management.
```
0xForerunner, line 184: These doc comments would be more useful on the stuct definition.

crates/world/flashblocks/src/payload/job.rs
```diff
@@ -0,0 +1,394 @@
...
+        let this = self.get_mut();
+        // check if the deadline is reached
+        if this.deadline.as_mut().poll(cx).is_ready() {
+            trace!(target: "payload_builder", "payload building deadline reached");
+            return Poll::Ready(Ok(()));
+        }
+
+        let mut tick_fut = Box::pin(this.interval.tick());
+
+        while tick_fut.as_mut().poll(cx).is_pending() {
```
0xForerunner: I'm confused by this, don't we only want to call the waker when tick_fut is Ready?

crates/world/flashblocks/src/payload/job.rs
```diff
@@ -0,0 +1,394 @@
...
+{
+    type PayloadAttributes = Builder::Attributes;
+    type ResolvePayloadFuture = ResolveBestPayload<Self::BuiltPayload>;
+    type BuiltPayload = Builder::BuiltPayload;
+
+    fn best_payload(&self) -> Result<Self::BuiltPayload, PayloadBuilderError> {
+        if let Some(payload) = self.best_payload.payload() {
+            Ok(payload.clone())
+        } else {
+            info!(target: "payload_builder", id=%self.config.payload_id(), "no best payload available, building empty payload");
```
0xForerunner, line 316: Is there any situation where this shouldn't be a warn!?

0xForerunner (approved):
Nice man this looks great.

crates/world/flashblocks/src/payload/job.rs
0xOsiris: I don't think this is an uncommon path since Op Node will ask for blocks sometimes almost immediately after the FCU. In this case we race an empty payload, but we haven't published any pre-confs so I think its fine

.github/workflows/rust-ci.yml
0xOsiris: Will add it back, have no idea

PR #285, opened 2025-08-22
use first 5bytes in userOp nonce as PBH flag

karankurbur:
In bedrock, we will set the first 5 bytes of the userOp nonce to "bytes("pbhtx")". We are using 5 bytes instead of 2 bytes now to reduce the chance of a collision of a non-worldapp tx using these same bytes as well to match the 5 byte length of the magic "bdrck" constant as well.

The magic "bdrck" will be used for non-pbh transactions for the indexer to determine it was a bedrock tx. The indexer will also look for the magic "pbhtx"


0xOsiris (changes requested):
You need to update the e2e UserOp construction to reflect this change

PR #287, opened 2025-08-26
Flashblocks pending state

0xForerunner:


crates/flashblocks/primitives/src/flashblocks/args.rs
```diff
@@ -0,0 +1,14 @@
+use clap::Args;
+use ed25519_dalek::{SigningKey, VerifyingKey};
+
+use hex::FromHex;
+
+pub fn parse_sk(s: &str) -> eyre::Result<SigningKey> {
```
0xOsiris: This is a dup

crates/flashblocks/primitives/src/flashblocks/args.rs
```diff
@@ -0,0 +1,14 @@
...
+    let bytes = <[u8; 32]>::from_hex(s.trim())?;
+    Ok(SigningKey::from_bytes(&bytes))
+}
+
+pub fn parse_vk(s: &str) -> eyre::Result<VerifyingKey> {
```
0xOsiris: This is also a dup

0xForerunner: Yeah flashblocks node crate need to be refactored to use the new RPC components.

crates/flashblocks/primitives/src/flashblocks.rs
```diff
@@ -245,72 +239,106 @@ impl TryFrom<Flashblock> for RecoveredBlock<Block<OpTxEnvelope>> {
...
+    /// of the collection (index must be contiguous).
+    /// Returns `Ok(true)` if the collection was reset, `Ok(false)` if the flashblock
+    /// was added to the existing collection, or an error if the conditions are not met.
+    pub fn push(&mut self, flashblock: Flashblock) -> eyre::Result<bool> {
+        if flashblock.payload_id() != self.payload_id() {
+            if flashblock.flashblock.index != 0 {
+                bail!("New flashblock has different payload_id and index is not 0");
+            }
+            let Some(base) = &flashblock.flashblock.base else {
+                bail!("New flashblock has different payload_id and must contain the base payload");
```
0xOsiris, line 296: Seems like these should be `debug_assert`s

0xForerunner: I prefer the hard check here, just in case haha. Things get get really bad if this check fails.

0xOsiris (approved):

PR #291, opened 2025-08-28
Forerunner/fix features

0xForerunner:


Cargo.toml
```diff
@@ -7,25 +7,20 @@ homepage = "https://world.org/world-chain"
 repository = "https://github.com/worldcoin/world-chain/"
 
 [workspace]
-resolver = "2"
-members = [
-    "crates/toolkit",
+resolver = "3"
+members = ["crates/tests/*", "crates/toolkit", "crates/world/*"]
+default-members = [
```
0xOsiris, line 12: Doesn't default members just need to be the bin?

crates/world/test-e2e/tests/builder.rs
0xOsiris, line 1: Why can't these just stay in a subdirectory of `node/` as they were, and import `world-chain-test` crate as a dev dependency of `world-chain-builder-node`

crates/world/test-e2e/tests/testsuite.rs
0xOsiris, line 1: You need to add a `main.rs` here that adds `builder`, and `testsuite` as submodules otherwise cargo check won't pick these targets up

crates/world/test-e2e/Cargo.toml
0xOsiris, line 1: Need to add 

```
[[test]]
name = 
path = 

```

For running nextest binaries

Cargo.toml
0xForerunner: no, default members are all crates that "cargo check" will apply to when called from the root.

crates/world/test-e2e/tests/builder.rs
0xForerunner: I like the separation better. Feels a bit weird to hid our e2e tests inside the node crate.

crates/world/test-e2e/tests/testsuite.rs
0xForerunner: I don't think you're correct. Cargo test is running these tests just fine.

crates/world/test-e2e/Cargo.toml
0xForerunner: This isn't correct. It only applies when you have a subdir in tests.

PR #297, opened 2025-09-04
Osiris/eth api

0xOsiris:
Adds Basic Eth Api over pending block


crates/flashblocks/builder/src/lib.rs
```diff
@@ -456,6 +456,7 @@ where
         }
     }
 
+    #[expect(clippy::type_complexity)]
```
0xForerunner, line 459: What does expect do? Don't we want allow?

crates/world/node/tests/e2e-testsuite/testsuite.rs
```diff
@@ -350,8 +352,95 @@ async fn test_flashblocks() -> eyre::Result<()> {
     Ok(())
 }
 
+// #[tokio::test(flavor = "multi_thread")]
```
0xForerunner, line 355: Stray comments?

crates/world/test/src/mock.rs
```diff
@@ -761,6 +761,10 @@ impl StateProvider for MockEthProvider {
 }
 
 impl StateProviderFactory for MockEthProvider {
+    fn maybe_pending(&self) -> ProviderResult<Option<StateProviderBox>> {
```
0xForerunner, line 764: Do we not want to implement this?

crates/flashblocks/builder/src/lib.rs
0xOsiris: Advantage is it throws a lint if it's no longer necessary imo

crates/world/test/src/mock.rs
0xOsiris: This should already be implemented on `BlockchainProvider`? Not sure we are really using the `MockEthProvider` anywhere where this would have relevance

crates/world/node/tests/e2e-testsuite/testsuite.rs
0xOsiris: I'd like to leave this if that's okay.. this test is a bit difficult to setup. Can wait to merge until it's fixed if you'd prefer

crates/flashblocks/builder/src/lib.rs
0xForerunner: oh cool. TIL.

crates/world/node/tests/e2e-testsuite/testsuite.rs
0xForerunner: Nah all good.

PR #298, opened 2025-09-05
add flashbots authorizer contract

karankurbur:
We need to be able to update the flashbots authorizer on all flashbots enabled nodes in case Alchemy wants to rotate the keys. Currently it is passed as a CLI arg and would require a restart, instead nodes will read from this contract instead.

### DIFF
--- a/contracts/src/Flashbots.sol
+++ b/contracts/src/Flashbots.sol
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
+ * @title FlashbotsAuthorizerRegistry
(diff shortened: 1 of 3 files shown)


contracts/src/FlashbotsAuthorizer.sol
0xOsiris, line 1: `Flashblocks` not `Flashbots`. I would rename it to just `AuthorizerRegistry`

contracts/src/Flashbots.sol
0xOsiris, line 1: Rename this to `AuthorizerRegistry.sol`

0xOsiris, line 1: This is a dup contract

contracts/test/FlashbotsAuthorizer.t.sol
0xOsiris, line 1: FlashbotsAuthorizer.t.sol -> `AuthorizerRegistry.t.sol`

PR #301, opened 2025-09-10
Make FlashblocksPayloadBuilderBuilder generic over CtxBuilder

0xForerunner:


crates/world/flashblocks/src/builder/mod.rs
```diff
@@ -387,7 +393,13 @@ where
                 .with_prev(transactions.iter().map(|tx| *tx.hash()).collect::<Vec<_>>());
 
             if ctx
-                .execute_best_transactions(&mut info, &mut builder, best_txns.guard(), gas_limit)?
+                .execute_best_transactions(
```
0xOsiris, line 396: It'd be nice if we could adhere to the API of `OpPayloadBuilderCtx` on `execute_best_transactions` & `execute_sequencer_transactions` as this is going to be the implementation for the context most are using.

i.e. https://github.com/paradigmxyz/reth/blob/424974ca370d41969e3161230faa5f0dfd16c77a/crates/optimism/payload/src/builder.rs#L654

0xOsiris: Although I don't know if this is actually possible. The context builder can hold the pool, but in any case the `gas_limit` will always need to be passed through

crates/world/node/src/args.rs
```diff
@@ -24,104 +33,167 @@ pub struct WorldChainArgs {
...
+        default_value_t = Default::default(),
     )]
     pub world_id: Address,
 
     /// Sets the ERC0-7766 Signature Aggregator contract address
     /// This contract signifies that a given bundle should receive priority inclusion if it passes validation
     #[arg(
-        long = "builder.signature_aggregator",
-        default_value = "0x0000000000000000000000000000000000000000"
+        long = "pbh.signature_aggregator",
```
0xOsiris, line 118: Let's make sure we update the deployment scripts

crates/world/flashblocks/src/builder/mod.rs
0xForerunner: That makes it much more difficult, and would require us to have yet another trait `PayloadBuilderCtxBuilderBuilder`.

I don't think that's worth it IMO.

crates/world/node/src/args.rs
```diff
@@ -24,104 +33,167 @@ pub struct WorldChainArgs {
...
     /// Sets the ERC-4337 EntryPoint Proxy contract address
     /// This contract is used to validate 4337 PBH bundles
     #[arg(
-        long = "builder.pbh_entrypoint",
-        default_value = "0x0000000000000000000000000000000000000000"
+        long = "pbh.entrypoint",
+        default_value_t = Default::default(),
     )]
-    pub pbh_entrypoint: Address,
+    pub entrypoint: Address,
```
0xOsiris, line 105: I think `pbh_entrypoint` is a better name here. Other something differentiating it from the 4337 entry point

crates/world/node/src/args.rs
0xForerunner: Yeah I just updated the devnet. We can update dev as well.

crates/world/node/src/args.rs
0xForerunner, line 105: I figured pbh.entrypoint is enough context. pbh.pbh_entrypoint feel redundant no?

crates/world/flashblocks/src/builder/mod.rs
0xForerunner: It would be better to change the OpPayloadBuilderCtx implementation if we want to go that route.

crates/world/node/src/args.rs
```diff
@@ -24,104 +33,167 @@ pub struct WorldChainArgs {
...
+    pub authorizer_vk: Option<VerifyingKey>,
 
-    // TODO: Make this optional
+    /// Flashblocks signing key
+    /// used to sign authorized flashblocks payloads.
     #[arg(
         long = "flashblocks.builder_sk", 
         env = "FLASHBLOCKS_BUILDER_SK", 
-        value_parser = parse_sk,
+        requires = "builder.enabled",
```
0xOsiris, line 179: Shouldn't this be `flashblocks.enabled`?

crates/world/node/src/args.rs
```diff
@@ -24,104 +33,167 @@ pub struct WorldChainArgs {
...
 }
 
-impl TryFrom<WorldChainArgs> for WorldChainNodeConfig {
-    type Error = eyre::Report;
-
-    fn try_from(args: WorldChainArgs) -> Result<Self, Self::Error> {
+impl WorldChainArgs {
+    pub fn into_config(mut self, spec: &OpChainSpec) -> eyre::Result<WorldChainNodeConfig> {
         // Perform arg validation here for things clap can't do.
+        match spec.chain.named() {
```
0xOsiris, line 39: Nice, this is sweet

0xOsiris (approved):

crates/world/node/src/args.rs
0xForerunner: Nope. We need the builder.private_key if we are going to have the flashblocks.builder_sk. Doesn't make sense to have the 2nd without the first. You can verify that this is correct with the tests below.

0xOsiris: The `builder.private_key` is specific to PBH for back running the block with a spend nullifiers transaction. This is independent of flashblocks. It would probably make more sense to consolidate all `builder` args onto `pbh`

0xForerunner: I don't think so.

You will never have flashblocks.builder_sk _without_ builder.enabled.
However you might have flashblocks.enabled without flashblocks.builder_sk or builder.enabled (rpc nodes for example)

PR #306, opened 2025-09-16
Osiris/remove in memory state implementation

0xOsiris:
Sends the pending block to the `FlashblocksEthApi` through a watch channel rather than the `CanonicalInMemoryState` as this state is crucial for persistence of blocks to the database.


0xForerunner (approved):
LGTM! I suppose we can get rid of the Canonical State provider trait then?

PR #309, opened 2025-09-18
feat: Flashblocks metrics

cichaczem:


0xOsiris (approved):

crates/flashblocks/payload/src/job.rs
```diff
@@ -150,6 +153,16 @@ where
         let flashblock = Flashblock::new(payload, self.config.clone(), self.block_index, offset);
         trace!(target: "payload_builder", id=%self.config.payload_id(), "creating authorized flashblock");
 
+        let flashblock_bytes = bincode::serialize(&flashblock)
```
0xForerunner, line 156: We shouldn't be serializing this with bincode just for a metric, directly in the publish_payload pipeline.

crates/flashblocks/payload/src/job.rs
```diff
@@ -283,6 +320,33 @@ where
                 Poll::Ready(Err(error)) => {
                     // job failed, but we simply try again next interval
                     debug!(target: "payload_builder", %error, "payload build attempt failed");
+                    match &error {
```
0xForerunner, line 323: I'm not sure these error metrics are useful.

PR #311, opened 2025-09-18
feat: flashblocks metrics

0xForerunner:


crates/flashblocks/payload/src/job.rs
```diff
@@ -297,7 +286,6 @@ where
                             this.metrics.inc_p2p_publishing_errors();
                             error!(target: "payload_builder", %err, "failed to publish new payload to p2p network");
                         } else {
-                            this.metrics.inc_preconfirmations_produced();
```
0xOsiris, line 300: This is a high signal metric.. this will help us infer appropriate intervals as we scale gas limits. I think we should keep this

crates/flashblocks/p2p/src/protocol/handler.rs
```diff
@@ -556,6 +557,19 @@ impl FlashblocksP2PCtx {
                 );
             }
 
+            if let Some(flashblock_timestamp) = payload.metadata.flashblock_timestamp {
+                let latency = Utc::now()
+                    .timestamp_nanos_opt()
+                    .expect("time went backwards")
+                    - flashblock_timestamp;
+                metrics::histogram!("flashblocks_latency")
```
0xOsiris: This is measuring latency between the instantiation of the struct `Flashblock` & the entry to `publish`.. wouldn't this almost be instant? What are we measuring the latency of here?

Shouldn't we be recording this metric when the flashblock is actually received over p2p?

crates/flashblocks/payload/src/job.rs
0xForerunner: okay I'll add back.

crates/flashblocks/p2p/src/protocol/handler.rs
0xForerunner: Yes this will read instantly for the builder, however for consuming nodes it will be very important.

crates/flashblocks/p2p/src/protocol/handler.rs
```diff
@@ -570,7 +584,15 @@ impl FlashblocksP2PCtx {
                     "publishing flashblock"
                 );
                 self.flashblock_tx.send(flashblock_event.clone()).ok();
-                // Update the index
+
+                // Don't measure the interval at the block boundary
+                if state.flashblock_index != 0 {
+                    let interval = now - state.flashblock_timestamp;
+                    histogram!("flashblocks_interval").record(interval as f64 / 1_000_000_000.0);
```
0xOsiris, line 591: Nit, but this is measuring the interval including latency over p2p. `this.metrics.inc_preconfirmations_produced` is a strict measurement on execution time. They are slightly different but both useful

0xForerunner: If we're the builder, then there won't be any latency.

PR #313, opened 2025-09-18
chore: prepend all tracing targets with flashblocks

0xOsiris:


crates/flashblocks/builder/src/lib.rs
```diff
@@ -260,13 +260,13 @@ where
         let Self { best } = self;
         let span = span!(
             tracing::Level::INFO,
-            "flashblock_builder",
+            "flashblocks::flashblock_builder",
```
0xForerunner: This isn't a target.

0xForerunner (changes requested):

0xForerunner (approved):

PR #314, opened 2025-09-19
chore: increase runner size

0xOsiris:


0xForerunner:
You can use arc-2xlarge-amd64-runner

PR #318, opened 2025-09-21
feat(flashblocks): override op_supportedCapabilities rpc module

0xOsiris:
Adds `op_supportedCapabilities` for `flashblocks` compliant nodes.

Ref: https://github.com/flashbots/rollup-boost/blob/main/specs/flashblocks.md#op_supportedcapabilities


crates/flashblocks/rpc/src/op.rs
```diff
@@ -0,0 +1,24 @@
...
+#[async_trait]
+pub trait OpApiExt {
+    /// Method to get supported capabilities
+    #[method(name = "supportedCapabilities")]
+    async fn supported_capabilities(&self) -> RpcResult<Vec<String>>;
+}
+
+#[async_trait]
+impl OpApiExtServer for FlashblocksOpApi {
+    async fn supported_capabilities(&self) -> RpcResult<Vec<String>> {
```
karankurbur: can we do this sync?

PR #322, opened 2025-09-25
minor cleanup

0xForerunner:


crates/world/bin/src/main.rs
```diff
@@ -19,7 +20,11 @@ fn main() {
     dotenvy::dotenv().ok();
 
     reth_cli_util::sigsegv_handler::install();
-    eyre::install().unwrap();
+
+    HookBuilder::default()
```
0xOsiris, line 24: Want to add to the flashblocks bin as well?

0xOsiris (approved):

PR #323, opened 2025-09-30
Install cast

0xForerunner:


Dockerfile
```diff
@@ -64,6 +67,9 @@ RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "/tmp/aws
 ARG WORLD_CHAIN_BUILDER_BIN="world-chain"
 COPY --from=builder /app/target/maxperf/${WORLD_CHAIN_BUILDER_BIN} /usr/local/bin/
 
+COPY --from=builder /usr/local/cargo/bin/cast /usr/local/bin/
+RUN cast --version
```
0xOsiris, line 71: Do we need this?

0xForerunner: We do yeah. cast --version is to just make sure the command is available.

PR #328, opened 2025-10-04
feat: concurrent tx pool re-commitment

0xOsiris:


crates/flashblocks/cli/src/lib.rs
```diff
@@ -51,6 +51,37 @@ pub struct FlashblocksArgs {
         required = false
     )]
     pub spoof_authorizer: bool,
+
+    /// The interval to publish pre-confirmations
+    /// when building a payload in milliseconds.
+    #[arg(
+        long = "flashblocks.interval",
+        env = "FLASHBLOCKS_INTERVAL",
+        requires = "authorizer",
```
0xForerunner: should probably be `requires = "builder_sk"` I think.

crates/flashblocks/cli/src/lib.rs
```diff
@@ -51,6 +51,37 @@ pub struct FlashblocksArgs {
...
+    pub recommit_interval: u64,
+
+    /// The maximum number of concurrent payload build tasks.
+    #[arg(
+        long = "flashblocks.max_payload_tasks",
+        env = "FLASHBLOCKS_MAX_PAYLOAD_TASKS",
+        requires = "authorizer",
+        default_value_t = 50
+    )]
+    pub max_payload_tasks: usize,
```
0xForerunner: Why not use the regular max_payload_tasks arg? Is there ever a situation where they should be different?

crates/flashblocks/payload/src/generator.rs
```diff
@@ -333,14 +345,28 @@ impl FlashblocksJobGeneratorConfig {
...
 }
 
 impl Default for FlashblocksJobGeneratorConfig {
     fn default() -> Self {
         Self {
             interval: Duration::from_millis(200),
+            recommitment_interval: Duration::from_millis(200),
             deadline: Duration::from_secs(2),
             enable_authorization: true,
+            max_payload_tasks: 50,
```
0xForerunner: I'm not sure but isn't 50 way too high? I would think we wouldn't want more tasks than we expect to have threads? Maybe 4 is a sane default?

0xForerunner (commented):
Looks great man. Just a few comments.

crates/flashblocks/cli/src/lib.rs
0xOsiris: I just wanted to isolate our CLI args from reth's `builder` args since it may be confusing. We are not using any reth `builder` args anymore

0xForerunner: I think it might be more confusing to have 2 flags with similar names, and just ignore one of them. I think we should use the value from the original fIag IMO.

PR #332, opened 2025-10-06
feat: cache built payloads from the state executor

0xOsiris:


0xForerunner (approved):
Nice this looks sick.

PR #440, opened 2026-03-06
Bring back healthcheck feature

piohei:
This is not only revert but also includes other changes.

---

> [!NOTE]
> **Medium Risk**
> Adds a new optional HTTP server with background probe evaluation and disk/network/provider checks; while gated behind `--health.enabled`, it introduces new runtime tasks and an exposed endpoint that could affect operations if misconfigured.
> 
> **Overview**
> Reintroduces an *optional* healthcheck HTTP server for World Chain nodes, enabled via new CLI flags (`--health.enabled`, `--health.addr`, `--health.port`, `--health.config`) and started from `crates/world/bin` using the node’s provider/network handles.
> 
> Adds a `world_chain_node::health` module backed by `axum` with `/startup`, `/ready`, and `/live` endpoints returning cached probe results, plus a JSON-configurable set of checks (peer count, syncing status, block progress/timestamp freshness, and disk space via `statvfs`). Updates tests/config constructors to include default `health` args, wires in new deps (`axum`, `nix`, `serde`), and documents configuration/usage in `specs/health_checks.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 69a5ff65c0429b8c83290477429ccd228b1912bc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/world/bin/src/main.rs
+++ b/crates/world/bin/src/main.rs
@@ -45,12 +45,12 @@ fn main() {
             let config: WorldChainNodeConfig = args.into_config(builder.config_mut())?;
 
             info!(target: "reth::cli", "Starting in Flashblocks mode");
-            let node = WorldChainNode::<FlashblocksContext>::new(config.clone());
+            let worldchain_node = WorldChainNode::<FlashblocksContext>::new(config.clone());
             let NodeHandle {
                 node_exit_future,
-                node: _node,
+                node,
             } = builder
-                .node(node)
+                .node(worldchain_node)
                 .extend_rpc_modules(move |ctx| {
                     let provider = ctx.provider().clone();
                     let pool = ctx.pool().clone();
--- a/crates/world/bin/src/main.rs
+++ b/crates/world/bin/src/main.rs
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+
 use clap::Parser;
 use eyre::config::HookBuilder;
 use reth_node_builder::NodeHandle;
@@ -62,13 +64,15 @@ fn main() {
                 .launch()
                 .await?;
 
-            if let Some(addr) = config.args.health.addr {
+            if config.args.health.enabled {
+                let health_addr =
+                    SocketAddr::new(config.args.health.addr, config.args.health.port);
                 let health_config = match &config.args.health.config {
                     Some(path) => HealthConfig::from_file(path)?,
                     None => HealthConfig::default(),
                 };
                 let health_server =
-                    health_config.build(addr, node.provider.clone(), node.network.clone());
+                    health_config.build(health_addr, node.provider.clone(), node.network.clone());
                 node.task_executor.spawn(health_server.serve());
             }
 
--- a/crates/world/bin/src/main.rs
+++ b/crates/world/bin/src/main.rs
@@ -65,8 +65,7 @@ fn main() {
                 .await?;
 
             if config.args.health.enabled {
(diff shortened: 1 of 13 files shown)


alessandromazza98 (approved):
🚢

crates/world/bin/src/main.rs
```diff
@@ -61,6 +63,18 @@ fn main() {
...
+            if config.args.health.enabled {
+                let health_addr = SocketAddr::new(config.args.health.addr, config.args.health.port);
+                let health_config = match &config.args.health.config {
+                    Some(path) => HealthConfig::from_file(path)?,
+                    None => HealthConfig::default(),
+                };
+                let health_server =
+                    health_config.build(health_addr, node.provider.clone(), node.network.clone());
+                node.task_executor.spawn(health_server.serve());
+            }
```
0xForerunner, line 76: I don't think this is the right place to init this. We should do it when actually constructing WorldChainNode, not in main. This won't init for tests.

crates/world/node/src/health/config.rs
```diff
@@ -0,0 +1,275 @@
...
+    },
+    BlockTimestamp {
+        #[serde(default = "defaults::max_age_secs")]
+        max_age_secs: u64,
+    },
+    DiskSpace {
+        path: PathBuf,
+        #[serde(default = "defaults::min_gb")]
+        min_gb: f64,
+    },
```
0xForerunner, line 62: What is the actionable purpose of this check? We can already setup dd based alerts.

crates/world/node/src/health/config.rs
```diff
@@ -0,0 +1,275 @@
...
+    BlockProgress {
+        #[serde(default = "defaults::period_secs")]
+        period_secs: u64,
+        #[serde(default)]
+        mode: BlockNumberMode,
+    },
+    BlockTimestamp {
+        #[serde(default = "defaults::max_age_secs")]
+        max_age_secs: u64,
+    },
```
0xForerunner, line 57: I don't think these checks quite make sense. In order to know if a node should accept traffic we really need to be comparing against other similar nodes. To me this seems like a job for proxyd.

crates/world/node/src/args.rs
```diff
@@ -202,6 +210,38 @@ pub struct PbhArgs {
...
+    pub addr: IpAddr,
+
+    /// Port for the health HTTP server.
+    #[arg(long = "health.port", id = "health.port", default_value_t = 8080)]
+    pub port: u16,
+
+    /// Path to health check configuration file (JSON).
+    /// If omitted, all probes return 200 OK unconditionally.
+    #[arg(long = "health.config", id = "health.config", required = false)]
+    pub config: Option<PathBuf>,
```
0xForerunner, line 231: I don't think we want a separate config file for this. We should config via cli or a global single config file.

0xForerunner (changes requested):
In general I don't really think this approach is the right way to go. I think the only way to solve our problems robustly here is to use an external service that can talk to all connected nodes and can "fail open". We should probably use proxyd, and add any additional functionality we want there.

0xOsiris:
What's the status of this? Are we good to close? @0xForerunner @piohei

piohei:
AFAIR we agreed we don't want it in world-chain. I forget to close it.

PR #444, opened 2026-03-07
Osiris/fix pending rpc

0xOsiris:


crates/flashblocks/rpc/src/eth/filter.rs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            let from_pending = from_block.is_some_and(|b| b.is_pending());
+            let to_pending = to_block.is_some_and(|b| b.is_pending());
+
+            if from_pending || to_pending {
+                if let Some(logs) = self.pending_flashblock_logs(&filter) {
+                    return Ok(logs);
+                }
+            }
+        }
```
0xForerunner, line 170: Does this also intercept latest..pending? If so, we only return the pending logs here and drop the historical part of the range.

0xForerunner (changes requested):

crates/flashblocks/rpc/src/eth/filter.rs
0xOsiris: ah, you're right nice catch!

crates/world/node/tests/e2e-testsuite/testsuite.rs
```diff
@@ -1170,3 +1264,394 @@ async fn test_continuous_block_production_with_validation() -> eyre::Result<()>
...
+/// This directly tests the PROTO-4338 bug where all polls returned the same
+/// full block's logs instead of the current flashblock's accumulated logs.
+///
+/// Setup: 2 nodes, flashblocks enabled, tx spammer active
+/// 1. Start mining a block with flashblocks
+/// 2. Poll eth_getLogs("pending") multiple times during flashblock production
+/// 3. Record log counts at each poll
+/// 4. Assert that log counts grow over time as flashblocks include more txs
+#[tokio::test(flavor = "multi_thread")]
+async fn test_eth_get_logs_incremental_growth() -> eyre::Result<()> {
```
alessandromazza98, line 1413: it doesn't seem to me that this test is ensuring that log counts grow over time. We only test once that count > 0

crates/world/node/tests/e2e-testsuite/testsuite.rs
```diff
@@ -1170,3 +1264,394 @@ async fn test_continuous_block_production_with_validation() -> eyre::Result<()>
...
+/// This verifies that eth_getBlockByNumber("pending") returns an incrementally
+/// updated pending block as new flashblocks are published.
+///
+/// Setup: 2 nodes, flashblocks enabled, tx spammer active
+/// 1. Start mining a block with flashblocks
+/// 2. Poll eth_getBlockByNumber("pending") multiple times
+/// 3. Record transaction counts at each poll
+/// 4. Assert tx counts grow over time as more flashblocks include more txs
+#[tokio::test(flavor = "multi_thread")]
+async fn test_pending_block_growth() -> eyre::Result<()> {
```
alessandromazza98, line 1528: similar in this test. It seems to me that we are just testing that in the end there is at least 1 transaction

alessandromazza98 (approved):
LGTM. One note is that it doesn't seem to me that tests are ensuring proper behaviour.

Question: does this fix the `pending` problem for other queries such as `eth_getBlockByNumber` ?

crates/world/bin/src/main.rs
```diff
@@ -57,6 +58,15 @@ fn main() {
...
                     ctx.modules
                         .replace_configured(FlashblocksOpApi.into_rpc())?;
+
+                    // Replace the default EthFilter with FlashblocksEthFilter to handle
+                    // pending flashblock logs in eth_getLogs
+                    let eth_filter = ctx.registry.eth_handlers().filter.clone();
+                    let pending_block = ctx.registry.eth_api().pending_block_receiver();
+                    let flashblocks_filter = FlashblocksEthFilter::new(eth_filter, pending_block);
+                    ctx.modules
+                        .replace_configured(flashblocks_filter.into_rpc())?;
```
0xForerunner, line 68: Since we're only setting this up in main, we're not actually initializing any of this in the tests. We should be doing this at the `WorldChainNode<FlashblocksContext>` level.

0xForerunner (changes requested):

PR #445, opened 2026-03-07
wip: mapped stream validation handler over event multiplexor

0xOsiris:


crates/flashblocks/rpc/src/eth/filter.rs
```diff
@@ -0,0 +1,173 @@
...
+        // Check if this is a pending block range query
+        if let FilterBlockOption::Range {
+            from_block,
+            to_block,
+        } = &filter.block_option
+        {
+            let from_pending = from_block.is_some_and(|b| b.is_pending());
+            let to_pending = to_block.map_or(false, |b| b.is_pending());
+
+            if from_pending || to_pending {
```
0xForerunner, line 161: Does this also intercept `latest..pending`? If so, we only return the pending logs here and drop the historical part of the range.

alessandromazza98:
can we close this @0xOsiris ?

PR #487, opened 2026-04-03
wip: (1003) World ID Transaction Subsidies

0xForerunner:


wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+- `virtualGasLimit`
+
+These values are out of protocol. They control transaction admission and ordering in the World Chain client, but they do not affect block validity at the protocol level.
+
+`virtualGasLimit` is the builder-local effective gas limit used for transaction selection and verified blockspace segmentation.
+`virtualGasTarget` is the target used by the virtual fee controller when deriving the next `virtualBaseFee`.
+
+This WIP assumes the virtual fee parameters are made deterministic by a system contract, referred to here as `VirtualFeeOracle`.
+
+- A builder-owned transaction at the end of block `N` writes the virtual fee parameters that apply to block `N + 1`.
```
Kemperino, line 55: does anything speak against just adding it to the existing system transactions at the top of the block?

0xForerunner: Yeah I don't see any reason why we can't do this! It would certainly be cleaner!

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+- a real fee market for non-verified traffic
+- fully subsidized transactions for verified humans up to a bounded limit
+
+PBH solves priority and fairness, but it does not provide a unified fee model for free verified transactions and market-priced non-verified transactions. This WIP makes the subsidy model native to normal transaction inclusion. Verified transactions can be literally free up to a per-epoch wei limit, while the rest of the chain still sees a fee floor and blockspace competition.
+
+## Proposal
+
+### 1. Protocol Fee Market
+
+- The protocol gas target and protocol gas limit are set arbitrarily high, so the protocol base fee does not rise above the protocol minimum base fee.
```
0xOsiris, line 33: You should be more precise here. Why would the gas limit be set arbitrarily high? That would have significant impact on the priority fee market.

I think you probably mean `target == gas_limit`

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+- After that, the protocol `baseFeePerGas` is always `0`.
+- This WIP does not introduce a new transaction type.
+
+### 2. Virtual Fee Market
+
+Builders maintain a virtual fee market with the following values:
+
+- `virtualBaseFee`
+- `virtualMinBaseFee`
+- `virtualGasTarget`
```
0xOsiris, line 45: Why do we need the target and limit to be virtual? It's possible today to just set the target equal to the gas limit by setting the EIP-1559 denominator to 1

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+- `maxPriorityFeePerGas` must be at least `virtualBaseFee`
+- any amount above `virtualBaseFee` is an optional tip
+
+For a subsidy-eligible transaction, the required priority fee may be reduced by the account's remaining subsidy.
+
+### 4. Transaction Validation
+
+The World Chain transaction validator is extended to enforce the virtual fee floor for in-scope transactions.
+
+- If a transaction is not subsidy-eligible, it is invalid unless it pays at least the virtual fee floor through `maxPriorityFeePerGas`.
```
0xOsiris, line 74: How will this be differentiated? Or is that out of scope

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+### 5. Subsidy Eligibility
+
+A transaction is subsidy-eligible if either of the following is true:
+
+- the sender is verified in the address-book for the current epoch
+- the transaction is a valid address-book registration transaction that verifies a specific address for the current epoch or next epoch
+
+For address-book registration transactions:
+
+- the subsidy is attributed to the address being verified, not necessarily `msg.sender`
```
0xOsiris, line 92: I think this is way too restrictive

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+Within the verified segment, transactions may still compete via optional tip.
+
+### 8. Address-Book Updates
+
+The address-book is extended to track subsidy usage per address per epoch.
+
+At minimum it must support:
+
+- checking whether an address is verified for the current epoch
+- tracking total subsidized wei used by an address for a given epoch
```
0xOsiris, line 136: This could be tracked out of the address book. Only the builder actually needs to track this

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+5. checks that the sender-paid priority fee covers `virtualBaseFeeLiability - actualSubsidy`
+6. updates the charged account by `actualSubsidy`
+
+Within the verified segment, transactions may still compete via optional tip.
+
+### 8. Address-Book Updates
+
+The address-book is extended to track subsidy usage per address per epoch.
+
+At minimum it must support:
```
0xOsiris, line 133: I would add support non-4337 paths as well as multiple addresses

wips/wip-1003.md
```diff
@@ -0,0 +1,200 @@
...
+
+- `eth_maxPriorityFeePerGas`
+- `eth_gasPrice`
+- `eth_feeHistory`, where needed for wallet compatibility
+
+`eth_estimateGas` continues to estimate gas usage, not fees.
+
+For ordinary transactions, overriding these fee RPCs is sufficient for compatibility with wallets that already rely on them.
+
+Wallets that derive fees directly from the block header `baseFeePerGas` may underquote, because the protocol header value remains `0`. Wallets integrating World Chain should therefore prefer the overridden fee RPCs rather than raw block-header base fee.
```
0xOsiris, line 163: This will break indexers, and explorers as well. Since the block header no longer commits to the EIP-1559 params

wips/wip-1003.md
alessandromazza98: modifying the 1st system tx would require a change in the op-node or kona node binaries though

wips/wip-1003.md
alessandromazza98: I'd say it's important to know how it's gonna be differentiated. It should go into scope of this WIP imo

wips/wip-1002.md
```diff
@@ -0,0 +1,200 @@
...
+3. computes `virtualBaseFeeLiability = gasUsed * virtualBaseFee`
+4. computes `actualSubsidy = min(remainingSubsidy, virtualBaseFeeLiability)`
+5. checks that the sender-paid priority fee covers `virtualBaseFeeLiability - actualSubsidy`
+6. updates the charged account by `actualSubsidy`
+
+Within the verified segment, transactions may still compete via optional tip.
+
+### 8. Address-Book Updates
+
+The address-book is extended to track subsidy usage per address per epoch.
```
alessandromazza98: nit: we're speaking of an address book in this WIP (and others), but we do not have it clearly defined anywhere. We do have an address book now in `world-id-utility-contracts` repo but that serves a different purpose. I'd suggest to create a new name for this new type of contract. Instead, if we want to use the same address book we already have and extend it, then that's another thing and we can probably discuss.

wips/wip-1003.md
alessandromazza98: I agree

wips/wip-1003.md
alessandromazza98: I wouldn't phrase it as "it breaks indexers and explorers". It's more that indexers and explorers will not show the full picture as it would always seem that the gas limit is huge compared to actual gas used. They would need to read the custom virtual base fee data to have the full picture.

alessandromazza98 (commented):
I think what we need to take into account is that this WIP requires a change to the way TFH transactions are sent. Right now we use erc-4337 flow where the msg.sender is the bundler. This WIP instead changes that such that the msg.sender can actually be the real user that wants to send that tx onchain.

So we also need to take into account that mini app developers should change the way they usually create transactions for their users

0xOsiris:
@0xForerunner Are we good with closing this in favor of #511 

I'm good with keeping it since it's a different spec

0xForerunner:
@0xOsiris yeah we should keep this open as it's a different topic.

wips/wip-1003.md
0xForerunner: Why do you say that. I think we can expand this later with native AA, but IMO this is perfectly fine for an MVP.

wips/wip-1003.md
0xForerunner: I do not know what you mean here. This needs to be tracked somewhere on chain.

wips/wip-1003.md
0xForerunner: This version of the spec I think should just be non-4337.

0xOsiris (approved):

PR #492, opened 2026-04-08
feat: add metrics to flashblocks p2p pipeline

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4470/add-metrics-to-flashblocks-p2p-pipeline

Here is an example of our local devnet grafana dashboard showing these new metrics for the flashblocks p2p pipeline.

Screenshot 2026-04-08 at 10 40 15 AM

### Next Step

- add metrics for payload building pipeline with BAL
- create datadogs dashboards for dev/stage/prod

---

> [!NOTE]
> **Medium Risk**
> Touches flashblocks P2P connection/state paths to add peer-scoped metrics and to rebind labels when trust classification resolves, which could affect runtime behavior/overhead despite being mostly observability-focused.
> 
> **Overview**
> **Adds first-class flashblocks P2P observability.** Introduces `FlashblocksP2PMetrics` + `FlashblocksPeerMetrics` (via `metrics-derive`) and replaces ad-hoc `metrics::*` calls with typed helpers to track connected peers, stale epochs, flashblock size/gas/tx count/interval, and per-peer bandwidth/latency/score/missed blocks.
> 
> **Changes peer state/trust handling to support labeled metrics.** `FlashblocksPeerState` now owns a metrics handle that is rebound when `set_trusted()` resolves trust (trusted peers use concrete `peer_id` labels; untrusted share a single label), and the event stream reducer now receives metrics to count stale epochs.
> 
> **Dev tooling updates.** Adds a new `flashblocks-p2p.json` Grafana dashboard, updates devnet docs/`Justfile` to point at the shared `scripts/stress/stress.sh`, and rewrites the stress script to auto-resolve endpoints from Kurtosis with sane fallbacks and stricter bash settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f1700b136dc7ce68007a24e0052e2ad65f802bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/p2p/src/protocol/connection.rs
+++ b/crates/p2p/src/protocol/connection.rs
@@ -1,10 +1,10 @@
 use crate::protocol::handler::{
     FlashblocksP2PNetworkHandle, FlashblocksP2PProtocol, MAX_FLASHBLOCK_INDEX, PublishingStatus,
 };
+use crate::protocol::metrics::FlashblocksPeerMetrics;
 use alloy_primitives::bytes::BytesMut;
 use chrono::Utc;
 use futures::{Stream, StreamExt};
-use metrics::gauge;
 use reth_ethereum::network::{api::PeerId, eth_wire::multiplex::ProtocolConnection};
 use reth_network::types::ReputationChangeKind;
 use world_chain_primitives::{
@@ -48,6 +48,8 @@ pub enum ReceiveStatus {
 /// Shared connection metadata for a single peer connection.
 #[derive(Clone, Debug)]
 pub struct FlashblocksPeerState {
+    /// Metric handles bound to this peer's stable label set.
+    pub metrics: FlashblocksPeerMetrics,
     /// Whether this peer is marked as trusted or not.
     pub trusted: bool,
     /// Whether we are currently sending flashblocks to this peer.
@@ -66,8 +68,9 @@ pub struct FlashblocksPeerState {
 }
 
 impl FlashblocksPeerState {
-    pub(crate) fn new() -> Self {
+    pub(crate) fn new(peer_id: PeerId) -> Self {
         Self {
+            metrics: FlashblocksPeerMetrics::for_peer(peer_id),
             trusted: false,
             send_enabled: false,
             receive_status: ReceiveStatus::NotReceiving,
@@ -75,14 +78,6 @@ impl FlashblocksPeerState {
             outbound_tx: None,
             control_msg_count: 0,
             control_msg_window_start: Instant::now(),
-        }
-    }
-
-    pub(crate) fn peer_id_labels(&self, peer_id: PeerId) -> Vec<(&'static str, String)> {
-        if self.trusted {
-            vec![("peer_id", format!("{:#x}", peer_id))]
-        } else {
-            vec![("peer_id", "untrusted".to_string())]
         }
     }
 }
@@ -124,8 +119,6 @@ impl<N: FlashblocksP2PNetworkHandle> FlashblocksConnection<N> {
             .handle
             .on_peer_connected(protocol.network.clone(), peer_id, outbound_tx);
 
-        gauge!("flashblocks.peers", "capability" => FlashblocksP2PProtocol::<N>::capability().to_string()).increment(1);
-
         Self {
             protocol,
             conn,
@@ -144,8 +137,6 @@ impl<N> Drop for FlashblocksConnection<N> {
         );
 
         self.protocol.handle.on_peer_disconnected(self.peer_id);
-
-        gauge!("flashblocks.peers", "capability" => FlashblocksP2PProtocol::<N>::capability().to_string()).decrement(1);
     }
 }
 
@@ -364,10 +355,8 @@ impl<N: FlashblocksP2PNetworkHandle> FlashblocksConnection<N> {
             return;
         };
 
-        let peer_id_labels = peer_state.peer_id_labels(self.peer_id);
-
-        metrics::counter!("flashblocks.bandwidth_inbound", &peer_id_labels)
-            .increment(buf_len as u64);
+        let peer_metrics = peer_state.metrics.clone();
+        peer_metrics.record_inbound_bandwidth_bytes(buf_len);
 
         match &peer_state.receive_status {
             ReceiveStatus::Requesting => {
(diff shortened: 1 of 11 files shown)


pkg/devnet/grafana/dashboards/flashblocks-p2p.json
```diff
@@ -0,0 +1,1277 @@
...
+        },
+        "showPercentChange": false,
+        "textMode": "auto",
+        "wideLayout": true
+      },
+      "pluginVersion": "12.4.0",
+      "targets": [
+        {
+          "editorMode": "code",
+          "expr": "sum(increase(reth_flashblocks_p2p_tx_count_sum{instance=~\"$node\"}[$__rate_interval])) / clamp_min(sum(increase(reth_flashblocks_p2p_tx_count{instance=~\"$node\"}[$__rate_interval])), 1)",
```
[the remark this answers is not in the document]
alessandromazza98: that's incorrect. Since the metric name already ends with `count`, there's no need to add another `count`. I was doing that initially and it was showing `No Data`. Then I changed to what you can see right now and it correctly shows data.

crates/p2p/src/protocol/metrics.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub flashblock_gas_used: Histogram,
+    /// Histogram of transaction counts per broadcast flashblock.
+    #[metric(rename = "tx_count")]
+    pub flashblock_tx_count: Histogram,
+    /// Histogram of time between sequential flashblocks in seconds.
+    #[metric(rename = "interval")]
+    pub flashblock_interval_seconds: Histogram,
+}
+
+impl FlashblocksP2PMetrics {
```
0xForerunner, line 31: These setters feel a bit unnecessary

pkg/devnet/grafana/dashboards/flashblocks-p2p.json
0xForerunner, line 1: Are we sure we want to commit to maintaining grafana dashboards in our repo?

crates/p2p/src/protocol/connection.rs
```diff
@@ -77,14 +82,6 @@ impl FlashblocksPeerState {
...
         }
     }
-
-    pub(crate) fn peer_id_labels(&self, peer_id: PeerId) -> Vec<(&'static str, String)> {
-        if self.trusted {
-            vec![("peer_id", format!("{:#x}", peer_id))]
-        } else {
-            vec![("peer_id", "untrusted".to_string())]
-        }
-    }
```
0xForerunner, line 87: Removing this logic means you will get extremely high cardinality. I think we need to revert back to this.

0xForerunner (changes requested):

pkg/devnet/grafana/dashboards/flashblocks-p2p.json
alessandromazza98: I think yes, these are helpful for local devnet only, so not a big deal

crates/p2p/src/protocol/metrics.rs
alessandromazza98: I think they're just better than every time calling `increment(1.0)` or similar. I took it from reth as reference

crates/p2p/src/protocol/connection.rs
alessandromazza98: yeah I see the point, I can add this back. I was thinking that our nodes being 3 will never have lots of peers connected nevertheless. But I can add this back and we may change this in the future

crates/p2p/src/protocol/metrics.rs
0xForerunner: Most of them are only used a single time though

pkg/devnet/grafana/dashboards/flashblocks-p2p.json
kilianglas: I'd even say that we could think about adding other dashboards as well. I kinda like the idea that other people can just run a world chain node and get the corresponding dashboards from the repo.

crates/p2p/src/protocol/connection.rs
kilianglas: What about making this configurable? Maybe there are scenarios where we actually want this, e.g., for fanout/ proxy nodes that forward flashblocks to external consumers

0xForerunner: I don't think you ever want 100+ tag values ahah. IMO it's always very bad practice to have tags for which you don't control all the values.

pkg/devnet/grafana/dashboards/flashblocks-p2p.json
alessandromazza98: that can be done in the `simple-worldchain-node` repo that is what external people use to run a node. There are already some grafana dashboards there. I can add some dashboard there as well for flashblocks

crates/p2p/src/protocol/metrics.rs
alessandromazza98: that's true, but I still think it's better than having the underlying `increment(1.0)` which is less maintainable. In terms of performance there is basically no difference because the compiler is probably inlining all these fns

crates/p2p/src/protocol/connection.rs
alessandromazza98: done

crates/p2p/src/protocol/connection.rs
```diff
@@ -448,15 +449,13 @@ impl<N: FlashblocksP2PNetworkHandle> FlashblocksConnection<N> {
...
-            metrics::histogram!("flashblocks.latency", &peer_id_labels).record(latency_secs);
+            peer_metrics.record_flashblock_latency_seconds(latency_secs);
             if let Some(peer_state) = p2p_state.connection_state_mut(&self.peer_id)
                 && let ReceiveStatus::Receiving { score } = &mut peer_state.receive_status
             {
                 score.record(latency);
-                if let Some(value) = score.value()
-                    && peer_state.trusted
-                {
-                    metrics::gauge!("flashblocks.peer_score", &peer_id_labels).set(value as f64);
```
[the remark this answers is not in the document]
alessandromazza98: fixed

0xForerunner (approved):

PR #493, opened 2026-04-08
feat: add metrics to BAL builder / validator

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4469/add-payload-builder-validator-metrics-to-bal-builder

This PR adds metrics to the BAL (block access list) payload builder and validator such that it's equal to the flashblock payload builder / validator.

---

> [!NOTE]
> **Medium Risk**
> Primarily adds timing instrumentation, but it also adjusts finalize sequencing in the parallel BAL validator (database index finalization and error handling), which could subtly impact execution correctness or performance.
> 
> **Overview**
> Adds *per-stage timing metrics* to the BAL payload builder and parallel validator, recording durations for `MergeTransitions`, `StateRoot`, `BlockAssembly`, `PreExecutionChanges`, `SequencerTxExecution`, and `Finalize` via `PayloadBuildStage`.
> 
> Refactors parts of block finalization to support this instrumentation (e.g., making metrics mutable, moving database index finalization into `finish`, and simplifying error handling/ordering in `executor.rs`), and updates e2e fuzz tests to pass the new attempt-metrics reference into `validate_flashblock_parallel`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8886f1cde5428a6ef3b8d63be3fbebc404a4ddd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/builder/src/executor.rs
+++ b/crates/builder/src/executor.rs
@@ -184,14 +184,15 @@ where
         db.bundle_state_mut().reverts = flattened;
 
         // calculate the state root
-        let hashed_state = state.hashed_post_state(db.bundle_state());
         let state_root_started = Instant::now();
-        let state_root_result = state.state_root_with_updates(hashed_state.clone());
+        let hashed_state = state.hashed_post_state(db.bundle_state());
+        let (state_root, trie_updates) = state
+            .state_root_with_updates(hashed_state.clone())
+            .map_err(BlockExecutionError::other)?;
         if let Some(metrics) = metrics.as_mut() {
             metrics
                 .record_stage_duration(PayloadBuildStage::StateRoot, state_root_started.elapsed());
         }
-        let (state_root, trie_updates) = state_root_result.map_err(BlockExecutionError::other)?;
 
         let (transactions, senders) = self
             .inner
@@ -201,7 +202,7 @@ where
             .unzip();
 
         let block_assembly_started = Instant::now();
-        let block_result = self.inner.assembler.assemble_block(BlockAssemblerInput::<
+        let block = self.inner.assembler.assemble_block(BlockAssemblerInput::<
             '_,
             '_,
             OpBlockExecutorFactory<R>,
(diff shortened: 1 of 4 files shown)


crates/builder/src/executor.rs
```diff
@@ -184,14 +184,15 @@ where
...
-        let state_root_result = state.state_root_with_updates(hashed_state.clone());
+        let hashed_state = state.hashed_post_state(db.bundle_state());
+        let (state_root, trie_updates) = state
+            .state_root_with_updates(hashed_state.clone())
+            .map_err(BlockExecutionError::other)?;
         if let Some(metrics) = metrics.as_mut() {
             metrics
                 .record_stage_duration(PayloadBuildStage::StateRoot, state_root_started.elapsed());
         }
-        let (state_root, trie_updates) = state_root_result.map_err(BlockExecutionError::other)?;
```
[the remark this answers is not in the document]
alessandromazza98: this is intentional and it's correct. We don't care about backward compatibility now

crates/builder/src/executor.rs
```diff
@@ -214,14 +215,13 @@ where
             db.bundle_state(),
             &state,
             state_root,
-        ));
+        ))?;
```
piohei, line 218: Are you sure we don't want to collect that metric on fail?

crates/builder/src/validator.rs
```diff
@@ -548,6 +551,7 @@ pub struct BalBlockValidator<'a, DbRef: DatabaseRef + 'a, R: OpReceiptBuilder, E
     pub temporal_db_factory: &'a TemporalDbFactory<DbRef>,
     pub evm_env: EvmEnv<OpSpecId>,
     pub index_range: (u16, u16),
+    pub attempt_metrics: &'a mut FlashblockValidationAttemptMetrics,
```
piohei, line 554: `FlashblockExecutionMetric` is optional (I assume that it can be disabled). Do we want that metric also to be optional?

crates/builder/src/executor.rs
alessandromazza98: yes, it won't be tracked on failure even without this change. And this is intentional so that we don't get metrics for failed job (note that we do have a metric for failed jobs though)

crates/builder/src/validator.rs
alessandromazza98: no. That's optional (for now, I'll probably change it) just so that tests can use `None`.

piohei (approved):

PR #496, opened 2026-04-09
wip(1001): WorldID native account abstraction framework

kilianglas:
## Summary
- Add WIP-1004, which describes a framework for WorldID account abstraction
- More of a draft for now. Intentionally high level.


wips/wip-1001.md
```diff
@@ -0,0 +1,346 @@
...
+
+### WorldID Gas Accounting
+
+The accounting logic is implemented in a dedicated Gas Accounting component. Time is divided into **periods** (e.g., one month). Each WorldID gets a nullifier per period, and budget must be refreshed each period. There is a governance-configurable map from credentials (identified by `issuerSchemaId`) to claimable budget amounts, where each claim requires one WorldID proof.
+
+The component maintains an authorization map from authorized account addresses to nullifiers, and a gas budget map from nullifiers to remaining budget. These maps can be used to look up whether an address is an authorized account and, if so, to query and update its remaining budget.
+
+The gas accounting component can be deployed as either:
+
+- **A precompile**, enabling the protocol to natively read gas accounting state during transaction validation and execution. This is required if gas subsidies are enforced at the protocol level (e.g., the EVM itself checks and deducts budget).
```
Kemperino, line 51: i am feeling a bit uncomfortable with the idea of having a stateful precompile.

wips/wip-1001.md
```diff
@@ -0,0 +1,346 @@
...
+
+The gas budget is bound to a per-period nullifier derived from a WorldID proof:
+
+- Relying party is World Chain (e.g., RP ID `480`)
+- Action is `"period_proof" || periodNumber`
+- Signal contains the action (add or remove) and the list of authorized addresses
+
+Every WorldID proof natively carries an `issuerSchemaId` (mandatory public input in the circuit). The nullifier is credential-independent: the same action produces the same nullifier regardless of which credential is used. This means all credential proofs for a given WorldID in a given period map to the same nullifier.
+
+Period could be e.g., one month. It likely makes sense to divide the period into multiple slots with sub-budgets for each slot to prevent traffic spikes (e.g., when all users claim their monthly budget at once after a token launch).
```
Kemperino, line 125: arguably this is taken care of  by adjusting the budget based on gas price instead of hard gas limit, which was argued in favor of above?

otherwise we might as well just reduce period length?

wips/wip-1001.md
kilianglas: I don't think its a big deal. State is still in the storage trie, e.g., Tempo uses this all the time

wips/wip-1004.md
```diff
@@ -0,0 +1,346 @@
...
+
+### WorldID Gas Accounting
+
+World Chain needs a mechanism to subsidize gas for verified WorldID holders while preventing abuse. The following requirements drive the design:
+
+- **Per-credential gas budget.** Different credentials (e.g., Proof-of-Human, phone, NFC passport) should grant different budget amounts.
+- **Sybil resistance.** Each WorldID should only be able to claim a gas budget once per period.
+- **Anonymity.** The subsidy mechanism must not de-anonymize a WorldID. If a WorldID authorizes different accounts across periods, there should be no way to link them to the same identity.
+- **Periodic refresh.** The gas budget is not one-time; it refreshes periodically.
+- **Gas-price-based budget.** Budget is consumed based on `gasUsed * baseFee` at time of transaction, not raw gas units or transaction count, which could easily be abused.
```
0xOsiris: This is a good idea

wips/wip-1004.md
```diff
@@ -0,0 +1,346 @@
...
+- **Traffic classification.** For any transaction, it must be possible to determine whether it is authorized to consume a WorldID's gas budget.
+- **Multi-account.** A single WorldID should be able to authorize multiple accounts. Budget is shared across all authorized accounts.
+- **Authorization management.** Authorized accounts can be added and revoked over time. Revocation immediately stops an account from consuming the budget.
+- **Out-of-protocol usage.** The accounting machinery should be usable by out-of-protocol mechanisms, e.g., if gas subsidies are implemented at the builder level only.
+- **Account type agnostic (optional).** A WorldID should be able to authorize any account type, including legacy EOAs and smart accounts.
```
0xOsiris: I don't really understand what this means unless we are changing the definition of what a transaction authenticator is. The gas is always paid by an EOA, and we would I assume be delegating a shared subsidy to the wallet or many wallets

wips/wip-1004.md
```diff
@@ -0,0 +1,346 @@
...
+- **1559 support (optional).** Gas subsidies should be usable with standard EIP-1559 transactions to enable legacy third-party wallet support.
```
0xOsiris: I would make this a requirement

wips/wip-1001.md
0xOsiris, line 51: > This is required if gas subsidies are enforced at the protocol level (e.g., the EVM itself checks and deducts budget).

This statement is not correct. "preferred" is a better word

wips/wip-1004.md
kilianglas: My idea was, that this should work for existing EOA and the new world id account type

wips/wip-1001.md
kilianglas: Agreed

wips/wip-1001.md
```diff
@@ -0,0 +1,346 @@
...
+##### Gas Budget Map
+
+Maps nullifiers to remaining budget in Gwei. Budget accumulates as new credential proofs are submitted. Tracks which `issuerSchemaId` have been claimed per nullifier per period to prevent double-claims.
+
+#### Periods and Nullifiers
+
+The gas budget is bound to a per-period nullifier derived from a WorldID proof:
+
+- Relying party is World Chain (e.g., RP ID `480`)
+- Action is `"period_proof" || periodNumber`
```
0xOsiris, line 120: `periodNumber` this is simply a generational nonce mapped to a calendar interval? Is this consistent across accounts or a per account period. Is the period configurable etc.

Let's be much more granular in detail here on action, signal, etc.

`period_proof` also this is kind of an odd static string tbh. Think screaming snake case is more visually appealing 😄

wips/wip-1001.md
```diff
@@ -0,0 +1,346 @@
...
+- Signal contains the action (add or remove) and the list of authorized addresses
+
+Every WorldID proof natively carries an `issuerSchemaId` (mandatory public input in the circuit). The nullifier is credential-independent: the same action produces the same nullifier regardless of which credential is used. This means all credential proofs for a given WorldID in a given period map to the same nullifier.
```
0xOsiris, line 123: I honestly might leave credential agnosticness out of scope for simplicity here. We can obviously add the mapping from `issuerSchemaId` to budget in the contract for forward compatibility, but it should be implicit that this will only be rolled out for POH. 

I think this should be the case for Key Rings as well on launch.

0xOsiris: We should leave multi-credential changes to a future WIP

wips/wip-1001.md
```diff
@@ -0,0 +1,346 @@
...
+
+A user holds a WorldID with both credentials. At the start of period 7, they want to authorize two addresses (`0xAlice`, `0xBob`) and claim their full budget.
+
+**Step 1: First proof — authorize addresses and claim PoH budget.**
+
+The user generates a WorldID proof with:
+
+- Action: `"period_proof" || 7`
+- Credential: Proof-of-Human (`issuerSchemaId = 0x01`)
+- Signal: `{action: add, addresses: [0xAlice, 0xBob]}`
```
0xOsiris, line 152: Shouldn't this actually be a nullifier which derives to an admin key of a registered keyring?

Or is it implicit that `0xAlice`, and `0xBob` exist in a World ID key chain. Can you explain how these two things are correlated?

wips/wip-1004.md
```diff
@@ -0,0 +1,346 @@
...
+
+This WIP introduces native account abstraction for World Chain with three core features: (1) a WorldID gas accounting system that enables per-credential gas subsidies for verified humans using periodic nullifier-based budgets managed by a precompile, (2) native multi-authentication support allowing accounts to use different authenticator types (passkeys, signature schemes, WorldID) without bundler infrastructure, and (3) WorldID-based account recovery with timelock protection against compromised authenticators. A new transaction type is introduced that can be signed by any authorized session or root authenticator and optionally targets a specific gas budget.
+
+## Motivation
+
+### WorldID Gas Accounting
+
+World Chain needs a mechanism to subsidize gas for verified WorldID holders while preventing abuse. The following requirements drive the design:
+
+- **Per-credential gas budget.** Different credentials (e.g., Proof-of-Human, phone, NFC passport) should grant different budget amounts.
```
0xForerunner: This should say eth budget not a gas budget, as you clarify below.

wips/wip-1001.md
```diff
@@ -0,0 +1,346 @@
...
+- **Proof-based recovery.** A user can recover their account by submitting a WorldID proof, as long as they control the WorldID's authenticators.
+- **Recovery means updating authenticators.** Recovery updates the set of authenticators on the account, not a migration of funds to a new address.
+- **Protection from compromised authenticators.** A compromised WorldID authenticator must not immediately lead to a drained account. A safety mechanism (e.g., timelock) is required.
+- **Unlinkability**. A single world id should be able to enable recovery for multiple accounts. The accounts shoudl not be linkable.
+
+## Specification
+
+### WorldID Gas Accounting
+
+The accounting logic is implemented in a dedicated Gas Accounting component. Time is divided into **periods** (e.g., one month). Each WorldID gets a nullifier per period, and budget must be refreshed each period. There is a governance-configurable map from credentials (identified by `issuerSchemaId`) to claimable budget amounts, where each claim requires one WorldID proof.
```
0xForerunner, line 45: > budget must be refreshed each period.

I'm not yet convinced that this is a requirement. Why do we need to attach the gas budget to the nullifier? Obviously proofs will be required in order to update the list of verified addresses, but there should still be a way in which we don't strictly require a new proof every month.

It seems to me that we should use a session proof in order to update a list of addresses associated with a sessionID. I suppose the decision boils down to this:

Session Proofs - Better UX, fewer proofs.
Proof per Epoch - Worse UX but better guarantees with possibly revoked credentials.

How do we feel about this trade off?

wips/wip-1001.md
```diff
@@ -0,0 +1,417 @@
...
+| Created     | 2026/03/27                                                                                                                                                    |
+| Requires    | [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718), [ERC-4337](https://eips.ethereum.org/EIPS/eip-4337), [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) |
+
+## Abstract
+
+This WIP introduces native account abstraction for World Chain with three core features: (1) a **WorldID Gas Accounting** system enabling per-credential gas subsidies for verified humans via periodic nullifier-based budgets managed by a precompile, (2) **World ID KeyRings** — accounts whose session-key set is gated by a WorldID admin proof, supporting native multi-authentication (Secp256k1, P256, WebAuthn, EdDSA) without bundler infrastructure and with no sybil restriction on the number of keyrings per WorldID, and (3) WorldID-based recovery of keyring session keys with timelock protection against compromised authenticators. A new transaction type (`0x1D`) is introduced that is signed by any authorized session key of a keyring and optionally targets a specific gas budget.
```
alessandromazza98, line 14: nit: put each feature into a dotted list

alessandromazza98: further: what do you mean by WorldID admin proof? is this a normal world id uniqueness proof?

wips/wip-1001.md
```diff
@@ -0,0 +1,417 @@
...
+- **Multiple authentication methods.** Support for passkeys, different signature schemes, multi-sig, and WorldID as authenticators.
+- **Native support.** Users should be able to submit transactions directly without bundler infrastructure.
+- **Key rotation.** Authenticators can be added and removed after account creation.
+
+### WorldID Recovery
+
+- **Proof-based recovery.** A user can recover their account by submitting a WorldID proof, as long as they control the WorldID's authenticators.
+- **Recovery means updating authenticators.** Recovery updates the set of authenticators on the account, not a migration of funds to a new address.
+- **Protection from compromised authenticators.** A compromised WorldID authenticator must not immediately lead to a drained account. A safety mechanism (e.g., timelock) is required.
+- **Unlinkability**. A single world id should be able to enable recovery for multiple accounts. The accounts should not be linkable.
```
alessandromazza98, line 37: in general we're overloading too much the term account. Let's specify every time if we mean a World ID account or an Ethereum account, mainly in WIPs like this where they are frequently mentioned.

alessandromazza98: this way it becomes clear what we mean without any doubt

wips/wip-1001.md
```diff
@@ -0,0 +1,417 @@
...
+A proof with a previously-claimed credential can be resubmitted to update authorized accounts without re-claiming budget. A proof with a new credential and empty authorized accounts list can be submitted just to claim additional budget.
+
+#### Example: Authorization and Budget Claim Flow
+
+Assume governance has configured the following credential budgets:
+
+| Credential           | `issuerSchemaId` | Budget      |
+| -------------------- | ---------------- | ----------- |
+| Proof-of-Human (Orb) | `0x01`           | 50,000 Gwei |
+| Phone                | `0x02`           | 10,000 Gwei |
```
alessandromazza98, line 142: what do you mean by expressing the budget in Gwei? Gwei is a unit of price - i.e. how much ETH you pay for a unit of gas

wips/wip-1001.md
alessandromazza98: i think periodNumber is a generational nonce from the example below. It just keeps increasing every period

wips/wip-1001.md
```diff
@@ -0,0 +1,417 @@
...
+To prevent congestion from all accounts submitting refresh proofs at the start of a new period, refresh proofs for the next period can be submitted during the current period.
+
+Note that nullifiers of the same WorldID across different periods cannot be linked. If a WorldID authorizes independent addresses in each period, it remains fully anonymous.
+
+#### Gas Accounting Flow
+
+For an incoming transaction:
+
+1. The protocol looks up whether the sender address has an associated nullifier in the authorization map.
+2. If the address maps to more than one nullifier, a deterministic rule selects which nullifier's budget to use. With the new transaction type, the sender can explicitly specify which nullifier to use.
```
alessandromazza98, line 204: can you specify what you exactly mean with having more than one nullifier? If nullifier is unique per world id || rp id || action, it's not clear to me how you would obtain two or more nullifiers in the same period

wips/wip-1001.md
```diff
@@ -0,0 +1,417 @@
...
+
+### World ID KeyRing
+
+A **World ID KeyRing** is an account whose authorized signatory set is gated by a WorldID admin proof. The keyring address derives from an admin nullifier, which itself is a deterministic function of the owning WorldID identity and a user-chosen `nonce`. A single WorldID MAY manage any number of keyrings by varying `nonce`; keyring addresses produced by a single World ID are unlinkable to each other.
+
+World ID proofs act as the administrative Key over the Key Ring — We may optionally consider extending the Key Ring to support other forms of Admin Authentication. 
+
+Session keys are abstract signatories authorized to sign `0x1D` transactions on behalf of the keyring; they carry no intrinsic privileges beyond signing. Per-key permissions are out of scope for this WIP (see §Extensions).
+
+Notably KeyRings are orthogonal, not required for Gas Subsidies. They simply provided an abstraction over the signature verification of accounts which MAY be delegated for Gas Subsidies.
```
alessandromazza98, line 217: if that's the case I'd split them into a separate WIP. The gas subsidy feature is already big enough to have its own WIP without introducing further complexity by linking together another (orthogonal) feature

wips/wip-1001.md
```diff
@@ -0,0 +1,417 @@
...
+`signature_payload` per `signature_type`:
+
+| `signature_type` | Encoding                                            |
+| ---------------- | --------------------------------------------------- |
+| `0x00` Secp256k1 | `rlp([y_parity, r, s])`                             |
+| `0x01` P256      | `rlp([r, s])`                                       |
+| `0x02` WebAuthn  | `rlp([authenticator_data, client_data_json, r, s])` |
+| `0x03` EdDSA     | `rlp([R, s])`                                       |
+
+The envelope is extensible to 2D nonces, batched transactions, and native paymaster support in follow-on WIPs.
```
0xOsiris, line 356: This is just not true. I don't see how we could ever retain 2D nonce compatibility without a upgrade to the underlying account state

wips/wip-1001.md
0xOsiris: Sounds good. I will separate this. We will consider these 2 separate workstreams

0xOsiris:
Closing in favor of the final draft versions #511

PR #499, opened 2026-04-14
feat(rpc): add worldchain_simulateUserOp endpoint

karankurbur:
## Summary

Adds `worldchain_simulateUnsignedUserOp` — a custom RPC endpoint that simulates the execution phase of an unsigned ERC-4337 v0.7 UserOperation against on-chain state. The response format mirrors Blockaid's transaction simulation API so that `app-backend-main` can add World Chain's node as another security/simulation provider with no response-shape translation layer.

## Enabling the endpoint

The endpoint is **disabled by default** and is only registered when the operator passes:

```
--worldchain.simulate-enabled
```

When enabled, it is exposed on the public HTTP/WS/IPC RPC servers alongside `eth_*` and friends. There is no application-level authentication on the method itself.

### Why no JWT / API key

Nodes that run with `--worldchain.simulate-enabled` are deployed behind infrastructure-level authentication — e.g. an internal-only ingress, VPC-private listener, or auth proxy in front of the RPC port. Access control is enforced at that layer, so the node does not duplicate it. **Do not enable this flag on a publicly reachable RPC.**

```sh
curl -X POST http://<internal-rpc>:8545 \
  -H "Content-Type: application/json" \
  -d '{ "jsonrpc": "2.0", "id": 1, "method": "worldchain_simulateUnsignedUserOp", "params": { ... } }'
```

## How simulation works

The simulation only executes the **execution phase** of the UserOp — it calls the smart account with the provided `callData` as if `msg.sender` were the EntryPoint. It does NOT run the full UserOp lifecycle (validation, paymaster verification, gas accounting), so no paymaster, nonce, or gas pricing fields are needed.

1. Resolves the sealed header and state at the latest block
2. Constructs an EVM call: `caller = entryPoint`, `to = sender` (the smart account), `data = callData`
3. Executes via an Optimism-aware EVM with a custom `SimulationInspector` that captures:
   - **Full call stack trace** — every CALL/STATICCALL/DELEGATECALL at every depth with caller, target, selector, and value
   - Native ETH transfers at any depth (calls with value > 0)
4. Parses emitted logs for ERC-20/721/1155 `Transfer`, `TransferSingle`, `TransferBatch`, `Approval`, and `ApprovalForAll` events
5. Resolves on-chain token metadata (`name()`, `symbol()`, `decimals()`) for each token address, with a cross-request cache

## Call stack trace

Unlike Blockaid (which returns decoded event logs), Alchemy, or Pimlico (which don't include trace data), our `trace` array is a **full call-level execution trace** — every `CALL`/`STATICCALL`/`DELEGATECALL` at every depth, showing the actual contract-to-contract call graph. This is similar to what `debug_traceCall` with a call tracer produces, but normally only available on archive nodes with debug APIs enabled.

Each trace entry includes:
- `from` — the address making the call
- `to` — the address being called
- `method` — best-effort decoded function name (null if selector is unknown)
- `selector` — the 4-byte function selector
- `value` — ETH value sent with the call

Trace entries are emitted in execution order, so nested calls are interleaved with their parents — the call graph can be reconstructed by following the `from`/`to` chain.

Example for a token swap through a smart account:
```json
"trace": [
  { "from": "0xEntryPoint",   "to": "0xSmartAccount", "method": "execute",  "selector": "0xb61d27f6", "value": "0x0" },
  { "from": "0xSmartAccount", "to": "0xRouter",       "method": "swap",     "selector": "0x...",      "value": "0x0" },
  { "from": "0xRouter",       "to": "0xPool",         "method": "swap",     "selector": "0x...",      "value": "0x0" },
  { "from": "0xPool",         "to": "0xUSDC",         "method": "transfer", "selector": "0xa9059cbb", "value": "0x0" }
]
```

## Selector-based trace for mini app transaction filtering

The `trace` array includes decoded method names for well-known selectors. This allows `app-backend-main` to inspect what a mini app's UserOp is actually doing **at any call depth** and reject malicious transactions. The decoded selectors include:

- **Token operations**: `transfer`, `transferFrom`, `approve`, `safeTransferFrom`, `safeBatchTransferFrom`, `setApprovalForAll`, `permit`
- **Account execution**: `execute`, `executeBatch`, `multiSend`
- **Safe admin methods** (forbidden for mini apps): `addOwnerWithThreshold`, `removeOwner`, `swapOwner`, `changeThreshold`, `enableModule`, `disableModule`, `setGuard`, `setFallbackHandler`, `setModuleGuard`, `setup`

The backend can check `trace[].method` against a blocklist at any depth to reject UserOps that attempt privileged Safe operations (e.g. a mini app trying to `removeOwner` or `enableModule` on a user's wallet, even if nested inside a multicall).

## Request

```jsonc
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "worldchain_simulateUnsignedUserOp",
  "params": {
    "sender":     "0x...",       // the smart account address
    "callData":   "0x...",       // the encoded executeUserOp calldata
    "entryPoint": "0x0000000071727De22E5E9d8BAf0edAc6f37da032",
    "callGasLimit": "0x7A1200"   // optional, defaults to 8M (also the max)
  }
}
```

## Response

```jsonc
{
  "status": "success" | "revert",
  "revertReason": null | "string", // deepest reverted frame on nested reverts (see below)
  "blockNumber": 12345,
  "gasUsed": "0x...",
  "assetChanges": [
    {
      "type": "NATIVE" | "ERC20" | "ERC721" | "ERC1155",
      "from": "0x...",
      "to": "0x...",
      "rawAmount": "1000000",
      "tokenId": null | "1",     // ERC-721/1155 only
      "asset": {
        "address": "0x...",      // 0x000…0 for native ETH
        "symbol": "USDC",
        "name": "USD Coin",
        "decimals": 6,
        "type": "ERC20"
      }
    }
  ],
  "exposureChanges": [
    {
      "owner": "0x...",
      "spender": "0x...",
      "rawAmount": "...",
      "isApprovedForAll": false,
      "asset": { /* same AssetInfo shape */ }
    }
  ],
  "trace": [
    {
      "from": "0xEntryPoint...",
      "to": "0xSmartAccount...",
      "method": "execute",
      "selector": "0xb61d27f6",
      "value": "0x0"
    },
    {
      "from": "0xSmartAccount...",
      "to": "0xUSDC...",
      "method": "transfer",
      "selector": "0xa9059cbb",
      "value": "0x0"
    }
  ],
  "warnings": []
}
```

## `revertReason` semantics

`revertReason` is the **deepest** reverted frame's decoded payload, not the outermost. For ERC-4337 the EntryPoint always wraps inner reverts in `FailedOp(uint256 opIndex, string reason)`, so the outermost payload is never the contract-specific error — surfacing the deepest frame instead means consumers see the actual root cause.

Encoding rules:

- `Error(string)` → the string itself (`"ERC20: transfer amount exceeds balance"`)
- `Panic(uint256)` → `"Panic(<reason>)"` with a known label or `"Panic(unknown panic)"` for unrecognized codes
- **Custom errors** (`error InsufficientBalance(...)`) → hex-encoded raw bytes (selector + ABI-encoded args). The selector is `keccak256(sig)[..4]` and irreversible without the ABI, so the consumer fetches the originating contract's ABI from Etherscan and ABI-decodes from there
- Top-level halts (OOG, invalid opcode, etc.) → the `HaltReason` debug name (e.g. `"OutOfGas(BasicOutOfGas)"`)
- Success → `null`

For custom errors specifically the consumer flow is: identify the originating contract via the simulation context → fetch its ABI → ABI-decode `revertReason` against that ABI to get the error name and arguments.

## Testing

Fork-based integration tests in `crates/rpc/tests/fork_simulate.rs` run against live World Chain mainnet state and exercise the real WLD/USDC/Safe contracts. They run on every CI invocation via the `WORLDCHAIN_PROVIDER` secret (already wired into `rust-ci.yml`); locally:

```sh
WORLDCHAIN_PROVIDER=https://worldchain-mainnet.g.alchemy.com/v2/<KEY> \
  cargo test -p world-chain-rpc --test fork_simulate
```

Coverage:
- Log parsing for ERC-20/721/1155 `Transfer`, `TransferSingle`, `TransferBatch`, `Approval`, `ApprovalForAll`
- `SimulationInspector` capturing native ETH transfers and the full call trace
- Forbidden-Safe-selector detection at any depth
- Top-level revert with decoded reason
- Deepest-frame `revertReason` capture on REVERT and the halt-name fallback for top-level halts
- Token metadata view-call resolution against forked state

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Adds a new, compute-heavy RPC endpoint that executes EVM calls against live state and is exposed on public RPC interfaces when enabled, so correctness and operational safeguards (timeouts/concurrency) matter.
> 
> **Overview**
> Adds an opt-in `worldchain_simulateUnsignedUserOp` JSON-RPC endpoint that simulates the execution phase of an unsigned ERC-4337 UserOp against the latest block, returning gas used, a call trace, decoded revert reason, and parsed ERC-20/721/1155 transfer/approval effects (plus native ETH transfers).
> 
> Wires the endpoint into the node only when `--worldchain.simulate-enabled` is set, adds LRU-cached on-chain token metadata resolution, and enforces operational bounds via an 8M gas cap, shared blocking-pool execution with a concurrency guard, and a 5s wall-clock timeout.
> 
> Extends CI to provide `WORLDCHAIN_PROVIDER` for fork-based integration tests, updates CLI reference docs, and adds the required dependencies/dev-dependencies for simulation and fork testing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b94a2eafb9af6f7ddd7371cbd2ed341cb4c5782. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bin/world-chain/src/main.rs
```diff
@@ -46,10 +51,22 @@ fn main() {
...
+                    let eth_api_ext =
+                        WorldChainEthApiExt::new(pool, provider.clone(), sequencer_client);
                     ctx.modules.replace_configured(eth_api_ext.into_rpc())?;
                     ctx.modules
                         .replace_configured(FlashblocksOpApi.into_rpc())?;
+
+                    // Register worldchain_simulateUnsignedUserOp endpoint
+                    let chain_spec = ctx.provider().chain_spec();
+                    let evm_config = OpEvmConfig::new(chain_spec, OpRethReceiptBuilder::default());
+                    let simulate_api = WorldChainSimulate::new(
```
0xOsiris: Is this not a standard ERC-4337 API any bundler would implement?

Along the same lines as `eth_sendUserOperation`

crates/cli/src/cli.rs
```diff
@@ -51,6 +51,17 @@ pub struct WorldChainArgs {
...
+    /// Comma-separated list of IP addresses allowed to call
+    /// `worldchain_simulateUnsignedUserOp`. When set, requests from
+    /// non-whitelisted IPs are rejected. When unset, the endpoint is open.
+    #[arg(
+        long = "simulate.allowed-ips",
+        env = "SIMULATE_ALLOWED_IPS",
+        value_delimiter = ',',
+        value_name = "IP_ADDR"
+    )]
+    pub simulate_allowed_ips: Option<Vec<std::net::IpAddr>>,
```
0xOsiris: We should stick to the same authentication mechanism currently used I think. 

e.g. JWT over RPC

bin/world-chain/src/main.rs
karankurbur: https://www.alchemy.com/docs/wallets/api-reference/bundler-api/useroperation-simulation-endpoints/alchemy-simulate-user-operation-asset-changes

It only contains asset change which is not enough for all the backend checks that we need. We need approval, top level call trace (to prevent malicious miniapp transactions from changing safe settings etc)

crates/cli/src/cli.rs
karankurbur: is there any rate limiting by default?

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1045 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct SimulateUnsignedUserOpRequest {
+    /// The smart account address to call.
+    pub sender: Address,
+    /// The encoded execution payload (e.g. `executeUserOp` calldata).
+    pub call_data: Bytes,
+    /// The ERC-4337 EntryPoint address used as `msg.sender` in the simulation.
+    pub entry_point: Address,
+    /// Block to simulate against. Defaults to `latest`.
+    #[serde(default = "default_block")]
```
0xOsiris:
```suggestion
    #[serde(default = "BlockNumberOrTag::default")]
```

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1045 @@
...
+    pub block: BlockNumberOrTag,
+    /// Optional gas limit for the simulated call. Defaults to 30 000 000.
+    #[serde(default)]
+    pub call_gas_limit: Option<U256>,
+}
+
+fn default_block() -> BlockNumberOrTag {
+    BlockNumberOrTag::Latest
+}
```
0xOsiris:
```suggestion
```

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1017 @@
...
+
+#[cfg_attr(not(test), rpc(server, namespace = "worldchain"))]
+#[cfg_attr(test, rpc(server, client, namespace = "worldchain"))]
+#[async_trait]
+pub trait WorldChainSimulateApi {
+    /// Simulates an unsigned ERC-4337 v0.7 PackedUserOperation against the
+    /// specified block state. Returns asset transfers, approval changes,
+    /// decoded trace, and warnings.
+    #[method(name = "simulateUnsignedUserOp")]
+    async fn simulate_unsigned_user_op(
```
0xOsiris, line 333: Is this a standard, or are we just making this up?

0xOsiris: Also curious why we need custom call tracers. Tracers should be specified by the caller not imposed by the API imo.

0xOsiris: In any case I don't think we will ever need to hand roll our own tracer. Tracers are standard across the Eth, and Debug API. We should re-use types, not create our own.

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1023 @@
...
+            })?;
+        let block_number = header.number();
+
+        // 2. Build the EVM environment from the header
+        let evm_env = self
+            .evm_config
+            .evm_env(header.header())
+            .map_err(internal_err)?;
+
+        // 3. Get state at the target block
```
0xOsiris: Again, don't think this is necessary. We should simply gate the API with standard JWT if there is a whitelist

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1023 @@
...
+        let (inspector, inspector_handle) = new_simulation_inspector();
+        let mut evm =
+            OpEvmFactory::default().create_evm_with_inspector(&mut state, evm_env, inspector);
+
+        // 6. Build the transaction environment
+        //    from = entry_point → msg.sender inside the smart account is the EntryPoint
+        //    to   = sender      → call the smart account
+        //    data = callData    → the UserOp's execution payload
+        let gas_limit: u64 = request
+            .call_gas_limit
```
0xOsiris: We need to understand here how to prevent contention between important processes where stale page reads are held open for significant amounts of time. We should follow reth convention where possible in methods like `eth_call`, `debug_traceBlock`, etc

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1023 @@
...
+        // 8. Build the response
+        let (status, revert_reason, gas_used, logs) = match &result_and_state.result {
+            ExecutionResult::Success { gas_used, logs, .. } => {
+                (SimulationStatus::Success, None, *gas_used, logs.clone())
+            }
+            ExecutionResult::Revert { gas_used, output } => {
+                let reason = decode_revert_reason(output);
+                (SimulationStatus::Revert, Some(reason), *gas_used, vec![])
+            }
+            ExecutionResult::Halt { gas_used, reason } => {
```
0xOsiris: This isn't necessary since you are always passing `BundleState::default()`

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1023 @@
...
+                let reason_str = format!("{reason:?}");
```
0xOsiris: Not sure why this would be necessary for simulation purposes

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1023 @@
...
+    alloy_primitives::b256!("17307eab39ab6107e8899845ad3d59bd9653f200f220920489ca2b5937696c31");
+
+// ═══════════════════════════════════════════════════════════════════════════════
+// Log parsing — asset changes
+// ═══════════════════════════════════════════════════════════════════════════════
+
+pub fn parse_asset_changes(logs: &[alloy_primitives::Log]) -> Vec<AssetChange> {
+    let mut changes = Vec::new();
+
+    for log in logs {
```
0xOsiris, line 681: Would like to understand the use cases behind all of these fields. Why is `warnings` always the empty vec? Why do we need all these traces?

0xOsiris (changes requested):

crates/rpc/src/simulate.rs
karankurbur: does it hit halt for OOG? we arent specifying gas limit so i guess this would never get hit.

crates/rpc/src/simulate.rs
karankurbur: Its matching the existing blockaide response as much as possible. Warnings is unused here so will remove and just set to default value in app-backend.

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1023 @@
...
+        [0xe3, 0x18, 0xb5, 0x2b] => Some("swapOwner"),
+        [0x69, 0x4e, 0x80, 0xc3] => Some("changeThreshold"),
+        [0x61, 0x0b, 0x59, 0x25] => Some("enableModule"),
+        [0xe0, 0x09, 0xcf, 0xde] => Some("disableModule"),
+        [0xe1, 0x9a, 0x9d, 0xd9] => Some("setGuard"),
+        [0xf0, 0x8a, 0x03, 0x23] => Some("setFallbackHandler"),
+        [0xe3, 0x19, 0xf3, 0x23] => Some("setModuleGuard"),
+        [0xb6, 0x31, 0x28, 0x05] => Some("setup"),
+        // Permit2
+        [0x87, 0x51, 0x7c, 0x45] => Some("permit"),
```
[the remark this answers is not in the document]
karankurbur: these values are coming from app-backend

crates/rpc/src/simulate.rs
karankurbur: we want to be opinionated here, the tracer should not be specified by the caller  -> we specifically only care about `to` and `data` for each trace for our use case.

karankurbur: It simply used to detect unwanted admin calls to Safe contract like (addOwner) and the creation of new contracts (not allowed)

karankurbur: Here are future extensions that we might include. but regardless, its very specific to our case

Screenshot 2026-04-15 at 1 57 27 PM

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,974 @@
...
+            &self.evm_config,
+            header.header(),
+            block_id,
+            &self.metadata_cache,
+            &mut asset_changes,
+            &mut exposure_changes,
+        );
+
+        Ok(SimulateUnsignedUserOpResult {
+            status,
```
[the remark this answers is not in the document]
karankurbur: throwing an error if over 1000 ERC1155 logs

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,974 @@
...
+                    asset_type: "NATIVE".to_string(),
+                },
+            })
+            .collect()
+    }
+}
+
+pub fn new_simulation_inspector() -> (SimulationInspector, InspectorHandle) {
+    let state = Arc::new(Mutex::new(InspectorState::default()));
+    (
```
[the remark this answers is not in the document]
karankurbur: depth is not used by the consumer anyways, removing from response

crates/rpc/src/simulate.rs
karankurbur: wiring into `BlockingTaskPool` the same way that eth_call and debug_trace works. also added a hard 5s timeout on simulation so it drops the open page

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1021 @@
...
+        // succeed regardless of gas pricing, sender balance, or block limits.
+        evm_env.cfg_env.disable_block_gas_limit = true;
+        evm_env.cfg_env.disable_eip3607 = true;
+        evm_env.cfg_env.disable_base_fee = true;
+
+        // 3. Get state at the target block (same as eth_call)
+        let state_provider = self
+            .client
+            .state_by_block_id(block_id)
+            .map_err(internal_err)?;
```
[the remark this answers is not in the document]
karankurbur: Defaulting to use the node's latest block and return the block number in the response. The consumer can then decide if the response is outdated or compare the response from multiple different nodes and use the one with the latest block. This works nicely in case some nodes fall behind, the consumer will be able to determine which response to use

bin/world-chain/src/main.rs
```diff
@@ -43,13 +48,37 @@ fn main() {
...
+                    // HTTP/WS/IPC RPC servers, but only when the operator opts
+                    // in via `--worldchain.simulate-enabled`. The endpoint
+                    // performs no application-level auth — it is intended for
+                    // nodes deployed behind infrastructure-level auth (e.g. an
+                    // internal-only ingress).
+                    //
+                    // Wire it onto the same blocking pool and tracing guard as
+                    // eth_call / debug_trace* so a slow simulation can't open
+                    // unbounded MDBX read transactions or contend with the
+                    // general tokio runtime.
```
0xOsiris:
```suggestion
```

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1279 @@
...
+    /// The ERC-4337 EntryPoint address used as `msg.sender` in the simulation.
+    pub entry_point: Address,
+    /// Optional gas limit for the simulated call. Defaults to `MAX_SIMULATION_GAS`.
+    /// Must not exceed `MAX_SIMULATION_GAS`.
+    #[serde(default)]
+    pub call_gas_limit: Option<U256>,
+}
+
+/// Maximum gas limit accepted for a simulated call.
+pub const MAX_SIMULATION_GAS: u64 = 8_000_000;
```
0xOsiris, line 54: Shouldn't this be max transaction gas, so 20,000,000 post Pectra I believe

crates/rpc/tests/fork_simulate.rs
```diff
@@ -0,0 +1,758 @@
+//! Fork-based integration tests for `worldchain_simulateUnsignedUserOp`.
+//!
+//! These tests fork World Chain mainnet via an RPC URL and run simulations
+//! against real deployed contracts (WLD, USDC.e, etc.).
+//!
+//! Run with:
+//! ```sh
+//! WORLD_CHAIN_RPC_URL=https://worldchain-mainnet.g.alchemy.com/v2/<KEY> \
```
0xOsiris: Has this been added to CI?

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1281 @@
...
+/// Handles `Error(string)` (0x08c379a0) and `Panic(uint256)` (0x4e487b71).
+pub fn decode_revert_reason(output: &Bytes) -> String {
+    if output.len() < 4 {
+        return format!("0x{}", hex::encode(output.as_ref()));
+    }
+
+    let selector = &output[..4];
+
+    // Error(string) — 0x08c379a0
+    if selector == [0x08, 0xc3, 0x79, 0xa0] && output.len() >= 68 {
```
0xOsiris: Can you make these named constants

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1281 @@
...
+            if str_start + len <= output.len()
+                && let Ok(s) = std::str::from_utf8(&output[str_start..str_start + len])
+            {
+                return s.to_string();
+            }
+        }
+    }
+
+    // Panic(uint256) — 0x4e487b71
+    if selector == [0x4e, 0x48, 0x7b, 0x71] && output.len() >= 36 {
```
0xOsiris: Can you make this a named constant

crates/rpc/src/simulate.rs
```diff
@@ -0,0 +1,1281 @@
...
+        let code = U256::from_be_slice(&output[4..36]);
+        let reason = match code.try_into().unwrap_or(0xFFu64) {
```
0xOsiris: This makes no sense to me honestly

crates/rpc/src/simulate.rs
karankurbur: we only allow miniapps tx up to 8m gas

crates/rpc/tests/fork_simulate.rs
karankurbur: fixed, its called WORLDCHAIN_PROVIDER in CI secrets

0xOsiris (approved):

crates/rpc/src/simulate.rs
karankurbur: Panic returns a uint. uint can be mapped to a specific revert reason

PR #500, opened 2026-04-15
chore: cli snake case

0xForerunner:


Kemperino (approved):

crates/cli/src/cli/builder.rs
```diff
@@ -230,8 +237,8 @@ mod tests {
         CommandParser::try_parse_from([
             "bin",
             "--flashblocks.enabled",
-            "--flashblocks.override_authorizer",
```
Kemperino, line 233: shoulnd't this still work, since it's aliased?

0xForerunner: Yea it does. I tested with both. Just figured I'd change to the "correct" casing.

PR #509, opened 2026-04-21
Add trait abstraction over validator.

piohei:
> [!NOTE]
> **Medium Risk**
> Touches core flashblock execution/validation and state-root computation paths; although largely a refactor, strategy selection and first-flashblock handling could alter block construction behavior if miswired.
> 
> **Overview**
> Refactors flashblock validation to route execution through a new `ExecutionStrategy` + `StateRootStrategy` abstraction, with concrete BAL (parallel + async state-root) and legacy (sequential + sync state-root) implementations selected at runtime based on whether `access_list_data` is present.
> 
> Moves the previously inlined legacy and BAL validation/execution logic out of `FlashblocksBlockValidator` into `execution_strategy.rs`, introduces `flashblock_types.rs` to bind strategy pairs, and adds `CommittedState.is_first` so the legacy path only applies pre-execution changes on the first flashblock.
> 
> Updates transaction decoding to recover signers via `recover_signer()` and adjusts test/fuzz helpers to use the new strategy API and initialize `CommittedState` with `is_first`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53e1e6673bc34e69e126d0dec531539a9e5fa125. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/builder/src/execution_strategy.rs
+++ b/crates/builder/src/execution_strategy.rs
@@ -68,9 +68,7 @@ pub struct ValidationCtx<'a, Evm: ConfigureEvm> {
     pub attempt_metrics: &'a mut FlashblockValidationAttemptMetrics,
     pub chain_spec: Arc<OpChainSpec>,
     pub evm_env: EvmEnvFor<Evm>,
-    pub evm_config: Evm,
     pub execution_context: OpBlockExecutionCtx,
-    pub header: Arc<SealedHeader>,
 }
 
 /// Strategy for executing flashblock diff transactions.
--- a/crates/builder/src/execution_strategy.rs
+++ b/crates/builder/src/execution_strategy.rs
@@ -319,7 +319,7 @@ impl ExecutionStrategy<OpEvmConfig> for FlashblocksLegacyExecutionStrategy {
         );
 
         // Apply pre-execution changes on the first flashblock (no prior committed state).
-        if committed_state.transactions.is_empty() {
+        if committed_state.is_first {
             let pre_execution_changes_started = Instant::now();
             builder.apply_pre_execution_changes()?;
             ctx.attempt_metrics.record_stage_duration(
--- a/crates/builder/src/execution_strategy.rs
+++ b/crates/builder/src/execution_strategy.rs
@@ -1,7 +1,9 @@
 use std::{sync::Arc, time::Instant};
 
 use alloy_consensus::{BlockHeader, Header, Transaction};
-use alloy_op_evm::{OpBlockExecutionCtx, OpBlockExecutor, OpBlockExecutorFactory, OpEvmFactory, OpTx};
+use alloy_op_evm::{
+    OpBlockExecutionCtx, OpBlockExecutor, OpBlockExecutorFactory, OpEvmFactory, OpTx,
+};
 use alloy_primitives::{Address, B256, U256};
 use alloy_rpc_types_engine::PayloadId;
 use reth_evm::{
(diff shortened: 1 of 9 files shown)


crates/builder/src/execution_strategy.rs
```diff
@@ -0,0 +1,490 @@
...
+// ---------------------------------------------------------------------------
+
+pub struct AsyncStateRootStrategy;
+
+pub struct ChannelStateRootHandle {
+    pub receiver: crossbeam_channel::Receiver<Result<StateRootResult, BlockExecutionError>>,
+}
+
+impl StateRootStrategy for AsyncStateRootStrategy {
+    type EpochState = ();
```
0xOsiris: Let's just remove from the trait definition if it's unused.

crates/builder/src/execution_strategy.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    fn prepare(
+        &self,
+        client: impl StateProviderFactory + Clone + 'static,
+        parent_hash: B256,
+        bundle_state: BundleState,
+    ) -> Self::Handle;
+}
+
+pub trait StateRootHandle: Send {
```
0xOsiris: Can we separate out the trait definitions into their own file for readability?

crates/builder/src/execution_strategy.rs
```diff
@@ -0,0 +1,490 @@
...
+impl StateRootStrategy for AsyncStateRootStrategy {
+    type EpochState = ();
+    type Handle = ChannelStateRootHandle;
+
+    fn prepare(
+        &self,
+        client: impl StateProviderFactory + Clone + 'static,
+        parent_hash: B256,
+        bundle_state: BundleState,
+    ) -> Self::Handle {
```
0xOsiris:
```suggestion
    ) -> Result<Self::Handle> {
```

crates/builder/src/execution_strategy.rs
```diff
@@ -0,0 +1,490 @@
...
+pub trait StateRootStrategy: Send + Sync {
+    type EpochState: Default + Send;
+    type Handle: StateRootHandle;
+
+    fn prepare(
+        &self,
+        client: impl StateProviderFactory + Clone + 'static,
+        parent_hash: B256,
+        bundle_state: BundleState,
+    ) -> Self::Handle;
```
0xOsiris: Let's return Result here, and not panic in the hot path

crates/builder/src/execution_strategy.rs
```diff
@@ -0,0 +1,490 @@
...
+            PayloadBuildStage::SequencerTxExecution,
+            txs_execution_started.elapsed(),
+        );
+
+        let finish_state_provider = client
+            .state_by_block_hash(ctx.parent.hash())
+            .map_err(BalExecutorError::other)?;
+
+        let finalize_started = Instant::now();
+        let (outcome, bundle) = builder.finish_with_bundle(
```
0xOsiris: We actually want `outcome` here to be computed over a generic handle to the `hashed_state`, `trie_updates`, and `state_root`

There's two possible cases:
- Trie Updates w/ full state root
- A handle to a blocking task `DeferredTrieUpdates`. 

In my example `finish()` should resolve either of these. In the former case it should return instantly, in the latter it should block the current thread to resolve the `JoinHandle`

0xOsiris: So, instead of the `BlockBuilderExt` always computing the `state_root` it should take a state root strategy, and return a handle based on the strategy passed to `finish_with_bundle`

Kemperino (approved):

0xOsiris (approved):

0xOsiris:
Thanks this looks great! Nice work

PR #512, opened 2026-04-21
fix: Session IDs for native AA

0xForerunner:


wips/wip-1001.md
```diff
@@ -2,20 +2,20 @@
...
 | Requires    | [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718), [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), [ERC-4337](https://eips.ethereum.org/EIPS/eip-4337)                                                                                                                  |
 
 ## Abstract
 
-A *World ID Key Ring* is a precompile-managed account whose authorized signatory set is gated by a [World ID](https://worldcoin.org/world-id) 4.0 Uniqueness Proof over [OPRF](https://datatracker.ietf.org/doc/rfc9497/). We extend [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) with a typed transaction envelope (`0x1D`) signed by any authorized *session key* of a keyring.
+A *World ID Key Ring* is a precompile-managed account whose creation is gated by a [World ID](https://worldcoin.org/world-id) 4.0 Uniqueness Proof over [OPRF](https://datatracker.ietf.org/doc/rfc9497/), and whose subsequent mutations are gated by Session Proofs bound to a per-keyring session. We extend [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) with a typed transaction envelope (`0x1D`) signed by any authorized *execution key* of a keyring.
```
0xOsiris: Can we remove the `execution key` renaming. `session keys` is standard nomenclature for the authenticators composing a key chain. I don't think we should invent new terms

0xForerunner: session keys and world id sessions are very confusing to talk about together. delegateKey?

wips/wip-1001.md
```diff
@@ -115,72 +124,94 @@ enum UpdateMode { Create, Add, Remove }
 
 struct KeyringUpdate {
     UpdateMode   mode;
-    SessionKey[] keys;
+    ExecutionKey[] keys;
```
0xOsiris: I would strongly prefer sticking to `SessionKey` naming here. As this is conventional

wips/wip-1001.md
0xOsiris: It would also make the PR much easier to read as most of the changes are just `session key` -> `execution key`

0xOsiris: I think we should just add this explicit differentiation of terms to the constants. Or add a legend at the top

0xOsiris: Session Key is the correct term imo

0xForerunner: I'm on the fence but you maybe right. This will def be a source of confusion though. I can change back.

wips/wip-1001.md
0xForerunner: @0xOsiris reverted

wips/wip-1001.md
```diff
@@ -119,49 +128,71 @@ struct KeyringUpdate {
...
 ```
 
-The precompile MUST:
+For `create`, the precompile MUST:
+
+1. Require `createUpdate.mode == Create`.
+2. Recompute `action = keccak256(abi.encodePacked(WORLD_ID_KEYRING_TAG, uint256(keyringNonce)))`.
+3. Recompute `signalHash = uint256(keccak256(abi.encode(createUpdate))) >> 8`.
+4. Invoke `WorldID.verify(keyringNullifier, action, WORLD_CHAIN_RP_ID, proofNonce, signalHash, …, proof)`.
+5. Derive `keyring = bytes20(keccak256(keyringNullifier))`, associate the supplied `sessionId` with `keyring`, and apply `createUpdate` immediately.
```
kilianglas: Shouldn't we call this `address` instead of `keyring`?

kilianglas (approved):

kilianglas:
NIT: Could we get a primer on session proofs here?

wips/wip-1001.md
kilianglas: I'd also use SessionKeys. E.g., Tempo AA uses this term as well

PR #524, opened 2026-04-22
fix: bal validation should execute txs in parallel

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4519/bal-make-sure-that-tx-execution-is-done-in-parallel-when-validating-a

### Notes

I needed to wrap the reth `StateProviderBox` into a `Arc<Mutex<..>>`. The reason is that rayon (the crate we use to make tx execution in parallel) requires what it's going inside the parallel work to be `Clone + Send` because it needs to clone the object and share it between all the worker threads. Reth `StateProviderBox` is only `Send`, therefore it would make the code to not compile. To fix it we can wrap it with an `Arc`. But by doing this we lose the `Send` trait. In fact, `Arc<T>` is `Send` if and only if `T: Sync`. Since `StateProviderBox` is not `Sync`, by wrapping it with an `Arc`, we lose `Send`. And to fix that we need to add a `Mutex`. This is why the final wrap is `Arc<Mutex<..>>`.

### CI failure

Benchmark CI is failing but it's expected. The reason is that main branch is actually not using BAL really, therefore all previous benches are not correct.

---

> [!NOTE]
> **Medium Risk**
> Enables Rayon-based parallel transaction execution during BAL validation and introduces shared mutable access to the upstream state provider via `Arc<Mutex<_>>`, which can affect correctness/perf under concurrency if locking or provider assumptions are wrong.
> 
> **Overview**
> Re-enables **parallel transaction execution** for BAL-based flashblock validation by switching the per-tx execution loop in `BalBlockValidator::execute_block` from sequential iteration back to `rayon` (`into_par_iter` + `map_with`).
> 
> Adds `SharedStateProviderDatabase`, a cloneable `revm::DatabaseRef` adapter that wraps `StateProviderBox` in `Arc<Mutex<_>>`, and updates the BAL validation path to use it so temporal DBs can safely fall back to the shared provider across parallel workers (including adding a `DbRef: Send` bound).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f8ac1129fb574c2651e64e5358250dcebe6583e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/builder/src/database/mod.rs
+++ b/crates/builder/src/database/mod.rs
@@ -1,4 +1,5 @@
 pub mod bal_builder_db;
 pub mod bundle_db;
+pub mod shared_state_provider_db;
 pub mod temporal_db;
 pub mod temporal_map;
--- a/crates/builder/src/database/shared_state_provider_db.rs
+++ b/crates/builder/src/database/shared_state_provider_db.rs
@@ -1,0 +1,62 @@
+use std::sync::Arc;
+
+use alloy_primitives::{Address, B256};
+use parking_lot::Mutex;
+use reth_provider::{
+    AccountReader, BlockHashReader, BytecodeReader, ProviderError, StateProvider, StateProviderBox,
+};
+use revm::DatabaseRef;
+
+/// A cloneable [`DatabaseRef`] adapter over [`StateProviderBox`] for BAL validation.
+///
+/// The upstream provider handle is `Send` but not `Clone` or `Sync`, so BAL workers share it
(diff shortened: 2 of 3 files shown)


alessandromazza98:
Note: benchmark CI is failing and it's ok. The reason is that main branch is actually not using BAL really

Kemperino (approved):

PR #550, opened 2026-04-24
chore: bump to reth v2.0.0 and op-reth v2.1.0

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4518/bump-world-chain-to-reth-v200

---

> [!NOTE]
> **Low Risk**
> Low risk CI-only change that adjusts Cargo env settings to reduce debug info generation; no production code paths are affected.
> 
> **Overview**
> Updates Rust CI workflow environment to set `CARGO_PROFILE_DEV_DEBUG=0` and `CARGO_PROFILE_TEST_DEBUG=0`, disabling debug info generation for `dev` and `test` builds in CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9758682b0f24988a7100f4b6291ad3014727b10f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/rpc/src/eth/transaction.rs
+++ b/crates/rpc/src/eth/transaction.rs
@@ -6,10 +6,10 @@ use alloy_primitives::{B256, Bytes, TxHash};
 use reth_node_api::BlockBody;
 use reth_optimism_primitives::OpPrimitives;
 use reth_optimism_rpc::{OpEthApi, OpEthApiError};
-use reth_primitives::{Recovered, TransactionMeta};
+use reth_primitives_traits::{Recovered, SignerRecoverable, TransactionMeta};
 use reth_provider::{ProviderReceipt, ProviderTx, ReceiptProvider, TransactionsProvider};
 use reth_rpc_eth_api::{
-    EthApiTypes, FromEthApiError, FromEvmError, RpcConvert, RpcNodeCore,
+    EthApiTypes, FromEthApiError, FromEvmError, RpcConvert, RpcNodeCore, RpcNodeCoreExt,
     helpers::{
         EthTransactions, LoadPendingBlock, LoadTransaction, SpawnBlocking, spec::SignersForRpc,
     },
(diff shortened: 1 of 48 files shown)


crates/rpc/src/eth/transaction.rs
```diff
@@ -56,22 +56,47 @@ where
...
+            TransactionMeta,
+            ProviderReceipt<Self::Provider>,
+            Option<Arc<Vec<ProviderReceipt<Self::Provider>>>>,
+        )>,
+        Self::Error,
+    >
     where
         Self: 'static,
     {
+        if let Some(cached) = self.cache().get_transaction_by_hash(hash).await
```
0xForerunner, line 74: Why is there a bunch of new logic here?

0xForerunner (approved):
Since we need to ship an upgrade next week, I wonder if we should delay this until I merge the hard fork override? I'm not sure we want to commit shipping 2.0 to mainnet next week. WDYT?

alessandromazza98:
> Since we need to ship an upgrade next week, I wonder if we should delay this until I merge the hard fork override? I'm not sure we want to commit shipping 2.0 to mainnet next week. WDYT?

I agree with you, I won't merge this until jovian is live on mainnet if this sounds good to you as well.

crates/rpc/src/eth/transaction.rs
alessandromazza98: it's from reth new logic. Essentially the tx can be already in this new cache and if it is we can immediately return it instead of fetching it

0xOsiris:
> > Since we need to ship an upgrade next week, I wonder if we should delay this until I merge the hard fork override? I'm not sure we want to commit shipping 2.0 to mainnet next week. WDYT?
> 
> I agree with you, I won't merge this until jovian is live on mainnet if this sounds good to you as well.

I agree as long as there is no Jovian related changes

0xOsiris (approved):

PR #560, opened 2026-04-28
fix: merge conflicts

0xForerunner:


alessandromazza98 (approved):
What is this new tx type for? Note that CI is not passing though (? maybe just re-run it?)

EDIT: no worry, I see this is for OP stuff and CI is failing because base doesn't compile

PR #561, opened 2026-04-28
feat: remove mutex db wrapper in bal validator

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4522/feat-remove-arcmutex-wrapper-from-db-in-bal-validation

### Problem

We are using a `Arc<Mutex<..>>` wrapper on the db in the BAL validator because the db returned by reth's `client.state_by_block_hash()` doesn't implement `Clone + Send + Sync`, but only `Send`. This causes the parallel tx execution to be still sort of serialized when they need to hit the underlying db that is under a mutex.

### Solution

Instead of sharing the same provider to all rayon workers, assign a new provider to each rayon worker. This way all db hits can be done in parallel without the need of a mutex wrapper on the db.

### Notes

This greatly improves BAL performances, even though it doesn't make them always equal to non-BAL performances. In particular BAL performances are still better for world id bn254 precompiles txs, but they are slightly lower for eth transfers or fib interactions (simple smart contract interactions)

Previous BAL performances:

```
flashblock_validation_process_flashblock_eth_transfers_with_bal_live/txs/50
                        time:   [2.0793 ms 2.0864 ms 2.0938 ms]

flashblock_validation_process_flashblock_eth_transfers_with_bal_live/txs/500
                        time:   [19.536 ms 19.566 ms 19.596 ms]

flashblock_validation_process_flashblock_eth_transfers_with_bal_live/txs/1000
                        time:   [39.012 ms 39.127 ms 39.275 ms]
```

Now:

```
flashblock_validation_process_flashblock_eth_transfers_with_bal_live/txs/50
                        time:   [1.4989 ms 1.5008 ms 1.5024 ms]

flashblock_validation_process_flashblock_eth_transfers_with_bal_live/txs/500
                        time:   [13.255 ms 13.355 ms 13.420 ms]

flashblock_validation_process_flashblock_eth_transfers_with_bal_live/txs/1000                                      
                       time:   [25.906 ms 26.023 ms 26.190 ms]
```

### Auxiliary Notes

This PR also fixes a CI failure around foundry that was looking for a `CI` inexistent profile.

---

> [!NOTE]
> **Medium Risk**
> Changes BAL parallel execution to create per-worker state providers instead of sharing a mutex-wrapped provider, affecting concurrency and state read paths during block validation. Also refactors `TemporalDbFactory`’s API and adds new profiling harnesses, so validation behavior/perf should be rechecked under load.
> 
> **Overview**
> **Removes the mutex-based fallback DB wrapper in BAL validation** by no longer sharing a single `StateProvider` across Rayon workers; each worker now initializes its own `state_by_block_hash` provider and builds its own `BundleDb` for fallback reads.
> 
> Refactors `TemporalDbFactory` to store only the cached BAL pre-state (no embedded underlying DB), requiring callers to pass a fallback `DatabaseRef` when constructing `TemporalDb`s; updates validator wiring and tests accordingly, and tightens several `Provider`/`client` bounds to require `Sync`.
> 
> Adds an ignored live-node `samply` profiling test target (`flashblock_validation_live_node_profile`) and a `foundry.toml` `[profile.ci]` stub.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd97251785a6ea80fe2479341549330125c7af4e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/builder/tests/flashblock_validation_live_node_profile.rs
```diff
@@ -0,0 +1,142 @@
+use std::sync::Arc;
```
alessandromazza98, line 1: this file is needed only to better profile with `samply`.

Kemperino (approved):
nice, that this improves performance 🙌

PR #592, opened 2026-05-01
perf: optimize allocator usage in builder crate

birs0:
Minor performance improvement to allocator handling in the builder crate. Reduces unnecessary heap allocations during benchmark runs.


birs0:
Closing - will revisit later.

Issue #532, opened 2026-04-23
WIP-1001: Keyring timelock (queue / cancel / execute)


0xOsiris:
Sub-task of the **WIP-1001** tracking issue.

## Scope
Implement the configurable delay for `Add` / `Remove` keyring updates and the cancellation surface.

Design constraints from §Timelock:
- `Create` executes immediately (no incumbents to cancel).
- `Add` and `Remove` are pending until the delay elapses.
- **Existing session keys MAY cancel a pending update during the delay window.**
- Mitigate the *revocation-then-add* attack path (a Session Proof removes all keys then adds malicious ones with no incumbents left to cancel) — e.g. delay removals, or allow recently-removed keys to cancel pending adds.

Define and implement the queue / cancel / execute ABI additions (TBD in spec — this issue should finalize them). Include storage for the pending-update queue and its TTL semantics.

## WIP reference
[`wips/wip-1001.md`](https://github.com/worldcoin/world-chain/blob/main/wips/wip-1001.md) — §Timelock and §Security Considerations → Timelock Attacks.

## Acceptance criteria
- Pending `Add`/`Remove` cannot be executed before delay elapses.
- An incumbent session key can cancel a pending update within the window.
- Revocation-then-add mitigation path has an explicit unit/property test.

## Dependencies
- Precompile `update` path issue.

PR #593, opened 2026-05-01
feat: WIP-1001 - Issue 532: Session Key Update Timelocks

Kemperino:
For https://github.com/worldcoin/world-chain/issues/532 , I think it's more user friendly if we unify Adding and Removing of Session Keys to avoid having to wait the timelock period twice in case of a rolling change.


Kemperino:
@codex @cursor review please

Kemperino:
@codex @cursor review again

0xOsiris (commented):
Sorry for the late review.

What do you think of updating the time locked update functionality to be a policy instantiated on the account at creation. e.g. The creation proof would delegate the keys, admin, and policy. The policy would determine whether update finalization would be time locked or not / the period of the timelock. The schema here also needs to be generalized for general `World Chain Accounts` which are not provisioned via a proof. 

I think we can generalize this policy in a future extension to also define the semantics around session key permissions, and scope but we should confine to just time locks for now.

Ideally I'd like to update the spec on this first to define the semantics and behavior I think, and then do the contract implementation in a separate PR?

PR #597, opened 2026-05-01
fix: just playground

0xForerunner:


crates/test-utils/src/e2e_harness/setup.rs
```diff
@@ -107,6 +107,7 @@ where
 {
     pub node: WorldChainNodeTestContext<T>,
     pub ext_context: WorldChainNodeExtContext<T>,
+    pub beacon_engine_handle: ConsensusEngineHandle<OpEngineTypes>,
```
0xOsiris: Are you able to bind `OpEngineTypes` to `T`?

0xOsiris (approved):

PR #623, opened 2026-05-07
fix(ci): repair sync workflow and target World Chain mainnet

alessandromazza98:
### Problem

The `sync test` workflow was a bit broken:

- It targeted Base mainnet instead of World Chain.
- It ran on every `pull_request` and on every `v*` tag push, making a full chain sync part of regular PR CI. Far too expensive.
- The `Run sync` step invoked the binary as `~/.cargo/bin/world-chain` while the verify and unwind steps used `world-chain` directly, so the steps disagreed about how the binary was on `PATH`.
- It was missing the hardening (scoped `permissions`, repo guard, `persist-credentials: false`) that Reth's reference workflow uses.

### Solution

- Switch the matrix to World Chain mainnet with tip `0x836c2233…5343913` (block 10000) and unwind target `0x7aed50a2…a7ef5867` (block 9800).
- Trigger only on `workflow_dispatch` and a 6-hour cron, so we still catch chain-breaking regressions quickly without running a sync per PR.
- Adopt Reth's hardening: top-level `permissions: {}`, job-level`permissions: contents: read`, `if: github.repository == 'worldcoin/world-chain'`, and `persist-credentials: false` on checkout.
- Drop the `~/.cargo/bin/` prefix so all `world-chain` invocations consistently rely on `PATH` after `just install`.

### Notes

This PR will make CI in every PR much more faster and more reliable. Most of the times the `sync test` CI job was not working at all, so it wasn't actually giving much information. With this change it can work:

- when manually triggered
- automatically every 6 hours so that we catch potential issues

---

> [!NOTE]
> **Low Risk**
> Low risk: CI-only workflow changes that adjust triggers and execution; main risk is reduced coverage (no PR/tag runs) or a misconfigured chain tip/hash causing scheduled sync failures.
> 
> **Overview**
> Updates the `sync test` GitHub Actions workflow to run only on manual dispatch and a 6-hour cron (removing PR and tag triggers) and to guard execution to `worldcoin/world-chain`.
> 
> Retargets the sync matrix from Base to **World Chain mainnet** (new `tip`/`unwind-target` hashes) and makes the job more secure/reliable by scoping `permissions`, disabling checkout credential persistence, running on `ubuntu-latest`, and invoking the built binary consistently via `PATH` (no `~/.cargo/bin` prefix).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9dbfc23b9e1ebe4951be99e6833c61f112cb589f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -20,7 +20,7 @@ jobs:
   sync:
     if: github.repository == 'worldcoin/world-chain'
     name: sync (${{ matrix.chain.bin }})
-    runs-on: arc-public-8xlarge-amd64-runner
+    runs-on: ubuntu-latest
     permissions:
       contents: read
     env:
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -5,11 +5,8 @@ name: sync test
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * 1"
-  push:
-    tags:
-      - v*
-  pull_request:
+    - cron: "0 */6 * * *"
+
 env:
   CARGO_TERM_COLOR: always
 
@@ -17,10 +14,15 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   sync:
+    if: github.repository == 'worldcoin/world-chain'
     name: sync (${{ matrix.chain.bin }})
-    runs-on: ubuntu-latest
+    runs-on: arc-public-8xlarge-amd64-runner
+    permissions:
+      contents: read
     env:
       RUST_LOG: info,sync=error
       RUST_BACKTRACE: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/sync.yml
```diff
@@ -5,22 +5,24 @@ name: sync test
...
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   sync:
+    if: github.repository == 'worldcoin/world-chain'
```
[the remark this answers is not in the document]
alessandromazza98: edited the description

.github/workflows/sync.yml
```diff
@@ -5,22 +5,24 @@ name: sync test
...
     name: sync (${{ matrix.chain.bin }})
     runs-on: ubuntu-latest
```
piohei, line 25: Shouldn't we use our own runners?

alessandromazza98: our own runners cannot access external peers, I've tried it before but it wasn't working

alessandromazza98: added a comment to explain this

piohei (approved):

PR #625, opened 2026-05-08
wip: wip-1001 rewrite with scoped EIP-1271 session and admin signers

0xOsiris:
> [!NOTE]
> **Medium Risk**
> Re-specifies core `0x1D` validation and account/keyring management semantics, so downstream implementations and clients must update for new EIP-1271-only flows, restricted validation frames, and timelocked activation rules.
> 
> **Overview**
> Rewrites `wip-1001` to make **EIP-1271 the sole authorization mechanism** for both admin updates and `0x1D` transaction signing, removing native dispatch over signature/key types (secp256k1/P256/WebAuthn/EdDSA/World ID) in favor of signer contracts (with default World ID and secp256k1 reference factories).
> 
> Replaces the precompile-centric model with a `WORLD_CHAIN_ACCOUNT_MANAGER` predeploy spec: deterministic address derivation from `(chainId, manager, adminSigner, accountSalt)`, full-keyring replacement updates via `queueKeyringUpdate` + permissionless `activateKeyringUpdate`, and an explicit **timelock** requirement (`MIN_KEYRING_UPDATE_DELAY > txpool expiration`) for both keyring and signer-state changes.
> 
> Defines **restricted, fixed-gas validation frames** for protocol calls into signers (allowlisted crypto precompiles only, forbidden opcodes/call targets), and adds post-execution **trace-based policy validation** where the session signer receives a bounded `WorldChainExecutionTrace` and can veto the transaction after tentative execution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0de0d44333496f5434445d9ba6e89c3a965c0db5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
-- **Secp256k1 admin.** Creation and keyring mutations gated by ECDSA signatures over a uniform `signalHash` from a single secp256k1 admin key fixed at creation. No recovery — admin-key loss is terminal.
-- **P256 admin.** Same shape as Secp256k1 with a P256 (`secp256r1`) admin key.
+- **WorldID admin.** Creation gated by a [World ID 4.0](https://github.com/worldcoin/world-id-protocol) Uniqueness Proof; subsequent keyring mutations gated by Session Proofs against a stored `sessionId`. Recovery from lost or compromised session keys is the degenerate case of the same upgrade path. A single WorldID `MAY` manage any number of WorldID-admin accounts indexed by a user-chosen `worldIDAccountNonce`; accounts owned by the same WorldID are unlinkable across distinct values.
```
alessandromazza98: sessionId has never been introduced before. It would be better to briefly describe what it is for better clarity here. I think just reminding that this `sessionId` is part of World ID 4.0 protocol may help.

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
+- **Secp256k1 EOA admin.** Creation and keyring mutations gated by Ethereum-style ECDSA signatures over a uniform `signalHash` from a single admin address fixed at creation. Admin-key loss is terminal.
+- **EIP-1271 contract admin.** Creation and keyring mutations gated by `isValidSignature(bytes32,bytes)` on a single contract signer fixed at creation and evaluated under the restricted validation frame. Recovery, if any, is delegated to the signer contract's own policy.
```
alessandromazza98: what's the motivation of having multiple admin types? And not just the World ID admin ?

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
-| `WebAuthn`  | `0x02` | [secp256r1](https://neuromancer.sk/std/nist/P-256) (NIST P-256)                         | 64 bytes — uncompressed `(x, y)` (P256 under [WebAuthn](https://www.w3.org/TR/webauthn-3/)) |
-| `EdDSA`     | `0x03` | [edwards25519](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1)               | 32 bytes — compressed public key (Ed25519)                                                  |
-
-`KeyType` and `adminType` occupy disjoint domains: their numeric values are defined in different enums and MUST NOT be used interchangeably in the precompile interface or the address-derivation preimage. Session keys carry no intrinsic privileges beyond signing `0x1D` transactions; per-key permissions (expiry, spend limits, call scopes) are out of scope for this WIP (see [Extensions](#extensions)).
+- **`Create`** -- installs `keys` as the full keyring at the account's address and initializes the account with the supplied admin authority. MUST revert if the account already exists, if `expectedCurrentKeyringHash != 0x00..00`, if `keys.length == 0`, or if `keys.length > MAX_SESSION_KEYS`.
```
alessandromazza98: why revert if `expectedCurrentKeyringHash != 0x00..00` ? shouldn't this revert if `expectedCurrentKeyringHash == 0x00..00` ? or if the hash is not actually the correct hash ?

wips/wip-1001.md
```diff
@@ -150,34 +234,23 @@ struct UpdateSignal {
...
-
-- `Create` executes immediately (no prior session keys exist to cancel).
-- `Add` and `Remove` are pending until the delay elapses.
-- Existing session keys MAY cancel a pending update during the delay window.
-- The revocation-then-add attack (an admin-authorized update removes all keys, then adds malicious ones with no incumbents left to cancel) MUST be mitigated — e.g., by delaying removals, or by allowing recently-removed keys to cancel pending adds.
-
-Queue / cancel / execute interface additions are TBD. Timelock semantics apply uniformly across admin instantiations.
+This is the abstract replay-protection contract; each instantiation realizes it via its own verification primitive. Instantiations MUST NOT introduce additional bespoke replay machinery -- `adminNonce` + account-existence is sufficient and uniform.
 
 #### Precompile Interface
```
alessandromazza98: this is the predeploy now right, not a precompile

alessandromazza98: here and below the term `precompile` has been used but I think `predeploy` should be used instead

alessandromazza98: or eventually just `smart contract`

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
 
 ### Constants
 
-| Name                              | Value                                        | Description                                                 |
-| --------------------------------- | -------------------------------------------- | ----------------------------------------------------------- |
-| `WORLD_TX_TYPE`                   | `0x1D`                                       | [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) tx type |
+| Name                              | Value                                        | Description                                                  |
+| --------------------------------- | -------------------------------------------- | ------------------------------------------------------------ |
+| `WORLD_TX_TYPE`                   | `0x1D`                                       | [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) tx type  |
 | `WORLD_CHAIN_ACCOUNT_PRECOMPILE`  | `0x000000000000000000000000000000000000001D` | Stateful precompile managing the World Chain Account registry |
```
alessandromazza98: is it really a precompile or just a predeploy / normal smart contract?

wips/wip-1001.md
```diff
@@ -211,24 +289,119 @@ interface IWorldChainAccount {
...
+1. Require `keyringUpdate.mode == SetKeyring`.
+2. Validate all `keyringUpdate.keys` and keyring constraints.
+3. Load the account's stored `adminType`, `verificationPublicId`, `keyringHash`, and `adminNonce`.
+4. Require `keyringUpdate.expectedCurrentKeyringHash == keyringHash`.
+5. Recompute the `Update` `signalHash` over `(accountAddress, adminNonce, msg.sender, keyringUpdate)`.
+6. Verify `adminAuthorization` per `adminType` against `signalHash` and `verificationPublicId`.
+7. Replace `sessionKeys` with `keyringUpdate.keys` and store `keyringHash = keccak256(abi.encode(keyringUpdate.keys))`.
+8. Increment `adminNonce`.
+
+> **Note -- Precompile gas accounting out of scope.** The full gas pricing schedule for interacting with `WORLD_CHAIN_ACCOUNT_PRECOMPILE` -- covering admin verification cost, per-session-key storage mutation, and the amortized cost of reads during `0x1D` validation -- is **NOT** fully defined here. A follow-on revision MUST specify gas costs per operation. EIP-1271 validation gas is separately constrained by [Restricted EIP-1271 Validation Frame](#restricted-eip-1271-validation-frame).
```
alessandromazza98: if `WORLD_CHAIN_ACCOUNT_PRECOMPILE` is a precompile ok, but from what I understand this is actually just a predeploy / smart contract. Therefore the gas logic should be the normal one for solidity smart contracts.

kilianglas:
I think we definitely want to support native P256 session keys for efficiency reasons.

Reasoning: Passkeys are the most userfriendly session key format and will thus be likely used in high frequency in the future.

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
+    address key;
+    uint32  validationGasLimit;
+    SessionPolicy policy;
+}
+
+struct SessionPolicy {
+    uint64      validAfter;
+    uint64      validUntil;
+    uint256     maxValuePerTx;
+    CallScope[] callScopes;
```
kilianglas: I think these should go into a separate WIP, since this is non-trivial to implement and requires EVM hooks. Also, maxValuePerTx is not sufficient, we need per ERC-20 token limits, otherwise this feature is useless. I guess a separate WIP would make more sense. We could either continue https://github.com/worldcoin/world-chain/pull/588 for this, or start something new.

wips/wip-1001.md
kilianglas: The idea is, that we allow for complex admin logic, e.g., multisigs, or alternative signature types. This makes the admin role extensible.

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
+```solidity
+enum KeyType {
+    Secp256k1EOA,  // 0x00
+    EIP1271        // 0x01
+}
+
+struct SessionKey {
+    KeyType keyType;
+    address key;
+    uint32  validationGasLimit;
```
kilianglas: I don't think the validation gas limit should be configurable. Couldn't I simply DoS the chain by setting this to a very high value, and implenting a 1271 signer with very high gas usage?

wips/wip-1001.md
```diff
@@ -211,24 +289,119 @@ interface IWorldChainAccount {
...
+CALL
+CALLCODE
+DELEGATECALL
+CREATE
+CREATE2
+SELFDESTRUCT
+```
 
-> **Note — Precompile gas accounting out of scope.** The specific gas pricing schedule for interacting with `WORLD_CHAIN_ACCOUNT_PRECOMPILE` — covering admin verification cost (Groth16 / ECDSA / P256), per-session-key storage mutation, timelock queue/cancel/execute operations, and the amortized cost of `isAuthorized` reads during `0x1D` validation — is **NOT** defined here. A follow-on revision MUST specify gas costs per operation per admin type.
+This list removes block-context, transaction-context, external-state, transient-state, logging, mutation, and arbitrary call-depth dependencies from protocol signature validation. EIP-1271 validity MAY depend on the signer contract's own persistent storage via `SLOAD`; this is the intentional stateful policy surface of contract signatures.
```
kilianglas: I think we should think about splitting this into a separate WIP. This should likely be audited independently as well.

wips/wip-1001.md
```diff
@@ -242,25 +415,25 @@ Each instantiation specifies:
...
+| Name                          | Value                                        | Description                                           |
+| ----------------------------- | -------------------------------------------- | ----------------------------------------------------- |
+| `WORLD_CHAIN_RP_ID`           | `480`                                        | World Chain's registered RP ID                        |
+| `WORLD_ID_ACCOUNT_TAG`        | `"WORLD_ID_ACCOUNT"` (UTF-8, 16 bytes)       | Domain tag for WorldID-admin account creation actions |
 
-`verificationPublicId` is `sessionId` (uint256), provisioned by the caller at creation and supplied by the WorldID authenticator via OPRF.
+`verificationPublicId` is `abi.encode(uint256 sessionId)`. `sessionId` is provisioned by the caller at creation and supplied by the WorldID authenticator via OPRF.
 
-Address-defining material is `ν_k`, the creation Uniqueness Proof's nullifier:
+Address-defining material is the creation Uniqueness Proof's nullifier:
```
kilianglas: We don't need a nullifier proof for the WorldID admin type. The address can be derived from the session key.

wips/wip-1001.md
```diff
@@ -315,159 +490,159 @@ WorldIDAdminAuth {
...
-
-- **(G5) Signal binding.** The `signalHash` is "dummy-squared" in the circuit (constrained to $h \cdot h = h^2$), preventing a proof from being reused with a different signal.
-
-For account creation, **(G4)** binds $\nu_k$ to the triple (owner leaf, `WORLD_CHAIN_RP_ID`, `action(worldIDAccountNonce)`) — making `worldIDAccountNonce` the sole account selector per identity — and **(G5)** binds the Uniqueness Proof to the exact `Create` payload it authorizes. For later updates, `verifySession` binds the proof to the account's stored `sessionId`; the accompanying `sessionNullifier` is a per-proof verifier input rather than persistent account state.
+For account creation, deterministic nullifier derivation binds the creation nullifier to the tuple (owner leaf, `WORLD_CHAIN_RP_ID`, `action(worldIDAccountNonce)`) -- making `worldIDAccountNonce` the sole account selector per identity -- and signal binding binds the Uniqueness Proof to the exact `Create` payload and stored `sessionId` it authorizes. For later updates, `verifySession` binds the proof to the account's stored `sessionId`; the accompanying `sessionNullifier` is a per-proof verifier input rather than persistent account state.
```
kilianglas: This section needs to be replaced by a section about WorldID session proofs

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
+A *World Chain Account* is a smart contract managed account whose 20-byte address is deterministic and lifetime-stable. Each account has a single *admin authority* fixed at creation, which gates mutation of the account's keyring -- the set of session keys authorized to sign on its behalf. We extend [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) with a typed transaction envelope (`0x1D`) signed by any authorized session key in a World Chain Account's keyring.
+
+This WIP deliberately minimizes the diff from Ethereum's existing signature model. The protocol recognizes only two session-key verification modes:
+
+- **Secp256k1 EOA key.** The `0x1D` transaction carries an Ethereum-style secp256k1 ECDSA signature. Validation recovers an address from the transaction signing hash and requires that address to be authorized in the account keyring.
```
kilianglas: Why do we only support Secp256k1 natively? I feel like we should support P256 natively as well.

wips/wip-1001.md
```diff
@@ -1,142 +1,226 @@
...
+| `keyType`          | Value  | `key` meaning                                     | `validationGasLimit`                         |
+| ------------------ | ------ | ------------------------------------------------- | -------------------------------------------- |
+| `Secp256k1EOA`     | `0x00` | Ethereum address recovered from secp256k1 ECDSA   | MUST be `0`                                  |
+| `EIP1271`          | `0x01` | Contract implementing `isValidSignature(bytes32,bytes)` | MUST be `> 0` and `<= MAX_EIP1271_VALIDATION_GAS` |
+
+`key` MUST NOT be the zero address. For `EIP1271`, `key` MUST have non-empty code at the time it is added to a keyring or fixed as an admin authority.
+
+Keyring uniqueness is defined by `(keyType, key)`. A keyring MUST NOT contain two entries with the same `(keyType, key)`, even if their `validationGasLimit` or policy values differ. Changing an EIP-1271 session key's gas limit or a session key's policy requires replacing the full keyring with `SetKeyring`.
```
kilianglas: Again, why no native P256?

wips/wip-1001.md
kilianglas: Having a full WorldID admin implies that we ship with WorldID recovery, which we can not. Either we block update operations for the WorldID admin somehow, or we implement this as a 1271 module in the future.

wips/wip-1001.md
```diff
@@ -315,159 +490,159 @@ WorldIDAdminAuth {
...
+WIP-1001 does not define a separate recovery path for EIP-1271-admin accounts. Recovery, signer rotation, quorum changes, passkey replacement, or social recovery MAY be implemented inside the signer contract's own state and policy. If the signer contract can no longer return the EIP-1271 magic value under the restricted validation frame, the account's keyring can no longer be mutated.
 
 ### Recovery
 
 Recovery semantics are per-instantiation:
 
-- **WorldID admin.** Recovery is the degenerate case of an `Add` or `Remove` submitted via Session Proof — see the WorldID admin instantiation. Lost or compromised session keys are recovered by any fresh Session Proof bound to the account's stored `sessionId`.
-- **Secp256k1 admin.** **NOT supported.** Loss of the admin key is terminal.
-- **P256 admin.** **NOT supported.** Loss of the admin key is terminal.
+- **WorldID admin.** Recovery is the degenerate case of `SetKeyring` submitted via Session Proof -- see the WorldID admin instantiation. Lost or compromised session keys are recovered by any fresh Session Proof bound to the account's stored `sessionId`.
```
kilianglas: Again, we can not ship with this

kilianglas (changes requested):

kilianglas:
Another thing we lilkely want to think about, is wether we want to restrict the set of 1271 signer implementations, at least initially. I.e., we would implement a set of supported modules and wouldn't allow anyone else to add new ones.

wips/wip-1001.md
alessandromazza98: then why EOA as well? and not just World ID and EIP-1271 contract admin?

0xOsiris: We could do this honestly. In fact all admin types could be under the umbrella of EIP-1271. Even World ID

wips/wip-1001.md
0xOsiris: It's a pre-deploy. Sorry for all the confusing nomenclature; I will clean this up

wips/wip-1001.md
0xOsiris: Yes, this is correct

0xOsiris:
> I think we definitely want to support native P256 session keys for efficiency reasons.
> 
> Reasoning: Passkeys are the most userfriendly session key format and will thus be likely used in high frequency in the future.

What do you mean by `efficiency reasons`. Where is the efficiency lost in the current implementations support for P256?

wips/wip-1001.md
kilianglas: Seems to be the most elegant design tbh

kilianglas:
> What do you mean by efficiency reasons. Where is the efficiency lost in the current implementations support for P256?

We'd still have the 1271 overhead, correct? The native one would be Rust-only, while a 1271 module is executed in the EVM, right?

Edit: Since we are using pre-compiles, this is likely resolved.

wips/wip-1001.md
0xOsiris: I think the main purpose of this WIP is tightly defining the protocol level validation frames for client implementors. 1271 Ref Impls, and Crypto Precompile extensions are out of scope.

wips/wip-1001.md
0xOsiris: Updated

wips/wip-1001.md
0xOsiris: I disagree; Restricted validation frames is core to the design

wips/wip-1001.md
0xOsiris: Talked off of Github. For visibility we favor a single API rather than splitting the design between 1271 and native.

wips/wip-1001.md
0xOsiris: For visibility this is an application layer decision not a protocol layer. The current design allows full programmability which would allow World ID administrated accounts, however if we do not feel comfortable with this we should both not recommend or allow through the app.

Nevertheless Admin's have complete programmability over their Key Rings mutation authorization. So, it's not possible to stop. This is a core design principle

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
-A Session Proof proves that the caller controls the same WorldID that previously established the stored `sessionId`, without reusing the account creation action. The `sessionId` is the long-lived verification public identifier retained for the account across updates. The `sessionNullifier` is fresh per proof, passed only to `verifySession`, and discarded after verification.
+#### `create`
 
-##### OPRF Nullifier Proof Primer
+`create` MUST:
 
-An `OPRFNullifierProof` is a [Groth16](https://eprint.iacr.org/2016/260) proof over [BN254](https://hackmd.io/@jpw/bn254) submitted on-chain as `uint256[5] = [a, b₀, b₁, c, merkle_root]`. The `WorldIDVerifier` reconstructs the 15 public inputs from calldata and on-chain registry lookups (notably `merkle_root` via `WorldIDRegistry.isValidRoot()` and `oprfPublicKey` via `OprfKeyRegistry[rpId]`) and invokes the Groth16 verifier.
+1. reject `adminSigner == address(0)`;
+2. require `adminSigner` and each initial session verifier to have deployed code;
+3. require `initialSessionVerifiers.length` to be in `[1, MAX_SESSION_VERIFIERS]`;
```
piohei, line 305: What do you think about using `[2, MAX_SESSION_VERIFIERS]`? I'm just thinking loudly if we want to force higher security restrictions here. I know it is more complicated for users but it is also a bid deal to lost World ID access.

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+14. otherwise commit the normal payload result, including a reverted payload result if the target call reverted.
 
-`adminAuthorization` is the byte-encoding of a P256 ECDSA signature `(r, s)` over the `Update` `signalHash`.
+Failure classes are:
 
-Update flow:
+| Failure | Nonce | Payload state/logs | Fee behavior |
+| --- | --- | --- | --- |
+| Envelope, account, balance, nonce, or authorization precheck failure | unchanged | none | transaction is invalid and not included |
+| Validation budget exhausted | unchanged | none | transaction is deferred |
```
piohei, line 464: Why transaction is deffered instead of rejected? For how long it will wait? As the transaction pool is ephemeral I think it is better to just reject instead of keep it in pool for unknown amount of time.

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+We define a native World Chain account type managed by the `WORLD_CHAIN_ACCOUNT_MANAGER` predeploy. Each account has one immutable [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) compliant admin signer and one active key ring of [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) compliant session verifiers.
 
-This WIP specifies three admin authority instantiations:
+We extend [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) with a new typed transaction envelope with type flag `0x1D`. `0x1D` transactions are executed with the World Chain account **Framed** as the sender. Session verifiers act as the transaction level signatories via programmable [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) smart contracts. Session verifiers dually function as signature verifiers, and session key policy evaluators.
```
alessandromazza98: why `Framed` is capitalized? it's not clear to me

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+```solidity
 struct Account {
-    uint8        adminType;             // ADMIN_TYPE_*
-    bytes        verificationPublicId;  // variable length per admin type
-    SessionKey[] sessionKeys;           // see Session Keys
-    uint64       adminNonce;            // monotonic; replay protection
-    // pendingUpdates: see Timelock
+    address adminSigner;
+    bytes32 accountSalt;
+    SessionVerifier[] sessionVerifiers;
```
alessandromazza98, line 88: note that you've mentioned `session verifiers` a lot of times in the document but you never actually define it. I suggest to specifically define what you mean before this section so that the reader doesn't have any doubt

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+        SessionVerifier[] calldata initialSessionVerifiers,
+        bytes calldata adminAuthorization
     ) external returns (address account);
 
-    /// @notice Apply an Add/Remove update. `adminAuthorization` layout is admin-type-specific.
-    function update(
+    function setKeyRing(
         address account,
-        KeyringUpdate calldata keyringUpdate,         // mode != Create
+        bytes32 expectedCurrentKeyRingHash,
```
alessandromazza98, line 256: why is this needed to set the new key ring?

wips/wip-1001.md
alessandromazza98: I wouldn't restrict this in protocol. The app-level can sort of "enforce" the use of 2 or more by just defaulting to 2 or more when creating the world chain account.

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
 
-- **(G1) Merkle membership.** The authenticator key $k$ is a leaf in the `WorldIDRegistry` tree at the declared root.
-- **(G2) Credential validity.** $\mathit{cred}$ bears a valid EdDSA-Poseidon2 signature over [BabyJubJub](https://eips.ethereum.org/EIPS/eip-2494) from the issuer identified by `issuerSchemaId`, and has not expired.
-- **(G3) OPRF correctness.** The OPRF response $R_{\mathrm{oprf}}$ was evaluated under the registered key for `rpId`, with a DLog equality proof binding blinded and unblinded responses.
-- **(G4) Deterministic nullifier derivation.**
+`setKeyRing` is the only post-creation method that mutates the key ring and MUST:
 
-$$q = \mathrm{Poseidon2}(\text{"World ID Query"}, \; i, \; \mathrm{rpId}, \; a)$$
+1. require `account` to exist;
+2. require `expectedCurrentKeyRingHash == keyRingHash`;
```
alessandromazza98, line 320: correlated to my above question. Why is this needed? Can you derive this data from onchain data nevertheless?

alessandromazza98 (approved):
left some comments, generally lgtm

wips/wip-1001.md
alessandromazza98: the reason is that the transaction is still valid. The only problem is that validation budges has been exhausted for this block. Therefore you can wait and re-try next block. reth mempool automatically evicts transactions after some time has elapsed, so I wouldn't worry about this

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
-
-Sharing a single keyring + transaction surface (`0x1D`, session keys, signal binding, timelock) across instantiations means session-key UX and integration surface are uniform regardless of admin type. A new admin instantiation reduces to specifying one verification primitive and one address-derivation rule.
+Programmable policies over Session Keys unlock many capabilities that simply are not possible on traditional accounts. Some examples include account subscription models, delegated agent payments with scoped permissions enforced by the VM, social recovery, policy based authentication (e.g. Multi Factor authentication for high value transactions). World Chain accounts are unique in that the policies constraining a Session Key are fully programmable. This allows for expressive policies, and authentication strategies to be composed and updated dynamically without prescriptive standards enforced by the protocol.
```
kilianglas, line 25: NIT: I guess social recovery would not happen at the session key level

wips/wip-1001.md
kilianglas: I think it's technically not needed, but a safety mechanism to prevent accidential updates. We use the same pattern for the World ID off-chain commitment

kilianglas (approved):

wips/wip-1001.md
kilianglas, line 25: We could think about adding concrete examples for policy and admine/ session key implementaitons to make the motivation section stronger.

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+| --- | --- |
+| `ecrecover` | Ethereum precompile |
+| `sha256` | Ethereum precompile |
+| `ripemd160` | Ethereum precompile |
+| `identity` | Ethereum precompile |
+| `modexp` | Ethereum precompile |
+| `bn254 add` | Ethereum precompile |
+| `bn254 scalar multiplication` | Ethereum precompile |
+| `bn254 pairing` | Ethereum precompile |
+| `secp256r1 verify` | [RIP-7212](https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md) |
```
Kemperino, line 76: this exists as ethereum precompile already since Fusaka

https://eips.ethereum.org/EIPS/eip-7951

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+| `BLOCK_VALIDATION_GAS_BUDGET` | `uint64` | TBD | Per-block gas budget reserved for `0x1D` validation calls. |
+| `MIN_VALIDATION_FAILURE_FEE` | `uint256` | TBD | Minimum wei charged to the account on validation failure. |
+| `MAX_EXECUTION_TRACE_BYTES` | `uint32` | TBD | Maximum ABI-encoded trace size passed to a session verifier. |
+| `MAX_PAYLOAD_DATA_BYTES` | `uint32` | TBD | Maximum length of `data` in a `0x1D` envelope. |
+| `MAX_ACCESS_LIST_ENTRIES` | `uint32` | TBD | Maximum number of [EIP-2930](https://eips.ethereum.org/EIPS/eip-2930) access-list entries in a `0x1D` envelope. |
+| `MAX_SESSION_SIGNATURE_BYTES` | `uint32` | TBD | Maximum length of the session `signature` field. |
+| `MAX_ADMIN_AUTHORIZATION_BYTES` | `uint32` | TBD | Maximum length of `adminAuthorization` calldata. |
+| `WORLD_ID_1271_SIGNER_FACTORY` | `address` | TBD | Default World ID EIP-1271 signer factory. |
+| `EDDSA_PRECOMPILE` | `address` | TBD | EdDSA verification precompile address and ABI. |
+| `BLS12_381_PRECOMPILE` | `address` | TBD | BLS12-381 verification precompile address and ABI. |
```
Kemperino, line 60: these already exist as well since pectra or fusaka
 
https://eips.ethereum.org/EIPS/eip-2537

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
+| `EXECUTION_TRACE_VALIDATION_GAS_LIMIT` | `uint64` | TBD | Fixed gas forwarded to `evaluateSessionPolicy`. |
+| `BLOCK_VALIDATION_GAS_BUDGET` | `uint64` | TBD | Per-block gas budget reserved for `0x1D` validation calls. |
+| `MIN_VALIDATION_FAILURE_FEE` | `uint256` | TBD | Minimum wei charged to the account on validation failure. |
+| `MAX_EXECUTION_TRACE_BYTES` | `uint32` | TBD | Maximum ABI-encoded trace size passed to a session verifier. |
+| `MAX_PAYLOAD_DATA_BYTES` | `uint32` | TBD | Maximum length of `data` in a `0x1D` envelope. |
+| `MAX_ACCESS_LIST_ENTRIES` | `uint32` | TBD | Maximum number of [EIP-2930](https://eips.ethereum.org/EIPS/eip-2930) access-list entries in a `0x1D` envelope. |
+| `MAX_SESSION_SIGNATURE_BYTES` | `uint32` | TBD | Maximum length of the session `signature` field. |
+| `MAX_ADMIN_AUTHORIZATION_BYTES` | `uint32` | TBD | Maximum length of `adminAuthorization` calldata. |
+| `WORLD_ID_1271_SIGNER_FACTORY` | `address` | TBD | Default World ID EIP-1271 signer factory. |
+| `EDDSA_PRECOMPILE` | `address` | TBD | EdDSA verification precompile address and ABI. |
```
Kemperino, line 59: i dont think we need eddsa for anything

Kemperino: sure we can add it, but i see this as a completely separate task to add this.

wips/wip-1001.md
```diff
@@ -1,600 +1,556 @@
...
 struct Account {
-    uint8        adminType;             // ADMIN_TYPE_*
-    bytes        verificationPublicId;  // variable length per admin type
-    SessionKey[] sessionKeys;           // see Session Keys
-    uint64       adminNonce;            // monotonic; replay protection
-    // pendingUpdates: see Timelock
+    address adminSigner;
+    bytes32 accountSalt;
+    SessionVerifier[] sessionVerifiers;
+    bytes32 keyRingHash;
```
Kemperino, line 89: is this hash in the state needed? we could compute once based on sessionVerifiers and keep in cache on the node.

wips/wip-1001.md
0xOsiris: I meant to italicize as this is a key word in EIP 8141

wips/wip-1001.md
0xOsiris: Yeah the motivation section could use some work

wips/wip-1001.md
0xOsiris: Yes

wips/wip-1001.md
0xOsiris: I think it's nice to have as it determines the validity of signatures

wips/wip-1001.md
0xOsiris: I added it as an activation parameter as it's related to signing capabilities, but I intentionally did not specify its behavior. Which should be a WIP of its own.

There's already ticket here to spec out the EdDSA precompile: https://linear.app/worldcoin/issue/PROTO-4576/eddsa-verify-precompile-wip

wips/wip-1001.md
0xOsiris: In the OVM?

wips/wip-1001.md
0xOsiris:
```suggestion
We extend [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) with a new typed transaction envelope with type flag `0x1D`. `0x1D` transactions are executed with the World Chain account *framed* as the sender. Session verifiers act as the transaction level signatories via programmable [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) smart contracts. Session verifiers dually function as signature verifiers, and session key policy evaluators.
```

wips/wip-1001.md
0xOsiris: > i dont think we need eddsa for anything

World ID Authenticators are ed25519 signers. XMTP also uses ed25519. It's certainly useful, and has been asked for directly.

wips/wip-1001.md
Kemperino: yes: https://specs.optimism.io/protocol/precompiles.html 
https://www.evm.codes/precompiled
this is a standard precompile.

0xOsiris: Thanks, I didn't know this

wips/wip-1001.md
0xOsiris: Isn't it defined directly below?

PR #628, opened 2026-05-09
feat: WorldChainSpec

0xForerunner:
Adds support for a custom `WorldChainSpec`


crates/chainspec/src/builder.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub fn jovian_activated(mut self) -> Self {
+        self = self.isthmus_activated();
+        self.inner = self
+            .inner
+            .with_fork(WorldChainHardfork::Jovian, ForkCondition::Timestamp(0));
+        self
+    }
+
+    /// Enable Tropo at genesis.
+    pub fn tropo_activated(mut self) -> Self {
```
0xOsiris, line 163: What is this?

crates/chainspec/src/hardfork.rs
```diff
@@ -0,0 +1,135 @@
...
+        Granite,
+        /// Holocene: OP Stack Holocene upgrade.
+        Holocene,
+        /// Isthmus: OP Stack Isthmus upgrade.
+        Isthmus,
+        /// Jovian: OP Stack Jovian upgrade. World Chain is already on this hardfork.
+        #[default]
+        Jovian,
+        /// Tropo: the first World Chain specific hardfork after Jovian.
+        Tropo,
```
0xOsiris, line 31: Do we want these activated at Genesis on the default ChainSpec? I think we can probably just remove for now

crates/chainspec/src/builder.rs
0xForerunner: Our hard forks obviously haha.

crates/chainspec/src/hardfork.rs
0xForerunner: > Do we want these activated at Genesis on the default ChainSpec

They are not activated by default. We need to add them so we can start making changes.

0xForerunner: We can't put in any conditional logic unless the hard forks are defined.

0xOsiris (approved):
This generally looks fine to me. I would remove `tropo/strato` or leave inactive by default. I would also change the default ChainSpec in the `CHAIN_SPEC` of the e2e tests to `WorldChainSpec` to catch any issues in parity between `OpChainSpec`

crates/chainspec/src/builder.rs
0xOsiris: You are activating at timestamp 0 which makes them active on the chain spec by default. Instead the chainspec builder should activate the hard forks in the e2e test context. So, we don't activate unwanted things on our default node

crates/chainspec/src/hardfork.rs
0xOsiris: Yes, but regardless they should be inactive by default

crates/chainspec/src/builder.rs
0xForerunner: This is just the builder api? You would have to actually call this function to enable it, which we're not.

0xForerunner:
> This generally looks fine to me. I would remove `tropo/strato` or leave inactive by default. I would also change the default ChainSpec in the `CHAIN_SPEC` of the e2e tests to `WorldChainSpec` to catch any issues in parity between `OpChainSpec`

[crates/test-utils/src/e2e_harness/setup.rs:435](https://github.com/worldcoin/world-chain/blob/fr/world-chain-spec/crates/test-utils/src/e2e_harness/setup.rs#L180-L180)

PR #630, opened 2026-05-11
chore: add execution trace construction section to wip-1001

kilianglas:
## Summary

Adds a new `### Execution Trace Construction` subsection to `wips/wip-1001.md` under `## Specification`, between `### Restricted Validation Frames` and `### Transaction Type 0x1D`. The section specifies how implementations build the `WorldChainExecutionTrace` passed to `evaluateSessionPolicy`:

- inspector configuration (which fields to record, which to omit, why);
- struct layout for `CallTrace`, `LogTrace`, `WorldChainExecutionTrace` using a packed-blob encoding (fixed-size structs + `(offset, length)` pairs into per-role flat byte blobs);
- field-by-field construction rules (pre-order traversal, blob append order, invariants);
- rationale (why bytes are exposed rather than hashed, ABI overhead amortisation, CALLCODE inclusion, fan-out trade-off).

The proposed struct layout supersedes the corresponding definitions in `### Signer Interfaces`. This PR notes the conflict but intentionally does not modify `### Signer Interfaces`, to keep the diff additive and minimise merge friction with in-flight edits on this branch. The interface section should be updated in a follow-up.

## Key design choice — bytes vs hashes

The current `### Signer Interfaces` struct definitions commit calldata and return data via `keccak256` (`inputHash`, `outputHash`, `dataHash`). This limits the verifier to commit-and-compare against precomputed constants, which only expresses fully fixed-template policies. Policies whose arguments vary at submission time (allowance caps, recipient allowlists, swap slippage) are inexpressible.

The proposed encoding exposes raw bytes via three flat blobs (`inputsBlob`, `outputsBlob`, `logDataBlob`) with `(offset, length)` pairs in each `CallTrace` / `LogTrace`. This costs roughly 10–40% more bytes than the hash-only layout on typical workloads and grows linearly with total subcall data. A verifier that wants the commit-and-compare primitive can compute `keccak256(blob[off:off+len])` itself.

Sizing note: `MAX_EXECUTION_TRACE_BYTES` should be set with the per-byte cost in mind (suggested range 32–64 KB). Wide-fan-out applications can still hit the ceiling and should split into multiple `0x1D` transactions; this is intentional and bounds verifier decode work within `EXECUTION_TRACE_VALIDATION_GAS_LIMIT`.

### DIFF
--- a/wips/wip-1001.md
+++ b/wips/wip-1001.md
@@ -14,7 +14,7 @@ requires: EIP-1271, EIP-2718, EIP-1559, EIP-2930, ERC-7562, RIP-7212
 
 WIP-1001 defines a native World Chain account type managed by the `WORLD_CHAIN_ACCOUNT_MANAGER` predeploy. Each account has one immutable [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) compliant admin signer and one active keyring of EIP-1271 compliant session verifiers.
 
-We extend [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) with a new typed transaction envelope with type flag `0x1D`. `0x1D` transactions executed with the World Chain account **Framed** as the sender. Session verifiers act as the transaction level signatories via programmable [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271). Session verifiers dually function as signature verifiers, and session key policy evaluators.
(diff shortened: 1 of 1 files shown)


alessandromazza98:
Is this something that needs further work? Or is it stale and can we close it?

alessandromazza98:
I think we can close this @kilianglas as we are cancelling wip1001 work

PR #646, opened 2026-05-14
feat: add SP1 and Nitro TEE programs

0xForerunner:
Adds two programs: one is using SP1 to create a proof, second one is using Nitro TEE to create attestation.
This PR does not include any connection to on-chain contracts to verify proof or attestation. It also does not have anything connected to proposer or challenger. Programs are run using a CLI binary.
Due to some special requirements (like additional tooling for SP1 or Linux environment required for Nitro TEE) now the build flow is more complicated.
Both builds are hidden behind features in cargo.


0xOsiris:
yoo aha

PR #661, opened 2026-05-20
fix: WIP-1001: Store Verifier Address in instance look up to enable easy validation

Kemperino:
otherwise it would not have been possible with just instance -> keyRingHash lookup mapping

---

> [!NOTE]
> **Medium Risk**
> Updates the WIP-1001 spec’s session-verifier authorization/dispatch rules, which impacts how clients and the account router validate and route `0x1D` transactions. Risk is moderate because it changes core validation semantics, though it is documentation/spec-only in this PR.
> 
> **Overview**
> Clarifies and extends WIP-1001 to require an on-chain resolver mapping `sessionVerifierInstance -> (verifier, keyRingHash)` (`InstalledSessionVerifier`) so the account router can dispatch session validation without trusting any verifier address from transaction calldata.
> 
> Updates the spec to treat session verifier instances as **active only when their stored `keyRingHash` matches the current account `keyRingHash`**, and propagates this resolver requirement through router dispatch rules, `create`/`setKeyRing` state transition steps, manager authorization helpers (`isAuthorizedSessionVerifier`/`getAuthorizedSessionVerifier`), tx inclusion prechecks, txpool revalidation conditions, and test vector expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a73b8faef895cc8ac2392e900277a5921bc00e6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wips/wip-1001.md
+++ b/wips/wip-1001.md
@@ -87,6 +87,7 @@ struct Account {
     WorldChainAccountVerifier admin;
     bytes32 accountSalt;
     WorldChainAccountVerifier[] sessionVerifiers;
+    mapping(bytes32 sessionVerifierInstance => InstalledSessionVerifier) sessionVerifierInstances;
     bytes32 keyRingHash;
     uint64 adminNonce;
     uint64 transactionNonce;
@@ -95,6 +96,11 @@ struct Account {
 struct WorldChainAccountVerifier {
     address verifier;
     bytes installation;
+}
+
+struct InstalledSessionVerifier {
+    address verifier;
+    bytes32 keyRingHash;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


wips/wip-1001.md
```diff
@@ -87,6 +87,7 @@ struct Account {
     WorldChainAccountVerifier admin;
     bytes32 accountSalt;
     WorldChainAccountVerifier[] sessionVerifiers;
```
Kemperino, line 89: we actually in the implementation PR didnt store this for efficiency, but rather in a mapping like below

0xOsiris (approved):

PR #677, opened 2026-05-27
feat(exex): implement the op-proposer as a reth execution extension

0xOsiris:
> [!NOTE]
> **High Risk**
> Handles L1 signing keys and submits dispute-game transactions that anchor L2 state; incorrect output roots or proposal timing would be consensus/security relevant.
> 
> **Overview**
> Adds a new **`world-chain-exex`** crate that embeds the **OP Proposer** (L2 output submitter for the Dispute Game Factory) as a **reth execution extension**, aligned with upstream `op-proposer` v1.16.3-rc.1 (single-chain / pre-interop only).
> 
> When enabled via **`--proposer.*` / `OP_PROPOSER_*`**, the ExEx builds an L1 wallet provider (fallback RPC, retries, cached nonces, custom gas filler), runs a polling driver that calls **`DisputeGameFactory.create`**, and persists the last proposal and ExEx head in **MDBX**. Output roots are computed **in-process** from L2 headers (post-Isthmus `withdrawalsRoot` as message-passer storage root); optional **`--proposer.rollup-rpc`** keeps the rollup-RPC source path. **Admin RPC** (`admin_startProposer` / `admin_stopProposer`) and **Prometheus-style metrics** are included.
> 
> Workspace wiring adds **`crates/exex`**, **`reth-exex` / `reth-libmdbx`**, **`alloy-node-bindings`**, lockfile updates for versioned **`alloy-hardforks`**, and **`MACOSX_DEPLOYMENT_TARGET=11.0`**. **Ignored Anvil e2e tests** cover propose, dedup, and MDBX restart.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79a2de3ff50ee159a4d530c4833e8f28ad6bf138. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Kemperino (commented):
if we start going this route we should zoom out and think what else could/should be an ExEx

0xOsiris:
> if we start going this route we should zoom out and think what else could/should be an ExEx

We could potentially build the op-batcher as an ExEx, and well as a Withdrawal Proof Cacher for Withdrawal relays. These would all be quite a simplification from the current stack

0xOsiris:
> > if we start going this route we should zoom out and think what else could/should be an ExEx
> 
> We could potentially build the op-batcher as an ExEx, and well as a Withdrawal Proof Cacher for Withdrawal relays. These would all be quite a simplification from the current stack

Actually withdrawal proof relayers could simply be an ExEx

0xForerunner:
+4k lines, I don't think this makes sense to add unless we actually unlock something useful here.

PR #685, opened 2026-05-28
feat: multi proofs

0xOsiris:
> [!NOTE]
> **High Risk**
> New L1 dispute/finality contracts, bonded proposals, and anchor updates are security-critical; devnet also uses well-known signing keys for the proposer.
> 
> **Overview**
> Introduces **WIP-1006** multi-lane L1 proof games (factory, per-root games, anchor registry, mock verifiers/staking) and wires them into the **native HA devnet** by default.
> 
> New Rust crates **`world-chain-proofs`** (domain hashing, `proposalKey`/`rootId`, ABI bindings) and **`world-chain-proposer`** walk the anchor chain, fetch OP output roots via `optimism_outputAtBlock`, and submit bonded `propose` txs. **`full_stack`** deploys contracts with Foundry (`DeployProofSystem.s.sol`), then runs an in-process proposer; topology gains `WorldProofSystem` / `WorldChainProposer` components. **`--no-proof-system`** disables deployment; **`wip-1006.md`** documents `proposalKey` vs `rootId`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af63e6e8075fc88b7c6d2cf0ef8f3c6bcd47a040. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
```diff
@@ -0,0 +1,93 @@
...
+
+    function setGameBlacklisted(address game, bool blacklisted) external onlyOwner {
+        blacklistedGames[game] = blacklisted;
+        emit GameBlacklistedSet(game, blacklisted);
+    }
+
+    function setGameRetired(address game, bool retired) external onlyOwner {
+        retiredGames[game] = retired;
+        emit GameRetiredSet(game, retired);
+    }
```
alessandromazza98: what's the difference between these two functions? they accomplish the same goal, so I think we can remove one

pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
```diff
@@ -0,0 +1,93 @@
...
+        if (blacklistedGames[game]) revert GameBlacklisted(game);
+        if (retiredGames[game]) revert GameRetired(game);
+
+        IWorldChainProofSystemGame proofGame = IWorldChainProofSystemGame(game);
+        if (proofGame.state() != IWorldChainProofSystemGame.RootState.FINALIZED) {
+            revert GameNotFinalized(game);
+        }
+
+        address parent = proofGame.parentRef();
+        if (parent != address(this) && !acceptedGames[parent]) {
```
alessandromazza98, line 65: I'd add a brief explanation about why we're checking `parent != address(this)`. I'm assuming this is needed for the first game, is it correct?

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -0,0 +1,251 @@
...
+            state = WorldChainProofLib.RootState.CHALLENGED;
+            challengedAt = uint64(block.timestamp);
+            proofDeadline = uint64(block.timestamp + proofPeriod);
+        }
+
+        emit Challenged(msg.sender, proofDeadline);
+    }
+
+    function submitProofLane(uint8 laneId, bytes calldata proof) external {
+        if (state != WorldChainProofLib.RootState.CHALLENGED) {
```
alessandromazza98, line 150: I'd remove this if statement. Otherwise we're saying that if the game is not challenged, then the proposer cannot add proofs to support the game. Instead I think for example the prover could (almost) always also submit the tee proof (as it's cheaper) so that a challenger is way less incentivized to even challenge because there is already 1 valid proof supporting the game

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -0,0 +1,251 @@
...
+            uint256 amount = challengerBonds[challenger];
+            challengerBonds[challenger] = 0;
+            _transfer(payable(challenger), amount);
+        }
+
+        uint256 forfeited = address(this).balance;
+        if (forfeited != 0 && challengers.length != 0) {
+            _transfer(payable(challengers[0]), forfeited);
+        } else if (forfeited != 0) {
+            _transfer(proposer, forfeited);
```
alessandromazza98, line 222: what is the scenario in this path? no one has challenged the game but no one has even finalized it?

alessandromazza98: I'd add a brief comment to explain this. It's fine but it's also a very rare scenario

alessandromazza98 (approved):
LGTM

I only reviewed the world chain specific contracts and world chain bindings and they LGTM. I'm down to merge this.

Do we need all the base code? I'd prefer to not include it in this PR and eventually add what we really need in follow up PR.

Still approving but I'd remove base code and contracts unless there is a reason to keep it. We can still take inspiration but it's not necessary to keep their code inside our repo.

alessandromazza98:
take a look at https://github.com/worldcoin/world-chain/pull/687 - I think we can merge it inside this and then merge this one as well into main. At that point we can start working toward the needed services.

I think we can have 3 services:

- proposer: only posts l2 roots
- challenger: actively monitor games and eventually challenge when there is an invalid one
- defender: monitor active games that are challenged. If they are valid, provide the required proofs.

The reason to detach the defender from the proposer is that this way the proposer can be a very simple and lightweight binary. And most of the times this will be enough. Then we have the defender that is responsible to post proofs (if needed)

piohei:
I'm not sure if it should be part of the Game here but for proofs also World Chain protocol config is included (harforks' timestamps) https://github.com/worldcoin/world-chain/blob/fr/prover/crates/proof/core/src/range.rs#L224.

alessandromazza98 (approved):
I'm down to merge this and work iteratively on this. We're not touching production code, therefore there is no risk in merging this into main right now

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -0,0 +1,252 @@
...
+        emit Challenged(msg.sender, proofDeadline);
+    }
+
+    function submitProofLane(uint8 laneId, bytes calldata proof) external {
+        if (state == WorldChainProofLib.RootState.FINALIZED) {
+            revert AlreadyFinalized();
+        }
+        if (block.timestamp >= proofDeadline) {
+            revert ProofPeriodElapsed(block.timestamp, proofDeadline);
+        }
```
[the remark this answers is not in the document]
alessandromazza98: fixed thanks

PR #686, opened 2026-05-28
feat(consensus): integrate kona as an in-process consensus engine

0xOsiris:
> [!NOTE]
> **High Risk**
> Touches consensus/sequencing, engine forkchoice/payload handling, and HA devnet wiring; misconfiguration or engine/IPC integration bugs could stall or fork the chain.
> 
> **Overview**
> This PR makes the **world-chain** binary a monolithic execution + consensus client: when **`--kona.enabled`** is set, the Optimism **Kona** stack runs in-process instead of a separate op-node, and Engine API work on the hot path goes through **`WorldChainKonaEngineClient`** into reth’s **`ConsensusEngineHandle`** / **`PayloadStore`** (no HTTP/JWT for FCU, `newPayload`, `getPayload`).
> 
> A new **`world-chain-kona`** crate wires **`KonaService`**, config, and CLI (**`KonaArgs`**, P2P, sequencer, L1/beacon, conductor, node RPC). Workspace **`Cargo.lock`** pulls in **`kona-*`** from `ethereum-optimism/optimism` (`op-reth/v2.2.0`) plus **libp2p** and related networking deps. Node add-ons spawn Kona after launch and require reth **IPC** for slower L2 reads.
> 
> The **full-stack devnet** drops containerized op-nodes: native sequencers start with in-process Kona (P2P bootstore, conductor RPC, admin sequencer bootstrap). **op-conductor** and HA flows now talk to each sequencer’s **kona RPC** (`optimism_syncStatus`, `admin_startSequencer`, etc.).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54a3f03152773820b13692e31a7100feb1b6dbcd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/cli/src/cli/kona.rs
```diff
@@ -0,0 +1,821 @@
...
+}
+
+/// P2P network configuration for the in-process Kona consensus node.
+///
+/// These flags mirror kona's `P2PArgs` (see `bin/node/src/flags/p2p.rs`) and allow operators to
+/// configure persistent P2P identity, bootnodes, peer scoring, gossip mesh parameters, discovery
+/// settings, and sequencer signing — all via the world-chain binary's CLI.
+#[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
+#[command(next_help_heading = "Kona P2P")]
+pub struct KonaP2PArgs {
```
0xForerunner, line 79: Why do we need to recreate this? Can we not just import the cli?

crates/devnet/src/full_stack.rs
```diff
@@ -46,9 +46,6 @@ use crate::{
 };
 
 const ANVIL_RPC_PORT: u16 = 8545;
-const OP_NODE_RPC_PORT: u16 = 9545;
```
0xForerunner, line 67: I'm not sure it makes sense to switch to Kona yet here.

crates/kona/src/client.rs
```diff
@@ -0,0 +1,400 @@
...
+// we only surface whether self-authorization is enabled, never the key material.
+impl std::fmt::Debug for FlashblocksAuthorizationNotifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FlashblocksAuthorizationNotifier")
+            .field("self_authorizing", &self.keys.is_some())
+            .finish_non_exhaustive()
+    }
+}
+
+impl FlashblocksAuthorizationNotifier {
```
0xForerunner, line 73: Why? Can't we just use --flashblocks.force-publish

0xForerunner: That's exactly what this is for

0xForerunner (changes requested):
This adds a lot of complexity and I'm entirely sure it's worth it. Adds a ton of glue code we have to maintain.

0xOsiris:
> This adds a lot of complexity and I'm entirely sure it's worth it. Adds a ton of glue code we have to maintain.

Could you be more specific on where the complexity is? The changes are quite simple, easy to read, and straightforward. This is less than 1k additions of core logic, and most of it is directly from Kona.

Eliminates all codec bugs (we've hit many), bugs cause by versioning mismatch introducing logical errors/crashes (we've hit many), bugs from dns, core dns, eliminates http communication in the hot path, eliminates serialization/deserialization, solves the issue of missing blocks from FCU's catching up to the L1 origin because of asynchronicity of the Engine API, simplifies deployment for node operators with a single binary, simplifies upgrade schedules for node operators, and allows us to own consensus in our stack (a top priority with hardforks ahead).

crates/kona/src/client.rs
0xOsiris: > Why? Can't we just use --flashblocks.force-publish

--force-publish is a flag that could be set or not by the node operator? If we are prescribing force publish then we should just remove the flag and make it the default path

crates/cli/src/cli/kona.rs
alessandromazza98: +1

crates/kona/src/client.rs
```diff
@@ -0,0 +1,400 @@
...
+        });
+
+        self.to_jobs_generator
+            .send_modify(|slot| *slot = authorization);
+    }
+}
+
+/// An in-process Engine API client that bridges Kona's consensus layer to reth's execution engine
+/// for the consensus hot path, without any network transport.
+pub struct WorldChainKonaEngineClient<Engine: PayloadTypes = OpEngineTypes> {
```
alessandromazza98, line 94: nit: i'd remove the generic. We don't have plan to change the engine types at least for now, so I'd simplify this by always using OpEngineTypes

alessandromazza98 (commented):
For medium / long term future I also agree that merging EL + CL in the same binary is a win for us.

I'm worried if this is the best thing to do right now. It will definitely introduce some issues initially and we would need to spend time to fix them.

And we have another priority now - i.e. make the new proof system.

What about:

- make the proof system
- make sure it works properly with kona running as separate binary
- once it's shipped to mainnet and kona is working as consensus client in a separate binary
- at that point we can merge this and work toward fixing all eventual issues

0xOsiris:
> For medium / long term future I also agree that merging EL + CL in the same binary is a win for us.
> 
> I'm worried if this is the best thing to do right now. It will definitely introduce some issues initially and we would need to spend time to fix them.
> 
> And we have another priority now - i.e. make the new proof system.
> 
> What about:
> 
> - make the proof system
> - make sure it works properly with kona running as separate binary
> - once it's shipped to mainnet and kona is working as consensus client in a separate binary
> - at that point we can merge this and work toward fixing all eventual issues

Why medium to long term only? It's an optional flag, and it's working. We can, and should do both.

crates/cli/src/cli/kona.rs
0xOsiris: Because the default kona aliases collide with reth args. Crashing on startup. We've prefixed all args with `kona` to avoid collision

`id = "kona.no_discovery"`

crates/devnet/src/full_stack.rs
0xOsiris: What are we waiting for

crates/kona/src/service.rs
```diff
@@ -0,0 +1,753 @@
...
+        // before the safe head's, so it is already safe and the derived target can never exceed the
+        // safe head. (Strictly below, not at: consecutive L2 blocks usually share an L1 origin, so
+        // the safe head's own epoch may still contain an as-yet-unsafe block.)
+        //
+        // Only `BlockInfo::number` is consumed downstream — the L1 watcher forwards the block as-is
+        // and kona's finalizer keys solely off the number — so clamping the number suffices. In a
+        // synced node the finalized L1 block already trails the safe head's origin by far more than
+        // this, so the clamp is a no-op; it only engages while the node replays history behind the
+        // finalized point (e.g. a restart bootstrapping derivation from L1 with no unsafe-block
+        // gossip), which is exactly the window where the unclamped feed crashes the node.
```
alessandromazza98, line 455: I find this comment a bit hard to understand. I understand it now but I'd suggest to add a small example that explains it very fast. Something like:

```
// Example:
//
//   safe head L2      = #1000
//   safe head origin  = L1 block 100
//
// The node is replaying history and receives:
//
//   finalized L1 = 150
//
// Kona will try to finalize the highest L2 block derived from finalized L1 data.
// That block may be beyond the current safe head:
//
//   safe head      = L2 #1000
//   finalize target= L2 #1200   <-- Critical error
//
// To avoid this, we cap the finalized L1 block to strictly before the safe head's
// L1 origin:
//
//   finalized L1 = min(150, 100 - 1) = 99
//
// Why "100 - 1" instead of "100"?
//
// Multiple L2 blocks can share the same L1 origin:
//
//   L1 origin 100
//     ├─ L2 #1000 (safe head)
//     ├─ L2 #1001
//     └─ L2 #1002
//
// The safe head only guarantees that L2 #1000 is safe. Later L2 blocks from the same
// epoch may still be unsafe. By clamping to origin 99, we only finalize epochs that
// are completely before the safe head's epoch, guaranteeing the derived finalize
// target can never be ahead of the current safe head.
```

crates/kona/src/service.rs
```diff
@@ -0,0 +1,753 @@
...
+                        None => return Ok(()),
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
alessandromazza98, line 622: can you move this in the end of the file? Otherwise it may seem that from here forward everything is for tests, but right now this is not true

alessandromazza98 (approved):
left some non-blocking comments

PR #687, opened 2026-05-29
fix: remove vendored base

alessandromazza98:
This PR removes base vendor code + applies a couple of small fixes in world chain proof contracts


pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
```diff
@@ -15,7 +15,6 @@ contract WorldChainAnchorStateRegistry {
     event AnchorUpdated(address indexed game, bytes32 indexed rootId, bytes32 rootClaim, uint256 l2BlockNumber);
     event PausedSet(bool paused);
     event GameBlacklistedSet(address indexed game, bool blacklisted);
-    event GameRetiredSet(address indexed game, bool retired);
```
alessandromazza98, line 18: this is the only file that contains fixes that are not related to removing base vendor code

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -147,9 +147,6 @@ contract WorldChainProofSystemGame {
     }
 
     function submitProofLane(uint8 laneId, bytes calldata proof) external {
```
alessandromazza98, line 150: and this one

PR #693, opened 2026-05-29
fix: resolve simulation metadata from post-state

crystalt:
> [!NOTE]
> **Medium Risk**
> Changes simulation asset normalization and token metadata caching behavior. Scope is limited to read-only simulation output, but it affects how newly-created or initialized tokens are represented in normalized results.
>
> **Overview**
> Fixes token metadata (`name` / `symbol` / `decimals`) for `simulate_unsignedUserOp` by resolving metadata after applying the simulated transaction's state diff to the in-memory revm DB.
>
> This keeps simulation state non-persistent while allowing contracts created or initialized during the simulation to answer metadata view calls. Metadata caching remains backed by `AssetInfo`, but cached entries only copy metadata fields into each parsed change so ERC-20 / ERC-721 / ERC-1155 classification stays with the parser result.
>
> Adds a World Chain fork regression for the Datadog ANT simulation at block `30374933`, using the production calldata and block environment to assert the created token metadata matches `primaryNormalized` (`THE ANT`, `ANT`, `18`).

## Summary

- Resolve simulation token metadata after applying the simulated state diff to the in-memory DB.
- Keep the metadata cache on `AssetInfo`, but only apply cached `name`, `symbol`, and `decimals` so parsed asset type stays intact.
- Force-refresh metadata for contracts created during the simulation.
- Add a World Chain fork regression for simulation `621aad22-db31-46b7-9576-ef4502c58661` at block `30374933`, asserting the created ANT token metadata matches `primaryNormalized` (`THE ANT`, `ANT`, `18`).

## Root Cause

Metadata was resolved against the original fork/block state, before the simulated transaction's state changes were applied. For tokens created or initialized inside the simulation, metadata calls could not see the post-simulation contract state, causing normalized output to miss or mismatch token metadata.

## Notes

The fork regression pins the production block and block environment used by the Datadog example so the replay matches the real simulation path. The calldata is kept inline in the test for a self-contained repro.

## Validation

- `cargo +1.95.0 test -p world-chain-rpc --lib metadata`
- `source ~/.zshrc && cargo +1.95.0 test -p world-chain-rpc --test fork_simulate test_simulate_new_token_reads_post_state_for_normalized_asset -- --nocapture`


crates/rpc/tests/fixtures/ant_metadata_calldata.txt
```diff
@@ -0,0 +1,33 @@
+0x7bb3742800000000000000000000000038869bf66a61cf6bdb996a6ae40d5853fd43b526000000
```
karankurbur: put it in line for integration test, dont need a new file

karankurbur (approved):

PR #694, opened 2026-05-29
feat(wips): add wip for flashblock access lists

0xOsiris:
> [!NOTE]
> **Low Risk**
> Documentation-only WIP addition with no runtime or consensus code changes in this PR.
> 
> **Overview**
> Adds **WIP-1007**, a draft Networking standards track that defines **Flashblock Access Lists (FAL)** for sub-block flashblock pre-confirmations.
> 
> The WIP specifies an optional **`access_list_data`** field on **`ExecutionPayloadFlashblockDeltaV1`**: EIP-7928-style account/storage change records keyed by **block access index**, plus **`keccak256(rlp(access_list))`**, with canonical ordering and pruning rules. Builders may attach it; consumers with the field present can reconstruct per-tx pre-state, validate via parallel re-execution, and derive post-state roots from the diff, while absent payloads keep the legacy sequential path.
> 
> It also documents **Amsterdam** header reconstruction using **`EMPTY_BLOCK_ACCESS_LIST_HASH`** (sidecar not committed on-chain), builder recording rules, mandatory consumer verification steps, backwards compatibility, test expectations, and security notes (verify-before-trust, canonicalization, resource bounds).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11dfc81fafefbe806c0cc00de30ce61b927d5aa7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wips/wip-1007.md
+++ b/wips/wip-1007.md
@@ -12,7 +12,7 @@ requires: EIP-7928
 
 ## Abstract
 
-A Flashblock Access List (FAL) is an [EIP-7928][eip-7928]-style block access list scoped to a single flashblock. The builder records every account, balance, nonce, code, and storage change produced while executing a flashblock's transactions — each tagged with a block-local *block access index* — and publishes the resulting list, together with the Keccak256 hash of its RLP encoding, as an optional field on the flashblock pre-confirmation payload propagated over the Flashblocks P2P network.
+A Flashblock Access List (FBAL) is an [EIP-7928][eip-7928]-style block access list scoped to a single flashblock. The builder records every account, balance, nonce, code, and storage change produced while executing a flashblock's transactions - each tagged with a block-local *block access index* — and publishes the resulting list, together with the Keccak256 hash of its RLP encoding, as an optional field on the flashblock pre-confirmation payload propagated over the Flashblocks P2P network.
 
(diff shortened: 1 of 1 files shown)


alessandromazza98 (commented):
do we want a wip for block access list? even if there is the official eip. I know that the current implementation slightly differs but we've not committed to this impl for prod, so we may end up directly using the official eip one as it's soon going to be merged for glamsterdam

alessandromazza98 (approved):
approving to not block you, but I don't see a big utility for this wip

0xOsiris:
> approving to not block you, but I don't see a big utility for this wip

It's simply to document the implementation. Do you not think we should document it?

alessandromazza98:
> It's simply to document the implementation. Do you not think we should document it?

It's not a big deal, let's document and merge this

0xOsiris:
> do we want a wip for block access list? even if there is the official eip. I know that the current implementation slightly differs but we've not committed to this impl for prod, so we may end up directly using the official eip one as it's soon going to be merged for glamsterdam

The official EIP is an objectively worse implementation just in terms of performance. We can use a combination of both, where the block access list hash in the block header actually commits to the incremental access list of the stream.

PR #698, opened 2026-06-02
feat: add SP1 and Nitro TEE programs

piohei:
Adds two programs: one is using SP1 to create a proof, second one is using Nitro TEE to create attestation.
This PR does not include any connection to on-chain contracts to verify proof or attestation. It also does not have anything connected to proposer or challenger. Programs are run using a CLI binary.
Due to some special requirements (like additional tooling for SP1 or Linux environment required for Nitro TEE) now the build flow is more complicated.
Both builds are hidden behind features in cargo. They also share as much code as possible.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI and gitignore; no runtime auth, consensus, or on-chain verification logic is modified in this diff.
> 
> **Overview**
> This diff tightens **CI and repo hygiene** around the SP1 proof stack (the broader PR adds SP1/Nitro programs separately).
> 
> A new **`.github/workflows/elf.yml`** workflow installs SP1 **v6.1.0**, runs `just build-proof-elfs`, and **fails if** `crates/proof/succinct/elf/` changes—so committed guest ELFs stay in sync with program sources. It runs on `main` pushes, path-filtered PRs under proof crates, and `workflow_dispatch`.
> 
> **`rust-ci.yml`** adds **`arduino/setup-protoc`** to the **cargo-tests** and **cargo-clippy** jobs so protobuf-dependent crates build in CI.
> 
> **`.gitignore`** now ignores nested **`target/`** trees (including SP1 program builds), **`/proofs/`**, and the succinct programs target path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bab7a2bcd84b0a18a2fbf47e3893333109965132. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/rust-ci.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - uses: actions/checkout@v6
       - uses: taiki-e/install-action@just
       - uses: dtolnay/rust-toolchain@stable
+      - uses: arduino/setup-protoc@v3
```
alessandromazza98, line 26: what is this for?

docs/proof/devnet-succinct-fault-proof.md
```diff
@@ -0,0 +1,101 @@
+# World OP Succinct Lite devnet rollout
+
+This is the shipping plan for the World Chain zk fault proof devnet deployment. It follows OP
+Succinct Lite's architecture: existing OP Stack services keep producing the chain, while separate
+Succinct proposer and challenger services participate in a new fault dispute game type.
+
+## What can stay vanilla
+
+- `op-geth`, `op-node`, and `op-batcher` can stay on the normal devnet deployment path.
```
alessandromazza98, line 9: nit: `op-geth` is soon deprecated and we can immediately put `world-chain` client there to replace it.

docs/proof/devnet-succinct-fault-proof.md
```diff
@@ -0,0 +1,101 @@
...
+- `aggregation_vkey`: SP1 aggregation verifying key for `world-chain-aggregation`.
+- `range_vkey_commitment`: SP1 range verifying key commitment for `world-chain-range-ethereum`.
+- `rollup_config_hash`: pretty-JSON SHA-256 of Kona `RollupConfig` plus World `tropo_time` and
+  `strato_time`.
+- `starting_l2_block_number` and `starting_root`: anchor state for the new game.
+- `game_type`: use the OP Succinct Lite game type selected for devnet, expected to be `42` unless
+  the devnet factory reserves another type.
+- `verifier_address`: SP1 verifier gateway on the devnet L1, or the mock verifier only for a
+  throwaway devnet.
+- `proposer_addresses` and `challenger_addresses`: addresses funded on Sepolia and backed by
```
alessandromazza98, line 41: depending on how we will finalize our work, we will probably have 3 actors:
- proposer
- challenger
- defender

bin/proof/src/main.rs
```diff
@@ -0,0 +1,1151 @@
...
+        holocene_time: config.holocene_time,
+        isthmus_time: config.isthmus_time,
+        jovian_time: config.jovian_time,
+        tropo_time: config.tropo_time,
+        strato_time: config.strato_time,
+    }
+}
+
+fn finalized_l2_head(client: &Client, rpc_url: &str) -> eyre::Result<Option<u64>> {
+    Ok(rpc::<RpcBlock>(
```
alessandromazza98: nit: all these rpc calls to an ethereum node could be done using alloy providers instead of raw rpc queries. What do you think? btw not blocking of course

bin/proof/src/main.rs
```diff
@@ -0,0 +1,1151 @@
...
+
+    println!(
+        "enclave returned: l2_pre={pre:?} l2_post={post:?} block={block}",
+        pre = artifact.boot_info.l2PreRoot,
+        post = artifact.boot_info.l2PostRoot,
+        block = artifact.boot_info.l2BlockNumber,
+    );
+
+    let expected_user_data = range_user_data(&artifact.boot_info);
+    verify_attestation_doc(
```
alessandromazza98: nit: this is already done inside the `prover.prove_range_async`

alessandromazza98 (approved):
huge work Piotr. It's ok for me to merge it as we're not touching any production code and we can keep working on it iteratively instead of keeping this huge PR.

Left some small non blocking comments

crates/proof/core/src/witness/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+        Self {
+            preimage_store,
+            blob_data,
+            schedule,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default, rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
+pub struct EigenDAWitnessData {
```
0xOsiris: What is this?

crates/proof/core/src/witness/preimage_store.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+/// Check that the preimage matches the expected hash.
+pub fn check_preimage(key: &PreimageKey, value: &[u8]) -> PreimageOracleResult<()> {
+    if let Some(expected_hash) = match key.key_type() {
+        PreimageKeyType::Keccak256 => Some(keccak256(value).0),
+        PreimageKeyType::Sha256 => Some(sha2::Sha256::digest(value).into()),
+        PreimageKeyType::Local | PreimageKeyType::GlobalGeneric => None,
+        PreimageKeyType::Precompile => unimplemented!("Precompile not supported in zkVM"),
```
0xOsiris: Why?

crates/proof/core/src/range.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// Strato activation timestamp. This is a World-only fork.
+    #[serde(default, alias = "stratoTime")]
+    pub strato_time: Option<u64>,
+}
+
+impl WorldRangeHardforkConfig {
+    /// Returns `true` when the fork is active at the given L2 block/timestamp.
+    pub fn is_active(&self, fork: WorldRangeHardfork, block_number: u64, timestamp: u64) -> bool {
+        match fork {
+            WorldRangeHardfork::Bedrock => block_number >= self.bedrock_block.unwrap_or(0),
```
0xOsiris: Activation of all other forks uses `is_some_and` except `bedrock_block` returns true on `None`

crates/proof/core/src/range.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+fn timestamp_active(activation: Option<u64>, timestamp: u64) -> bool {
+    activation.is_some_and(|activation| timestamp >= activation)
+}
+
+/// World hardfork names used inside range proof public values.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum WorldRangeHardfork {
```
0xOsiris: Can't we just add `#[repr(u8)]` here, so we don't have to duplicate below?

crates/proof/core/src/types.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// Converts an SP1 vkey digest from `[u32; 8]` to the bytes used on-chain.
+pub fn u32_to_u8(input: [u32; 8]) -> [u8; 32] {
+    let mut output = [0u8; 32];
+    for (i, value) in input.into_iter().enumerate() {
+        output[i * 4..(i + 1) * 4].copy_from_slice(&value.to_be_bytes());
+    }
+    output
+}
```
0xOsiris:
```suggestion
pub fn u32_to_u8(input: [u32; 8]) -> [u8; 32] {
    std::array::from_fn(|i| input[i / 4].to_be_bytes()[i % 4])
}
```

crates/proof/nitro/src/lib.rs
```diff
@@ -0,0 +1,312 @@
+//! AWS Nitro TEE-attested prover for the World Chain OP Succinct Lite fault-proof stack.
```
0xOsiris:
```suggestion
#![cfg_attr(not(test), warn(unused_crate_dependencies)]

//! AWS Nitro TEE-attested prover for the World Chain OP Succinct Lite fault-proof stack.
```

crates/proof/core/src/range.rs
```diff
@@ -0,0 +1,341 @@
+//! Shared range-proof public-value types used by all World fault-proof backends.
```
0xOsiris:
```suggestion
#![cfg_attr(not(test), warn(unused_crate_dependencies))]

//! Shared range-proof public-value types used by all World fault-proof backends.
```

crates/proof/succinct/utils/client/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+//! SP1 guest/host execution utilities for the World Chain OP Succinct Lite programs.
```
0xOsiris:
```suggestion
#![cfg_attr(not(test), warn(unused_crate_dependencies))]

//! SP1 guest/host execution utilities for the World Chain OP Succinct Lite programs.
```

crates/proof/succinct/utils/client/src/precompiles/custom.rs
```diff
@@ -0,0 +1,40 @@
+//! Custom crypto provider for KZG proof verification.
+
+use kzg_rs::{Bytes32, Bytes48, KzgProof, KzgSettings};
+use revm::precompile::{Crypto, PrecompileHalt};
+
+/// Custom cryptography provider using kzg-rs for KZG proof verification.
+#[derive(Debug)]
+pub struct CustomCrypto {
```
0xOsiris: Are there not re-usable libs we can use here instead of rolling our own

crates/proof/succinct/utils/client/src/precompiles/factory.rs
```diff
@@ -0,0 +1,115 @@
...
+                .with_cfg(CfgEnv::new_with_spec(OpSpecId::BEDROCK))
+                .with_chain(L1BlockInfo::default())
+                .with_db(db)
+                .with_block(input.block_env)
+                .with_cfg(input.cfg_env)
+                .build_op_with_inspector(NoOpInspector {})
+                .with_precompiles(PrecompilesMap::from_static(
+                    OpPrecompiles::new_with_spec(spec_id).precompiles(),
+                )),
+            false,
```
0xOsiris: Double check this

crates/proof/world-chain-proof/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+//! World Chain fault-proof library.
```
0xOsiris:
```suggestion
#![cfg_attr(not(test), warn(unused_crate_dependencies))]

//! World Chain fault-proof library.
```

crates/proof/succinct/utils/host/src/witness_generation/online_blob_store.rs
```diff
@@ -0,0 +1,56 @@
...
+use alloy_primitives::B256;
+use anyhow::Result;
+use async_trait::async_trait;
+use kona_derive::BlobProvider;
+use kona_protocol::BlockInfo;
+use kzg_rs::{Blob as KzgRsBlob, Bytes48};
+use world_chain_proof_core::witness::BlobData;
+
+#[derive(Clone, Debug)]
+pub struct OnlineBlobStore<T: BlobProvider> {
```
0xOsiris: Do we need to rebuild this type?

0xOsiris (changes requested):

0xOsiris:
Nice work. I think we should make sure we understand every line of this as a hidden bug could be extremely difficult to find after merging, and this code is quite important. Let's try to shrink this down to the minimal amount of hand rolled code re-using sp1 libraries where possible. Then we can merge

bin/proof/src/main.rs
```diff
@@ -0,0 +1,1151 @@
...
+use world_chain_proof_succinct_client_utils::{
+    OutputRootWitness,
+    witness::executor::{WitnessExecutor, get_inputs_for_pipeline},
+};
+use world_chain_proof_succinct_ethereum_client_utils::executor::ETHDAWitnessExecutor;
+use world_chain_proof_succinct_host_utils::witness_generation::{
+    OnlineBlobStore, PreimageWitnessCollector,
+};
+
+const L1_BLOCK_PREDEPLOY: Address = Address::new([
```
0xOsiris: Can we use the `address!` macro with raw hex strings. Much easier to read

crates/proof/nitro/src/attestation.rs
```diff
@@ -0,0 +1,322 @@
...
+    Ok(ParsedAttestationDoc {
+        pcrs,
+        user_data,
+        module_id,
+        digest,
+    })
+}
+
+/// Verifies that a parsed attestation document matches the supplied expectations.
+pub fn verify_attestation_doc(
```
0xOsiris: **Critical (known TODO, flagging for visibility):** `verify_attestation_doc` does no cryptographic verification — it parses the COSE_Sign1 envelope but never (a) verifies the signature, (b) validates the cert chain to the AWS Nitro root CA, or (c) checks the document timestamp. It only compares PCR/`user_data` bytes that the untrusted host supplies, so a fully fabricated document passes as long as the byte fields match.

The name promises verification and it's wired into the live host path (`host.rs:114`, `host.rs:150`), so a future caller will assume it's safe. Combined with the all-zero-PCR skip below and `nonce: None` in `enclave_lib.rs:278` (replayable), the attestation provides no guarantee today.

**Reuse:** replace the hand-rolled CBOR/COSE parsing with `aws-nitro-enclaves-cose` (signature) + `aws-nitro-enclaves-attestation` / `attestation-doc-validation` (cert chain + root-CA pinning + timestamp) — one `validate()` call covers all three gaps; keep only the PCR/`user_data` assertions on the returned struct. At minimum, rename to `parse_and_check_pcrs` until verification lands so the name doesn't promise more than it delivers.

crates/proof/nitro/src/attestation.rs
```diff
@@ -0,0 +1,322 @@
...
+        .pcrs
+        .get(&index)
+        .ok_or(AttestationError::MissingField(match index {
+            0 => "pcr0",
+            1 => "pcr1",
+            2 => "pcr2",
+            _ => "pcrN",
+        }))?;
+    // All-zero expected digest is the placeholder — skip verification.
+    if expected.iter().all(|&b| b == 0) {
```
0xOsiris: Any PCR whose **expected** value is all-zero is silently skipped, and `ExpectedPcrs::PLACEHOLDER` / `Default` (`lib.rs:73-83`) is all-zero. A caller that forgets to set real PCRs — or that hits the `Default` — verifies nothing, with no error. Make a placeholder/all-zero expected PCR a hard error rather than a skip.

crates/proof/core/src/boot.rs
```diff
@@ -0,0 +1,222 @@
...
+pub fn hash_rollup_config<T: Serialize + ?Sized>(
+    config: &T,
+) -> Result<B256, RollupConfigHashError> {
+    Ok(sha256_b256(
+        serde_json::to_string_pretty(config)?.as_bytes(),
+    ))
+}
+
+/// Public boot values committed by the range proof.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
0xOsiris: **Reuse / dedup (biggest one):** `core` and `protocol` re-declare the same types verbatim:

| type/fn | `core` | `protocol` |
|---|---|---|
| `BootInfoPublicValues` | `boot.rs:28` | `lib.rs:458` |
| `RollupConfigHashError` | `boot.rs:11` | `lib.rs:23` |
| `hash_rollup_config` + `sha256_b256` | `boot.rs:19,165` | `lib.rs:35,63` |
| `WorldRollupConfigHashInput` | `boot.rs:130` | `lib.rs:70` |
| World hardfork config | `range.rs:8` | `lib.rs:84` |
| World spec-id enum | `range.rs:126` | `lib.rs:285` |

`main.rs:688` even has a converter whose only job is to copy 11 fields between the two duplicates. Pick one source of truth: the hash helpers / `BootInfoPublicValues` / `RollupConfigHashError` are pure and `no_std`-able — put them in `core` and re-export from `protocol`. The spec-id enums are identical numeric enums and can be unified (the reth `ForkCondition` dep only blocks sharing the *config struct*, not the enum). This also removes the host/guest `active_fork_at` divergence risk noted on `range.rs`.

crates/proof/core/src/boot.rs
```diff
@@ -0,0 +1,222 @@
...
+    tropo_time: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    strato_time: Option<u64>,
+}
+
+/// Hashes the rollup config committed by World range proofs.
+///
+/// This mirrors OP Succinct's pretty-JSON-then-SHA256 hash, but appends World-only fork fields
+/// that are not represented in upstream Kona's `RollupConfig`.
+pub fn hash_world_rollup_config(
```
0xOsiris: `.expect(...)` on the critical proof path (called from `BootInfoStruct::from_kona_boot_info` at line 83). The fallible generic variant `hash_world_rollup_config_generic` is defined right below — return `Result` and propagate instead of panicking.

crates/proof/core/src/oracle/blob_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+            .collect();
+
+        match kzg_rs::KzgProof::verify_blob_kzg_proof_batch(
+            blobs,
+            value.commitments,
+            value.proofs,
+            &get_kzg_settings(),
+        ) {
+            Ok(true) => {}
+            Ok(false) => panic!("KZG proof verification failed: invalid proofs"),
```
0xOsiris: Two `panic!`s inside a `From` impl on a KZG batch-verification failure, plus `.unwrap()` on a malformed blob length (line ~22) and on `versioned_blobs.pop()` (line ~59). These run on beacon/network-controlled data. A `From` impl can't fail gracefully — make this `TryFrom` (or a fallible free fn) and thread the error back. Violates the workspace no-`unwrap` rule.

crates/proof/succinct/utils/client/src/client.rs
```diff
@@ -0,0 +1,202 @@
...
+
+                // Adjust the target block number to the current safe head, as no more blocks
+                // can be produced.
+                if target.is_some() {
+                    target = Some(tip_cursor.l2_safe_head.block_info.number);
+                };
+
+                // If we are in interop mode, this error must be handled by the caller.
+                // Otherwise, we continue the loop to halt derivation on the next iteration.
+                if cfg.is_interop_active(driver.cursor.read().l2_safe_head().block_info.number) {
```
0xOsiris: `is_interop_active` takes a **timestamp**, but this passes `l2_safe_head().block_info.number`. Upstream kona passes `.block_info.timestamp`. Copy-from-upstream bug — makes the interop guard effectively always-false. Low real impact for World (interop/Karst → `Never`), but worth fixing to stay in sync. (This whole function is a near-verbatim copy of `kona_driver::Driver::advance_to_target`; the only real additions are the zkVM cycle-tracker `println!`s — consider wrapping `driver.advance_to_target()` instead of forking it.)

crates/proof/succinct/utils/ethereum-client/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "world-chain-proof-succinct-ethereum-client-utils"
+version = "2.2.0"
```
0xOsiris: This is a workspace member but hardcodes `version`/`edition`/`rust-version` and pins `kona-*` to literal `tag = "op-reth/v2.2.0"` instead of `*.workspace = true` like the sibling crates. When the workspace bumps kona, this crate stays pinned → two copies of kona types and silent ABI/type-mismatch. Switch to workspace inheritance.

0xOsiris (commented):
Review focused on (1) correctness bugs and (2) reusing existing kona/op-succinct/AWS libs rather than re-vendoring. Inline comments below cover the big-ticket items; a number of smaller `.unwrap()`/`.expect()` cleanups are noted but not commented individually.

0xOsiris:
/otto address this PR review with GPT 5.5 Thinking

crates/proof/core/src/witness/mod.rs
0xOsiris: We don't need this

0xOsiris:
/otto fix the ci issues as well with opus 4.8

crates/proof/succinct/utils/client/src/precompiles/custom.rs
0xOsiris: > Why not pull this from kona / op-succinct? kona-executor and op-succinct's client utils don't currently expose a revm::precompile::Crypto plugin — they use Op*Precompiles directly and patch KZG at the precompile-id level rather than at the Crypto trait. Since revm now offers Crypto as the pluggable seam, this is the smallest custom surface to swap blst → kzg-rs.

/otto let's do this

0xOsiris:
> Why not pull this from kona / op-succinct? kona-executor and op-succinct's client utils don't currently expose a revm::precompile::Crypto plugin — they use Op*Precompiles directly and patch KZG at the precompile-id level rather than at the Crypto trait. Since revm now offers Crypto as the pluggable seam, this is the smallest custom surface to swap blst → kzg-rs.

/otto address this

0xOsiris:
/otto address this with opus 4.8

https://github.com/worldcoin/world-chain/pull/698#discussion_r3351712323

0xOsiris (approved):

PR #740, opened 2026-06-15
fix: record incremental metrics correctly

0xOsiris:
> [!NOTE]
> **Low Risk**
> Observability-only change to how fee values are derived and labeled; no payload building, publishing, or fee calculation logic is altered.
> 
> **Overview**
> **Fixes committed payload fee metrics** so per-block revenue can be computed correctly when a block is built across multiple flashblock commits.
> 
> Previously each commit recorded the **cumulative** `payload.fees()` into `committed_payload_fees_wei`, so aggregating the histogram counted the running block total once per flashblock and overstated revenue. The histogram now records **marginal** fees (`cumulative_fees - prev_committed_fees`), captured from `committed_payload` before the state is overwritten on commit. The **`committed_payload_fees_wei_latest`** gauge still reflects the full cumulative block fees at the last commit.
> 
> `record_committed_payload` takes separate `incremental_fees_wei` and `cumulative_fees_wei` arguments; metric doc comments were updated to describe this behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eeba33e02f02a35587794fd0fe228fdb0971fc3e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/builder/src/payload_builder_metrics.rs
+++ b/crates/builder/src/payload_builder_metrics.rs
@@ -375,9 +375,10 @@ pub struct PayloadBuildMetrics {
     pub committed_payload_transaction_count: Histogram,
     /// Latest committed payload transaction count.
     pub committed_payload_transaction_count_latest: Gauge,
-    /// Histogram of committed payload fees in wei.
+    /// Histogram of marginal fees in wei added by each committed flashblock. Summing this
+    /// over a window yields true per-block revenue (records the delta, not the running total).
     pub committed_payload_fees_wei: Histogram,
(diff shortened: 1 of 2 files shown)


0xForerunner (changes requested):
`committed_payload_size_bytes` and `committed_payload_size_bytes_latest` should not both exist. Same for the other `latest` metrics here. We should just have `committed_payload_size_bytes` as a histogram. The gauge is useless.

0xForerunner (approved):

PR #741, opened 2026-06-15
chore: reth 2.3.0

0xForerunner:
> [!NOTE]
> **High Risk**
> Large dependency bump across execution, consensus, networking, and payload paths touches core node behavior; local builder/validator compile fixes are small but sit on critical block-production code.
> 
> **Overview**
> Bumps the workspace release to **2.3.0** and refreshes the dependency stack: **reth** git crates move to `7680d6d8…` (reth **2.2.0**), Optimism stack to **`op-reth/v2.3.1`** (including **op-alloy 2.0.0**, **op-revm 20**, **kona**), and **Alloy 2.0.5** / **alloy-primitives 1.6.0**. `Cargo.lock` picks up related transitive updates (e.g. **hickory-resolver 0.26**, **discv5** from a pinned git rev, **imbl** for the transaction pool).
> 
> World Chain’s **payload builder** is adjusted for upstream API changes: sequencer execution passes a new optional argument, `execute_transaction` gas is read via **`tx_gas_used()`**, `BlockBuilderOutcome` ignores **`block_access_list`**, and **`BuiltPayloadExecutedBlock`** now stores plain **`Arc`** hashed state/trie updates instead of `either::Left`. The builder crate drops the unused **`either`** dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9df44e4127aa1bc88cf3b9fe75516325a11478a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Cargo.toml
```diff
@@ -106,121 +106,121 @@ reth-codecs = { version = "0.3.1", default-features = false }
...
-reth-consensus = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
-reth-node-core = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
-reth-tracing = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
-reth-rpc-layer = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
-reth-node-ethereum = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
-reth-storage-api = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
-reth-storage-errors = { git = "https://github.com/paradigmxyz/reth", rev = "27bfddeada3953edc22759080a3659ccea62ca1f", default-features = false }
+reth-eth-wire = { git = "https://github.com/paradigmxyz/reth", rev = "7680d6d8a931c0af4f4eed26e971596970238b54", default-features = false }
```
0xOsiris: Can you change this to `tag = "v2.3.0"`. It's confusing that op-reth is on 2.3.1, but our version is v2.3.0

crates/validator/src/validator.rs
```diff
@@ -145,7 +153,8 @@ impl<Evm: ConfigureEvm + Clone, T: FlashblockTypes<Evm>> FlashblocksBlockValidat
                     )
                 })?;
 
-            self.validate_payload(&executed.into_executed_payload(), diff_clone)
```
0xOsiris, line 148: `into_executed_payload()` was deprecated?

Cargo.toml
0xForerunner: I don't think we can, because v2.3.0 doesn't line up with the deps listed in op-reth 2.3.1.

crates/validator/src/coordinator.rs
```diff
@@ -113,6 +115,13 @@ impl FlashblocksExecutionCoordinator {
             .event_stream(provider, move |event| f(event))
     }
 
+    fn into_executed_payload(
```
0xOsiris: This is a direct copy of the helper fn below

0xOsiris (changes requested):

crates/validator/src/validator.rs
0xForerunner: Yeah seems that way. I didn't see another way to do it.

Cargo.toml
0xOsiris: It should be: https://github.com/ethereum-optimism/optimism/blob/423d93e6374a65f3a83c1ed5e29c5998c972a5da/rust/Cargo.toml#L330

op-reth pins to v2.3.0 tag

crates/validator/src/coordinator.rs
0xForerunner: fixed

crates/builder/src/traits/context.rs
```diff
@@ -289,7 +289,7 @@ impl PayloadBuilderCtx for OpPayloadBuilderCtx<WorldChainEvmConfig, WorldChainSp
             >,
         Txs: PayloadTransactions<Transaction = Self::Transaction>,
     {
-        self.execute_best_transactions(info, builder, best_txs)
+        self.execute_best_transactions(info, builder, best_txs, Some(effective_gas_limit), None)
```
0xOsiris: I'd be down to just remove this implementation entirely. We are not using it anywhere. Just adds code

crates/builder/src/payload_builder.rs
```diff
@@ -610,6 +610,7 @@ where
         block,
         hashed_state,
         trie_updates,
+        block_access_list: _,
```
0xOsiris, line 619: We get an access list  👀

alessandromazza98: reth is starting to support block access list. Therefore I think pretty soon, when we want to take some time for BAL, we can deprecate some of our custom code and direclty use reth's internals whenever possible

crates/validator/src/coordinator.rs
```diff
@@ -408,6 +403,7 @@ where
         parent_hash: base.parent_hash,
         parent_beacon_block_root: Some(base.parent_beacon_block_root),
         extra_data: base.extra_data.clone(),
+        ..Default::default()
```
alessandromazza98: nit: what do you think of make the remaining fields explicit and still use Default::default() for them. This way it's generally better for debugging if something is wrong with default values for fields that we don't see here

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "2.2.0"
+version = "2.3.0"
```
alessandromazza98: what about bumping our own version in a follow up PR instead of the same one of reth bump?

Cargo.toml
alessandromazza98: Eric is right. When `op-reth` published `op-reth/v2.3.1` they were still using this hash commit for reth, not their `2.3.0` version

alessandromazza98 (approved):
nice🔥 I left some non-blocking comments. The main one is related to whether to bump our own version in this same PR or do it in a follow up PR which I would slightly prefer

Cargo.toml
0xOsiris: We shouldn't be pinning reth to a commit far before v2.3.0, I'd rather pin op-reth to a commit ahead of v2.3.1. We are not even on reth v2.3.0 currently

0xOsiris (approved):

PR #759, opened 2026-06-18
feat: live sync witness cache oracle

0xOsiris:
[PROTO-4718
](https://linear.app/worldcoin/issue/PROTO-4718/witness-data-oracle-design)
Creates an optional live witness cache that accumulates pre-image witnesses into a fixed size in memory circular ring buffer. The witnesses can then be fetched over a range externally with enormous compute savings. 

As historical witnesses require unwinding the database to the beginning of the range. Witnesses are deterministic based on execution (and the last proven state), and can be easily cached during live sync in the Block Executor minimizing both compute, and complexity of the pre-image witness collection in the proof system.

---

> [!NOTE]
> **Medium Risk**
> Touches block-import execution via an executor wrapper on the engine validation path; default-off limits blast radius, but enabled mode adds async witness assembly and a new debug RPC surface on proof nodes.
> 
> **Overview**
> Introduces an **optional live pre-image witness oracle** for proof workflows: when `--witness.collect` is set, the node captures witnesses during normal `newPayload` import instead of relying on repeated `debug_executePayload` re-execution.
> 
> **New `world-chain-witness` crate** defines `BlockWitness` / `RangeWitness` and a bounded `WitnessCache` (depth from `--witness.depth`, default 1024).
> 
> **EVM layer:** `WorldChainEvmConfig` is no longer a type alias—it wraps `OpEvmConfig` with `WorldChainBlockExecutorFactory` / `WorldChainBlockExecutor`, which snapshot `ExecutionWitnessRecord` from the live revm `State` on executor `finish` and send `BlockExecutionWitness` over a channel. A background collector assembles full `BlockWitness` values (parent state trie, headers, RLP header/txs) into the cache. With collection off, behavior stays a passthrough (no sender). Requires `#![feature(min_specialization)]` for the `State` DB witness hook.
> 
> **Wiring:** CLI `WitnessArgs`, node context creates channel + cache when enabled, `WorldChainExecutorBuilder` arms the sender, and add-ons spawn the collector and register **`debug_collectRangeWitness`** on the `debug` module when collection is on.
> 
> Adds design doc `docs/proof/live-preimage-witness-oracle.md`. **Kona host pre-seeding** (design Phase 5) is not included in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13239c386b468cf38c6d494d1bfa3784f0bbc898. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
```
alessandromazza98, line 1: why adding this? I'd remove it to be fair, in particular because the nightly changes every day, therefore it's high likely that we don't have this version anymore

alessandromazza98: if we want a toolchain, then I can be down to pin to a stable version of the rust compiler, not a nightly

docs/proof/live-preimage-witness-oracle.md
```diff
@@ -0,0 +1,222 @@
+# Live Pre-image Witness Oracle — Design
+
+Linear: [PROTO-4718 — witness data oracle: design](https://linear.app/worldcoin/issue/PROTO-4718/witness-data-oracle-design)
```
alessandromazza98, line 3: nit: I'd remove this. The linear ticket doesn't include any detail that is not already well documented in this readme

docs/proof/live-preimage-witness-oracle.md
```diff
@@ -0,0 +1,222 @@
...
+- reth (paradigmxyz): tag `v2.3.0` (`9384bc5`).
+- op-reth (ethereum-optimism/optimism fork): rev `423d93e`.
+- `alloy_rpc_types_debug::ExecutionWitness { state, codes, keys, headers }` (all `Vec<Bytes>`),
+  via `alloy-rpc-types-debug` `2.0.5`.
+- `reth_revm::witness::ExecutionWitnessRecord` { `hashed_state`, `codes`, `keys`,
+  `lowest_block_number` }; `record_executed_state` / `from_executed_state`.
+- `StateProofProvider::witness(input, hashed_state, mode)` (`storage-api/src/trie.rs`) produces the
+  trie-node `state` set by walking pre-state cursors for the touched hashed keys (no execution).
+
+## Implementation plan
```
alessandromazza98, line 126: you can remove the implementation plan now. I'd leave this document as a readme to explain why we build this live preimage oracle and its architecture. The impl plan is useless now that the code is already finalized

crates/evm/src/execution/executor.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/// Records an [`ExecutionWitnessRecord`] from a database when it is a [`State`] cache.
+trait MaybeWitness {
+    /// Returns the recorded witness if `self` is backed by a [`State`] cache.
+    fn witness(&self) -> Option<ExecutionWitnessRecord>;
+}
+
+impl<T> MaybeWitness for T {
+    default fn witness(&self) -> Option<ExecutionWitnessRecord> {
```
alessandromazza98, line 28: instead of this impl box, you can write the default impl inside the trait itself

alessandromazza98 (approved):
generally lgtm. I left some nits that you can address before merging

crates/evm/src/execution/executor.rs
0xOsiris: > instead of this impl box, you can write the default impl inside the trait itself

Yeah this is very weird tbh aha

rust-toolchain.toml
0xOsiris: I have not been able to find a way to implement this without min_specialization. We need a monomorphization of `MaybeWitness` over a generic `StateDB` because of Reth's trait bounds over the `BlockExecutor`, `BlockBuilder`, and `BlockExecutorFactory`. 

There is no way to create a blanket implementation of `MaybeWitness` over `T: StateDB` with a granular implementation over `T: StateDB = State<DB>`. 

There is also no way to easily capture the witness outside of the executors internals as we need to capture this from `engine_newPayload` execution.

PR #768, opened 2026-06-19
refactor: new proof system design

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4728/feat-make-prover-service-use-a-persistent-db


proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,906 @@
...
+        let Some(row) = row else {
+            return Err(classify_backend_update(&mut tx, backend_job_id).await);
+        };
+        let proof_id = proof_id_from_bytes(row.try_get("proof_id").map_err(queue_db)?)
+            .map_err(ProofJobQueueError::Internal)?;
+        mark_proof_failed(&mut tx, proof_id, reason)
+            .await
+            .map_err(queue_db)?;
+        tx.commit().await.map_err(queue_db)?;
+        Ok(())
```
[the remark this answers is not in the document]
alessandromazza98: fixed

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,962 @@
...
+                ),
+            );
+
+            let lease_token = LeaseToken::new();
+            sqlx::query(
+                "update proof_backend_jobs
+                 set locked_until = now() + ($2::bigint * interval '1 millisecond'),
+                     lease_token = $3,
+                     advance_attempts = advance_attempts + 1,
+                     updated_at = now()
```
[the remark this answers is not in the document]
alessandromazza98: fixed thanks

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,973 @@
...
+        let queued: i64 = sqlx::query_scalar(
+            "select count(*) from proof_jobs where backend = $1 and status = 'queued'",
+        )
+        .bind(backend.as_str())
+        .fetch_one(&mut *tx)
+        .await
+        .map_err(request_db)?;
+        if queued as usize >= self.config.max_queue_len {
+            return Err(ProofRequestError::QueueFull(backend));
+        }
```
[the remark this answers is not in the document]
alessandromazza98: This is true but for now I'd keep this without adding a heavier solution. It's not a critical problem, we're just adding more jobs to the queue. Eventually they will all get served

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,973 @@
...
+        )
+        .bind(&proof_id)
+        .bind(backend.as_str())
+        .bind(proof_request.game.as_slice())
+        .bind(proof_request.root_claim.as_slice())
+        .bind(l2_to_i64(proof_request.l2_block_number)?)
+        .bind(proof_request.l1_head.as_slice())
+        .execute(&mut *tx)
+        .await
+        .map_err(request_db)?;
```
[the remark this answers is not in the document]
alessandromazza98: this is also acceptable, there is no big problem

proofs/worker/src/worker.rs
```diff
@@ -266,55 +299,212 @@ fn spawn_job<B>(
...
+        }
+        BackendUpdate::Complete(proof) => {
+            queue
+                .submit_proof(
+                    ProofResponse { id, proof },
+                    ProofSubmissionLease::ProofJob { lease_token },
+                )
+                .await
+        }
+        BackendUpdate::Failed(reason) => queue.fail_proof(id, reason, lease_token).await,
```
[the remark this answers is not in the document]
alessandromazza98: this is fine, eventually I'll change this in next iterations if really needed. I think testing is critical now before making other changes

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,1043 @@
...
+                status,
+            }),
+        }
+    }
+
+    pub(crate) async fn get_next_proof(
+        &self,
+        backend: ProofBackend,
+    ) -> Result<Option<LeasedProofRequest>, ProofJobQueueError> {
+        let mut tx = self.begin_queue_tx().await?;
```
piohei, line 193: Are you sure we want to have transaction created before the loop instead of each loop cycle?

piohei: Also do we want to be explicit about tx type here?

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,1043 @@
...
+                warn!(%proof_id, attempts, "proof job failed: start attempts exhausted");
+                continue;
+            }
+
+            let request = request_from_row(&row).map_err(ProofJobQueueError::Internal)?;
+            let lease_token = LeaseToken::new();
+            sqlx::query(
+                "update proof_jobs
+                 set status = 'starting',
+                     locked_until = now() + ($2::bigint * interval '1 millisecond'),
```
piohei, line 233: What about creating `now` on rust side and the pass it using bind? This way we can easily have retries with original time.

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,1043 @@
...
+        let mut tx = self.begin_queue_tx().await?;
+        let backend: Option<String> = sqlx::query_scalar(
+            "update proof_jobs
+             set status = 'backend_pending',
+                 locked_until = null,
+                 lease_token = null,
+                 updated_at = now()
+             where proof_id = $1
+               and lease_token = $2
+               and status = 'starting'
```
piohei, line 270: What about using bind with status enum translated to string in rust instead of hardcoding into queries?

proofs/prover-service/src/store.rs
```diff
@@ -0,0 +1,1043 @@
...
+    }
+
+    async fn try_submit_proof_from_backend_job(
+        &self,
+        proof: ProofResponse,
+        backend_job_id: i64,
+        lease_token: LeaseToken,
+    ) -> Result<(), ProofJobQueueError> {
+        let mut tx = self.begin_queue_tx().await?;
+        let row = sqlx::query(
```
piohei, line 744: I would put all queries into method like `complete_proof` and use them. You could even make one more abstraction here keeping methods to call queries on database in its own module/type and then have one service to compound them and put more logic there.

proofs/prover-service/src/types.rs
```diff
@@ -71,8 +95,10 @@ impl std::fmt::Display for ProofRequestId {
 pub enum ProofStatus {
     /// Waiting in the backend queue for a worker.
     Queued,
-    /// Leased to a worker that is generating the proof.
-    InProgress,
+    /// Leased to a worker that is starting backend work.
+    Starting,
```
piohei, line 99: To be honest I don't get why we split `InProgress` status into `Starting` and `BackendPending`. I would keep the proof simple and then let us check the "child jobs" to see where we are on processing.

proofs/prover-service/src/types.rs
```diff
@@ -132,3 +188,214 @@ pub struct ProofResponse {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Durable backend proof phase.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum BackendProofPhase {
+    /// SP1 compressed range proof phase.
+    Range,
```
piohei, line 229: Does it mean that we moved responsibility to execute all steps to get the proof from SP1 worker to prover service?

proofs/prover-service/src/types.rs
alessandromazza98: They differ in how the re-try is handled. If a row is `BackendPending`, then it will never be retried (until it remains `BackendPending`. If it's `Started`, it can be retried when the `locked_until` expires. If we only use `InProgress` we should create something else to make this difference explicit.

proofs/prover-service/src/types.rs
alessandromazza98: no, the responsibility to advance the steps is on the proof worker side. The prover-service is just the middle layer between defender and proof workers. This type is here but I'll probably move it to a more general primitive crate in a follow up

proofs/prover-service/src/store.rs
alessandromazza98: sure, what's the difference between doing it on rust and sql? does it change a lot?

alessandromazza98: I mean, we don't care about milliseconds here

proofs/prover-service/src/store.rs
alessandromazza98: - I don't think we need to explicit the type of the tx here, it's not fundamental
- the difference is not big because most of the times the loop is executed once. But I don't have a strong opinion here, I can put the tx creation inside the loop

proofs/prover-service/src/store.rs
alessandromazza98: my idea is to have the `service.rs` file that contains the actual logic. Then the `store.rs` file is the `sql backend`. Do you have in mind to create another layer on top? such as `service.rs` -> `store.rs` -> `sql.rs` ? what would you put in `store` and `sql`?

proofs/prover-service/src/store.rs
alessandromazza98: the rust enum already exists for the statuses. so you mean something like:

```rust
let status = ProofStatus::Starting;
let backend: Option<String> = sqlx::query_scalar(
            "update proof_jobs
             set status = 'backend_pending',
                 locked_until = null,
                 lease_token = null,
                 updated_at = now()
             where proof_id = $1
               and lease_token = $2
               and status =$#3
             returning backend",
        )
        .bind(proof_id_bytes(proof_id))
        .bind(lease_token.0)
        .bind(status)
        .fetch_optional(&mut *tx)
        .await
        .map_err(queue_db)?;
```
?

proofs/prover-service/src/types.rs
alessandromazza98: further, this difference makes it possible to know in which table to look at for the lease token (to match)

piohei (approved):
Discussed with plan for next steps in refactor here. :)

alessandromazza98:
These are the next steps, discussed with Piotr:

- make the tx explicit for the postgres db

- now() in Rust instead of sql - and in general not hard code values (for example proof status are now hard coded -> I can use the rust enum that already exists)

- right now prover-service holds 2 tables: proof_jobs and proof_backend_jobs. What can be done is that the proof_backend_jobs table is not held by the prover-service but it’s a table that all proof workers use and maintain. 
        - worker id

PR #773, opened 2026-06-23
feat: use reth bal construction/execution for flashblocks

0xForerunner:
> [!NOTE]
> **High Risk**
> Large refactor of flashblock execution, BAL hashing/index semantics, and bundle composition on the critical block-building and validation path.
> 
> **Overview**
> Removes the **`world-chain-state`** crate and routes flashblock **Block Access List (BAL)** construction and validation through **upstream `revm_database::State`** (`with_bal_builder`, `take_built_alloy_bal`) and **`revm::state::bal::Bal`** on the validator side.
> 
> **Payload building** no longer wraps execution in `BalBuilderDb`; BAL-enabled builds use cached prestate from prior committed bundles (`cache_prestate_from_bundle`), merge bundles via `extend_flashblock_bundle`, and finish by emitting a **`FlashblockAccessList`** built from the alloy BAL. **Non-BAL** flashblock paths use the same prestate/extend pattern. Executor bounds now use **`reth_evm::block::StateDB`** instead of the removed local `StateDB` trait.
> 
> **`FlashblockAccessList`** in primitives is slimmed to a sidecar around **`alloy_eip7928::BlockAccessList`** ( **`u64`** tx index bounds, `from_block_access_list`); custom merge/flush/`extend_bundle` logic is dropped. **`record_op_l1_block_bal_reads`** records OP L1-block predeploy storage reads so upstream BAL construction matches Jovian fee accounting.
> 
> **`alloy-eip7928`** is bumped **0.3.7 → 0.4.3**. Built payload fees are taken from the current build only (`info.total_fees`), without adding prior committed fees again.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f7191d5a39a9b15bf1a8b5930a166c2b6b90062a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/builder/src/payload_builder.rs
```diff
@@ -637,7 +635,7 @@ where
     let payload = OpBuiltPayload::new(
         ctx.payload_id(),
         sealed_block,
-        info.total_fees + committed_state.fees,
+        info.total_fees,
```
alessandromazza98, line 638: why not adding the committed_state fees here?

alessandromazza98: oh I see, we already sum the committed state fees above 🙏

alessandromazza98 (approved):
sweet 🔥

pkg/contracts/scripts/devnet/DeployProofSystem.s.sol
0xOsiris, line 1: Why does this need any changes?

0xOsiris: Let's not touch anything with the proof system in this PR

crates/validator/src/execution_strategy.rs
```diff
@@ -23,30 +25,30 @@ use revm::database::{BundleAccount, states::bundle_state::BundleRetention};
...
-    database::{
-        bal_builder_db::{BalBuilderDb, NoOpCommitDB},
-        bundle_db::BundleDb,
-        temporal_db::TemporalDbFactory,
-    },
-};
+
+struct BalWorkerOutput<R> {
+    index: u64,
+    transaction: reth_primitives_traits::Recovered<OpTransactionSigned>,
```
0xOsiris:
```suggestion
    transaction: Recovered<OpTransactionSigned>,
```

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+            }
+            .boxed()
+            .into());
+        }
 
-        let mut noop_state = NoOpCommitDB::new(temporal_db);
-        let mut database = BalBuilderDb::new(&mut noop_state);
-        database.set_index(block_access_index);
+        let received_bal = Arc::new(
+            revm::state::bal::Bal::try_from(access_list.as_block_access_list())
```
0xOsiris:
```suggestion
            Bal::try_from(access_list.as_block_access_list())
```

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+            .with_database(StateProviderDatabase::new(finish_state_provider.as_ref()))
+            .with_cached_prestate(cache_prestate_from_bundle(&committed_state.bundle))
+            .with_bundle_update()
+            .with_bal_builder()
+            .build();
+        db.set_bal_index(BlockAccessIndex::new(min_tx_index));
 
-        let mut executor: OpBlockExecutor<_, OpRethReceiptBuilder, Arc<WorldChainSpec>> =
+        let evm = OpEvmFactory::<OpTx>::default().create_evm(&mut db, ctx.evm_env.clone());
+        let mut canonical_executor: OpBlockExecutor<_, OpRethReceiptBuilder, Arc<WorldChainSpec>> =
```
0xOsiris: Just abstract this under the `OpBlockExecutorFactory`

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
             );
 
-        executor.gas_used = committed_state.gas_used;
-        executor.da_footprint_used = committed_state.blob_gas_used;
-        executor.receipts = committed_state.receipts_iter().cloned().collect();
+        canonical_executor.gas_used = committed_state.gas_used;
+        canonical_executor.da_footprint_used = committed_state.blob_gas_used;
+        canonical_executor.receipts = committed_state.receipts_iter().cloned().collect();
 
-        let (validator, access_list_receiver) = BalBlockValidator::new(
```
0xOsiris, line 161: I'd still rather have a validator type that implements our `execute_block` interface for testability of BAL validation. Rather than packing all the complicated logic into this function.

crates/evm/src/execution/bal.rs
```diff
@@ -34,19 +42,42 @@ use revm::database::states::bundle_state::BundleRetention;
...
+    L1_OVERHEAD_SLOT,
+    L1_SCALAR_SLOT,
+    ECOTONE_L1_BLOB_BASE_FEE_SLOT,
+    ECOTONE_L1_FEE_SCALARS_SLOT,
+    DA_FOOTPRINT_GAS_SCALAR_SLOT,
+];
+
+/// Records the OP L1 block predeploy slots that OP fee logic reads outside the
+/// transaction result state.
+pub fn record_l1_block_bal_reads<DB>(db: &mut State<DB>) -> Result<(), BlockExecutionError> {
```
0xOsiris: What is the purpose of this, and why is it necessary? What is it even accomplishing?

0xOsiris: These slots should be touched by the EVM organically in the execution of the system transaction writing the L1 block predeploy (which should load the slots onto the state). I don't think we should be doing this manually. This seems incorrect to me

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+            PayloadBuildStage::SequencerTxExecution,
+            txs_execution_started.elapsed(),
+        );
 
-        let computed_access_list = access_list_receiver
-            .recv()
-            .map_err(BalExecutorError::other)?;
+        let basefee = ctx.evm_env.block_env.basefee;
+        let mut fees = U256::ZERO;
+        for output in worker_outputs {
```
0xOsiris, line 342: Just make this a separate helper function on `BalBlockValidator`

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+        let mut canonical_transactions: Vec<_> =
+            committed_state.transactions_iter().cloned().collect();
+
+        if committed_state.is_first {
+            let pre_execution_changes_started = Instant::now();
+            canonical_executor
+                .evm_mut()
+                .db_mut()
+                .set_bal_index(BlockAccessIndex::new(0));
+            canonical_executor.apply_pre_execution_changes()?;
```
0xOsiris: Here we touch the L1 Block Predeploy slots

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+            canonical_executor
```
0xOsiris: This is overwritten on the first output commitment below. It's a no op

crates/validator/src/execution_strategy.rs
```diff
@@ -112,102 +114,194 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+                    .expect("fee is always valid; execution succeeded");
+                fees += U256::from(gas_output.tx_gas_used()) * U256::from(miner_fee);
+            }
+
+            canonical_transactions.push(output.transaction);
+        }
 
-        let computed_access_list_hash =
-            world_chain_primitives::access_list::access_list_hash(&computed_access_list);
+        canonical_executor
```
0xOsiris, line 363: Shouldn't this be inclusive of the last

```rust
canonical_executor
                .evm_mut()
                .db_mut()
                .set_bal_index(BlockAccessIndex::new(output.index));
```

Are we considering the Fee Increments as a transaction?

crates/validator/src/execution_strategy.rs
```diff
@@ -216,66 +310,109 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
             .into());
         }
 
-        if outcome.block.receipts_root != diff.receipts_root {
+        let merge_started = Instant::now();
+        db.merge_transitions(BundleRetention::Reverts);
```
0xOsiris, line 403: The reverts need to be flattened onto a single array index here

0xOsiris (changes requested):
Left some comments, overall looks good!

crates/validator/src/execution_strategy.rs
0xOsiris: See: https://github.com/worldcoin/world-chain/blob/1150a36086856dbad84a61991a3ecfb2532a9d44/crates/evm/src/execution/basic.rs#L178

We need to follow this pattern identically

pkg/contracts/scripts/devnet/DeployProofSystem.s.sol
0xForerunner: Because the devnet was broken. It wouldn't deploy.

alessandromazza98: it should be fixed already in another PR, try to rebase

alessandromazza98: not sure if that's the same issue, but I modified it in a PR that has been merged already

crates/evm/src/execution/bal.rs
0xForerunner: Yeah, I tried removing it, but BAL validation breaks.

The L1 deposit updates the predeploy normally, but OP/Jovian fee accounting also reads the L1 block predeploy directly from the DB before each non-deposit tx. That read is outside the tx result state, so upstream BAL building does not record it on commit.

Without these read-only entries, workers fail with `AccountNotFound` for `0x420...0015`.

crates/validator/src/execution_strategy.rs
0xForerunner: I don’t think it should be the last tx index. `max_tx_index` is the post-last access index.

Each tx is committed at `output.index`. After the last tx, the builder has already bumped the BAL index once more, and `finish()` runs post-block balance increments at that final index.

crates/validator/src/execution_strategy.rs
```diff
@@ -216,66 +399,115 @@ impl<S: StateRootStrategy> ExecutionStrategy<WorldChainEvmConfig, S>
...
+        self.ctx
+            .attempt_metrics
+            .record_stage_duration(PayloadBuildStage::MergeTransitions, merge_started.elapsed());
+
+        let flattened = flatten_reverts(&db.bundle_state.reverts);
+        db.bundle_state.reverts = flattened;
+        let bundle = extend_flashblock_bundle(&committed_state.bundle, db.take_bundle());
+
+        let state_root_started = Instant::now();
+        let state_root_handle =
```
0xOsiris, line 413: This should happen at the start of `execute_block` via conversion of `access_list` to `BundleState`. We shouldn't block on parallel execution to compute the state root. Then we should initialize the state root handle in the background to walk the trie. After parallel execution we should block the handle for the result. We might spawn a dedicated OS thread for this at the very start

0xOsiris (approved):

PR #787, opened 2026-06-25
fix: pre-flight tx validation

0xOsiris:
> [!NOTE]
> **Medium Risk**
> Touches core payload building and flashblock validation gas accounting; incorrect refund math could reject valid txs or over-include, but scope is focused and covered by a unit test for `pre_refund_gas_used`.
> 
> **Overview**
> Fixes **pre-flight transaction validation** when continuing a flashblock build on top of a committed payload: block header **gas used** is canonical (after SDM/post-exec refunds), but the executor enforces limits against **pre-refund** gas.
> 
> Adds **`evm_gas_used`** on **`CommittedState`**, reconstructed via **`pre_refund_gas_used`** from post-exec **`gas_refund_entries`**. **`effective_gas_limit`** now subtracts committed **`evm_gas_used`** instead of **`gas_used`**. Pool execution records pre-refund gas through **`execute_transaction_with_result_closure`** and **`PreRefundGasUsed`**, and continuation builds reset **`cumulative_evm_gas_used`** to zero while seeding block executors with committed **`evm_gas_used`**. Builder, validator, and test fixtures are updated consistently.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 162da9ffe05f81642cdbee10f7b58a23620872c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/builder/src/payload_builder.rs
+++ b/crates/builder/src/payload_builder.rs
@@ -369,13 +369,14 @@ where
 
     let effective_gas_limit = ctx
         .effective_gas_limit()
-        .saturating_sub(committed_state.gas_used);
+        .saturating_sub(committed_state.evm_gas_used);
 
     trace!(
         target: "flashblocks::payload_builder",
         gas_limit = attributes.gas_limit,
         effective_gas_limit,
         committed_gas_used = committed_state.gas_used,
+        committed_evm_gas_used = committed_state.evm_gas_used,
         timestamp = ctx.attributes().timestamp(),
         "building new payload"
     );
@@ -539,7 +540,8 @@ where
         committed_payload.map_or(ExecutionInfo::default(), |p| ExecutionInfo {
             total_fees: p.fees(),
             cumulative_gas_used: p.block().gas_used(),
-            // The amount of EVM gas used gas_used() + gas_refunded() for an individual flashblock.
+            // `effective_gas_limit` already excludes committed pre-refund gas, so this tracks only
+            // new transactions executed in the continuation build.
             cumulative_evm_gas_used: 0,
             cumulative_da_bytes_used: committed_state
                 .transactions_iter()
@@ -694,6 +696,7 @@ where
         R::default(),
     );
     executor.gas_used = committed_state.gas_used;
+    executor.evm_gas_used = committed_state.evm_gas_used;
     executor.da_footprint_used = committed_state.blob_gas_used;
     executor.receipts = committed_state.receipts_iter().cloned().collect();
 
@@ -744,6 +747,7 @@ where
     );
 
     executor.gas_used = committed_state.gas_used;
+    executor.evm_gas_used = committed_state.evm_gas_used;
     executor.da_footprint_used = committed_state.blob_gas_used;
     executor.receipts = committed_state.receipts_iter().cloned().collect();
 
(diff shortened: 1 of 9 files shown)


crates/builder/src/payload_builder.rs
```diff
@@ -538,7 +539,8 @@ where
         committed_payload.map_or(ExecutionInfo::default(), |p| ExecutionInfo {
             total_fees: p.fees(),
             cumulative_gas_used: p.block().gas_used(),
-            cumulative_evm_gas_used: p.block().gas_used(),
+            // The amount of EVM gas used gas_used() + gas_refunded() for an individual flashblock.
+            cumulative_evm_gas_used: 0,
```
0xForerunner, line 545: Should we not use the committed pre-refund gas instead of resetting to zero? Since effective_gas_limit is only reduced by committed canonical gas_used, a prior refunded flashblock can leave too much apparent room later.

0xForerunner (changes requested):

crates/builder/src/payload_builder.rs
0xOsiris: We are checking if `tx.gaslimit() + flashblock.cumulative_gas_used() > effective_gas_limit (gas left)` to check if a transaction is valid.

For any flashblock if this holds the transaction is valid

`flashblock.cumulative_gas_used()` needs to account for both refunded, and non refunded gas

alessandromazza98: where is this `cumulative_evm_gas_used` used?

alessandromazza98: further, I also don't understand why we don't need to use the committed pre-refund gas instead of resetting to zero

0xOsiris: > further, I also don't understand why we don't need to use the committed pre-refund gas instead of resetting to zero

I don't understand the question

alessandromazza98: Why set it to 0 and not use the precommitted value?

0xOsiris: > Why set it to 0 and not use the precommitted value?

https://github.com/worldcoin/world-chain/pull/787#discussion_r3471789079

0xOsiris: > Why set it to 0 and not use the precommitted value?

Because we are cross referencing gas left. Not the total gas limit. Effective gas is subtracted from cumulative gas used

0xForerunner: Yeah, agreed that the cumulative value needs to count refunded + non-refunded gas. Right now `effective_gas_limit` is reduced by the committed block’s canonical `gas_used`. If that committed flashblock had refunds, that leaves too much “gas left” compared to pre-refund EVM gas. So resetting `cumulative_evm_gas_used` to 0 seems safe only if the remaining limit was also reduced by committed pre-refund gas. Otherwise the next flashblock can still over-admit by the refunded amount

0xOsiris: Yeah that's right

0xForerunner (approved):

0xOsiris (commented):
LGTM

PR #794, opened 2026-06-29
refactor: new api for prover-service

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4750/refactor-prover-service-logic

This PR completes the refactor of the `prover-service` api. In particular the api used by `world-chain-defender` doesn't change. The api used by the `proof-worker` changes a lot, therefore the proof worker changes a lot as well to adapt.

The architecture is easier to reason about now:

- a proof worker is generic over a backend
- it claims proof requests from the `prover-service`
- then it calls the `backend.handle_claimed_job` method that each proof backend must implement
- this call on the backend `handle_claimed_job` method is actually spawned on a join set so that it can work async from the main loop of the proof worker
- there is a semaphore permit (that defaults to 1) though that limits the amount of jobs that a proof worker can spawn
- the `handle_claimed_job` method is then implemented by nitro and sp1 and it does everything needed to generate the final proof
- the proof is then returned back to the proof worker who submits it back to the `prover-service` and finally releases the permit

The postgres db has two tables:
- `proof_requests`: the main table that holds all necessary information for a proof request (and eventually the final proof once it's submitted)
- `proof_sessions`: this is only used by the sp1 worker (currently not used, it will be done in a next step). It's used to save sp1 sessions so that if a proof worker fails in the middle of a sp1 proof generation, a new proof worker can resume the work instead of starting again from scratch.

### Next Steps

The current PR doesn't implement everything, it's already pretty big. Here are next steps, some of them are cleanup / chores, others are way more important:

- add the heartbeat system. Right now there is nothing that extend the `lock_expires_at`, therefore it's possible that if a prover takes too much time to generate the final proof, the job becomes stale. The heartbeat should periodically extend the `lock_expires_at`, to signal that the proof worker is still working on that job and there is no need to make it stale.
- add a status poller. Currently there is nothing that make a proof request fail after it has reached the `max_attempts` number of attempts. Therefore the `world-chain-defender` would indefinitely wait for this proof to come even though this is not possible anymore. We need a service (attached to the `prover-service`) that periodically checks for this kinds of proof requests and mark them as failed, so that `world-chain-defender` is aware of this and can eventually tries to submit a new proof request
- in the `store.rs` file there are some TODOs that must be fulfilled. This is gonna be a small and fast PR.
- currently the sp1 backend is not using the `proof_sessions` table at all and it's sync. Transform it in async and use the `proof_sessions` to store and retrive sessions.
- add retry functionality to the `submit_proof` method that the proof worker calls to submit the proof back to the `prover-service`


proofs/prover-service/src/store.rs
```diff
@@ -78,113 +75,135 @@ impl ProverServiceStore {
...
+                    retry_count = retry_count + 1,
+                    failure_reason = NULL,
+                    proof_data = NULL,
+                    completed_at = NULL,
+                    worker_id = NULL,
+                    lock_id = NULL,
+                    lock_expires_at = NULL,
+                    attempt = 0
+                WHERE id = $3
+                "#,
```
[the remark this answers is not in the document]
alessandromazza98: fixed

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +233,370 @@ impl ProverServiceStore {
...
+                job_status = $4,
+                attempt = attempt + 1,
+                lock_expires_at = $5,
+                updated_at = $6
+            WHERE proof_id = (
+                SELECT proof_id FROM proof_requests
+                WHERE backend = $7
+                    AND (
+                        job_status = $8
+                        OR (job_status = $9 AND lock_expires_at < $10 AND attempt < $11)
```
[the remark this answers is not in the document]
alessandromazza98: this is true but it's intentional. A next step is to build a `StatusPoller` that handles exactly this specific case and clears them

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +233,370 @@ impl ProverServiceStore {
...
+            let status_str: &str = row.get("status");
+            let status = SessionStatus::try_from(status_str).map_err(|err| {
+                ProofJobQueueError::Internal(format!(
+                    "invalid session status '{status_str}': {err}"
+                ))
+            })?;
+            if status.is_terminal() {
+                // return this proof session
+                // TODO: do it
+            }
```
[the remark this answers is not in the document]
alessandromazza98: fixed

proofs/prover-service/src/store.rs
```diff
@@ -78,113 +75,135 @@ impl ProverServiceStore {
...
         .map_err(request_db)?;
 
-        tx.commit().await.map_err(request_db)?;
-        info!(%id, %backend, "proof request queued");
-        Ok(id)
+        if insert_result.rows_affected() > 0 {
+            // no conflict
+            tx.commit().await.map_err(request_db)?;
+            return Ok(id);
+        }
```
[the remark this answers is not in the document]
alessandromazza98: with the new db approach, this max is not longer needed. I'll remove this in the next commit, fixed

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +233,370 @@ impl ProverServiceStore {
...
             )
-            .bind(backend_job_id)
-            .bind(lock_id.0)
-            .bind(next_poll_at)
-            .bind(attempts)
+            .bind(proof_id_bytes(proof_id))
+            .bind(session_type.as_str())
+            .bind(backend_session_id)
+            .bind(status.as_str())
+            .bind("failure reason".to_string()) // TODO: replace this with an input value
```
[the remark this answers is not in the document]
alessandromazza98: fixed

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +232,369 @@ impl ProverServiceStore {
...
+        for row in existing_backend_sessions {
+            let status_str: &str = row.get("status");
+            let status = BackendSessionStatus::try_from(status_str).map_err(|err| {
+                ProofJobQueueError::Internal(format!(
+                    "invalid session status '{status_str}': {err}"
+                ))
+            })?;
+            if status.is_terminal() {
+                // return this proof session
+                // TODO: do it
```
alessandromazza98, line 414: i'm gonna do this in a next step

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +232,369 @@ impl ProverServiceStore {
...
+                    status = $2,
+                    failure_reason = $3
+                WHERE id = $4
+                AND (status = $5 OR status = $6)
+                RETURNING id
+                "#,
+            )
+            .bind(backend_session_id)
+            .bind(status.as_str())
+            .bind("failure reason".to_string()) // TODO: replace this with an input value
```
alessandromazza98, line 454: same here

proofs/sp1-worker/src/backend.rs
```diff
@@ -60,95 +54,44 @@ where
...
-                self.config.allow_unfinalized,
-                self.config.split_count,
-                self.config.prover_address,
-            ),
-        )?;
+        // Witness generation and proving are CPU/IO heavy and fully synchronous; run them on
+        // the blocking pool so the worker's async scheduler is not starved. Range proofs
+        // parallelize on their own scoped threads inside `prove_validity`.
+        //
+        // TODO: drive the SP1 proving network asynchronously, checkpointing range (STARK) and
```
alessandromazza98, line 65: I'm gonna do this in a next step

proofs/prover-service/src/store.rs
```diff
@@ -78,113 +74,134 @@ impl ProverServiceStore {
...
+            ProofRequestError::Internal(format!("Invalid proof status in db: {}", e))
+        })?;
+
+        if proof_status == ProofStatus::Failed {
+            // retry the entire proof job if retry_count is less than the max_retry
+            let retry_count: i32 = row.get("retry_count");
+            if retry_count > self.config.max_retries as i32 {
+                tx.rollback().await.map_err(request_db)?;
+                return Err(ProofRequestError::TooManyRetries(id));
+            }
```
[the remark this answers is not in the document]
alessandromazza98: this is ok and tolerated

proofs/nitro/worker/src/lib.rs
```diff
@@ -67,12 +65,15 @@ impl NitroBackend {
...
+impl ClaimedProofJobHandler for NitroBackend {
     fn lane(&self) -> ProofBackend {
         ProofBackend::Nitro
     }
 
-    fn start(&self, request: &ProofRequest) -> anyhow::Result<BackendUpdate> {
+    async fn handle_claimed_job(&self, job: ProofJob) -> anyhow::Result<ProofData> {
+        let request = &job.request;
+
         let start_block = request
```
piohei, line 77: What do you think about sending start and end block number in the request instead of getting it from config? I thought that it should match with the proposer output and we could then take it from that and pass with the request.

proofs/prover-service/migrations/20260619_proof_requests.sql
```diff
@@ -1,71 +1,59 @@
...
-    advance_attempts    integer not null default 0,
-    next_poll_at        timestamptz not null,
-    lock_expires_at     timestamptz null,
-    lock_id             uuid null,
+    session_type        TEXT NOT NULL,
+    backend_session_id  TEXT NOT NULL,
 
-    artifact            bytea null,
-    failure_reason      text null,
+    status              TEXT NOT NULL,
```
piohei, line 46: What do you think about checking status field value also like we do in proof_requests table?

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +231,369 @@ impl ProverServiceStore {
...
         let row = sqlx::query(
-            "select start_attempts, backend
-             from proof_requests
-             where proof_id = $1
-               and lock_id = $2
-               and status = $3
-               and worker_id = $4
-             for update",
+            r#"
+            SELECT ps.backend_session_id, ps.status
```
piohei, line 291: Could this query return multiple rows? Or always single one?

proofs/prover-service/src/store.rs
```diff
@@ -214,677 +231,369 @@ impl ProverServiceStore {
...
-        if result.rows_affected() == 0 {
-            return Err(classify_backend_update(&mut tx, backend_job_id).await);
-        }
-        tx.commit().await.map_err(queue_db)?;
-        Ok(())
+        Ok(session)
     }
 
-    pub(crate) async fn advance_backend_job(
+    pub(crate) async fn record_proof_session(
```
piohei, line 325: It is quite long function. What do you think about dividing it into multiple ones?

proofs/prover-service/src/types.rs
```diff
@@ -90,58 +90,92 @@ impl std::fmt::Display for ProofRequestId {
...
-    /// Waiting in the backend queue for a worker.
-    Queued,
-    /// Leased to a worker that is starting backend work.
-    Starting,
-    /// External backend work has been requested and is being polled.
-    BackendPending,
-    /// Proof generated and available via `get_proof`.
-    Completed,
-    /// Permanently failed after exhausting all attempts.
+    /// Proof request has been created but not yet queued.
```
piohei, line 96: What does "not yet queued" meaning?

proofs/nitro/worker/src/lib.rs
alessandromazza98: in general yes this should change in the future but it's not related to this PR, so I'll address it in a different PR

proofs/prover-service/migrations/20260619_proof_requests.sql
alessandromazza98: I agree. This should part of follow up PRs

proofs/prover-service/src/store.rs
alessandromazza98: always one

alessandromazza98: at max one actually

proofs/prover-service/src/store.rs
alessandromazza98: sure I can do it, I'd rather do it in follow up PRs though

proofs/prover-service/src/types.rs
alessandromazza98: `world-chain-defender` has created the proof request and this is saved in the `prover-service` db but no proof worker has started working on it yet. So not queued here means not queued in a proof worker job yet

proofs/prover-service/src/store.rs
alessandromazza98: in fact there is this:

```
CREATE UNIQUE INDEX proof_sessions_active_unique_idx
    ON proof_sessions (proof_id, session_type)
    WHERE status IN ('SUBMITTING', 'RUNNING');
```

piohei (approved):

PR #808, opened 2026-07-02
fix: make record-proof-session idempotent

alessandromazza98:
Part of https://linear.app/worldcoin/issue/PROTO-4858/in-the-storers-file-there-are-some-todos-that-must-be-completed

This PR makes the `record_proof_session` fn idempotent, such that if the backend session status is already in a terminal state (completed or failed), if you try to record the same status, then it becomes a no-op. If the status is different than the finalized one, then return an error.

---

> [!NOTE]
> **Medium Risk**
> Changes worker write semantics and concurrency around proof session rows; incorrect early-return transaction handling could matter, but behavior is narrowly scoped to terminal session conflicts.
> 
> **Overview**
> **`record_proof_session`** now treats terminal backend session rows as immutable before the usual active-session update/insert path.
> 
> When a matching `proof_sessions` row (same proof, session type, and backend session id) is already terminal, re-recording the **same** status returns success as a no-op for worker retries. Recording a **different** status returns a new **`BackendSessionAlreadyTerminal`** error with structured details (stored vs attempted status).
> 
> That error is wired through **`ProofJobQueueError`**, JSON-RPC code **`-32020`**, and the HTTP client’s error mapping. The session lookup query also adds **`ORDER BY id`** for deterministic row locking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6305ed09906a2ca935135539c12e39ba9d38610. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proofs/prover-service/src/store.rs
+++ b/proofs/prover-service/src/store.rs
@@ -2,8 +2,9 @@ use crate::{
     ProofData,
     config::ProverServiceConfig,
     error::{
-        BackendMismatchErrorData, InvalidConfigError, ProofJobQueueError, ProofJobStatusErrorData,
-        ProofMismatchErrorData, ProofRequestError, ProverServiceInitError, TooManyRetriesErrorData,
+        BackendMismatchErrorData, BackendSessionAlreadyTerminalErrorData, InvalidConfigError,
+        ProofJobQueueError, ProofJobStatusErrorData, ProofMismatchErrorData, ProofRequestError,
+        ProverServiceInitError, TooManyRetriesErrorData,
     },
     types::{
         BackendSession, BackendSessionStatus, FailedProofResponse, LockId, LockedProofRequest,
@@ -384,6 +385,7 @@ impl ProverServiceStore {
                 WHERE proof_id = $1
                   AND session_type = $2
                   AND backend_session_id = $3
+                ORDER BY id
                 FOR UPDATE
             "#,
         )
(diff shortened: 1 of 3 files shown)


proofs/prover-service/src/store.rs
```diff
@@ -393,13 +395,26 @@ impl ProverServiceStore {
         .fetch_all(&mut *tx)
         .await?;
 
-        for row in existing_backend_sessions {
+        // A terminal backend session is immutable, so short-circuit before the
+        // active-session update below. Re-recording the same status is an idempotent
+        // retry, any other status is a conflict.
```
piohei, line 400: If other status is a conflict why do we only check for `terminal` one?

piohei (approved):

proofs/prover-service/src/store.rs
alessandromazza98: because we're only considering terminal statuses. non-terminal statuses can be updated of course. The comment may be misleading there, because I mean any other (terminal) status

alessandromazza98: let me change it

alessandromazza98: mmm actually no I think it's correct: the terminal status is the one stored in the db, but then the input could hold any status, therefore we must return an error if they don't match

alessandromazza98: examples:

stored Completed + attempted Failed, Running, or Submitting → all conflicts
stored Failed + attempted Completed, Running, or Submitting → all conflicts

PR #841, opened 2026-07-10
feat: add sp1-prover execute bin to get PGUs

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4883/estimate-pgu-for-sp1

### How to run

Here is how to run this command:

```rust
SP1_SKIP_PROGRAM_BUILD=true cargo run -p world-chain-prover-sp1 --release -- execute \
    --start-block 32151829 \
    --end-block 32151839 \
    --l2-rpc $WORLDCHAIN_PROVIDER \
    --l1-rpc $ETHEREUM_PROVIDER \
    --l1-beacon-rpc $ETHEREUM_BEACON_PROVIDER \
    --rollup-config ./rollup-config.json \
    --network worldchain
```

- it's important to use the `--release` flag, the dev binary is too slow and will fail the default 900s timeout
- you need to provide the rollup config, you can easily find it on alchemy [here](https://raas-backend.alchemy.com/rollups/worldchain-mainnet/rollup.json)
- you need to provide all rpcs, start and end block, remember that the start block is excluded, the end block is included

### How does it work

This command generates the witness by fetching all the data it needs from rpc, `SP1_SKIP_PROGRAM_BUILD=true` uses the already cached ELFs, remove it if the SP1 guest programs have not been built yet. Then it executes the sp1 proof work without actually generating the proof and prints a report like this:

```
execution succeeded
range:          32151830..=32151839
estimated PGUs: 11862381806
total cycles:   11840565729
total syscalls: 2372111
public values:  0x20000000000000008c189c96b1e9613416df0c38a2d09401d905ccbb85e29d7ecdd9e313857077762000000000000000ba6ff907ceec325361d9a9346370d7c527c05445bdd5e296ec2a53b9af4a6c0c2000000000000000fb7daa1506535c3b4cc3c2cdbd1b3cfba7175fc24ff74f98162675f59cbb54881f99ea01000000002000000000000000af3c3aec57f209e4e8df4364e0cf0232074171c16498a98f52b9866121ba0d9a
```

### PGUs

For the block range I used (32151830..=32151839), the estimated required PGUs are 11862381806 or 11.86 billion units. Note that these blocks are pretty light - average 8M-9M gas used. This result is in line with some past results shared by succinct.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19274,6 +19274,7 @@ dependencies = [
  "world-chain-proof-core",
  "world-chain-proof-succinct-elfs",
  "world-chain-proof-succinct-host-utils",
+ "world-chain-proof-succinct-utils",
  "world-chain-prover",
 ]
 
--- a/docs/proof/elf-management.md
+++ b/docs/proof/elf-management.md
@@ -8,9 +8,9 @@ The World Chain fault-proof system runs two SP1 guest programs:
 | `world-chain-proof-succinct-aggregation`     | Aggregates many range proofs into one     | `proofs/succinct/programs/aggregation`     |
 
 Both are compiled to RISC-V ELFs by `cargo prove build` (the SP1 toolchain) and are consumed by
-the `proof` CLI, the SP1 worker, and the devnet's full-stack tests. They are also referenced on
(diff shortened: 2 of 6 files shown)


Kemperino (approved):
this is really nice and helpful

PR #850, opened 2026-07-13
feat: world chain BFT specification

0xOsiris:
> [!NOTE]
> **Low Risk**
> Documentation-only draft WIP and index updates; no runtime or protocol implementation in this PR.
> 
> **Overview**
> Adds **draft WIP-1008** (*Staked Subblocks and Unsafe-Head BFT*), a Core standards-track proposal for a WLD-staked validator overlay on the unsafe chain while keeping the sequencer as sole block producer and L1 derivation unchanged.
> 
> The spec introduces a **`StakeManager`** L2 predeploy, per-block **subblock committees** (signed tx regions at the block front, integrated with WIP-1007 FBALs and flashblock index 0), **BLS attestations** and **≥2/3 stake attestation certificates**, a new **`attested`** RPC head between `unsafe` and `safe`, **`flblk` v3** gossip for subblocks/attestations/certificates, slashing (including builder bond S4), and rewards. It also documents graceful degradation when validators are offline.
> 
> **`specs/SUMMARY.md`** and **`wips/README.md`** are updated to list WIP-1008; the WIP index gains **WIP-1007** (Flashblock Access Lists) where it was missing from the table.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa902ec34ca54e87885924e3647b6a705c85f99b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wips/wip-1008.md
```diff
@@ -0,0 +1,268 @@
...
+
+Provisional values; final numbers to be fixed before Review status.
+
+| Parameter | Value | Notes |
+| --- | --- | --- |
+| `EPOCH_LENGTH` | 1800 blocks | 1 hour at 2 s blocks; validator set snapshots per epoch |
+| `MAX_ACTIVE_VALIDATORS` | 128 | top-N by stake |
+| `MIN_STAKE` | TBD (WLD) | governance-set |
+| `SUBBLOCK_COMMITTEE_SIZE` | 4 | proposers per block |
+| `SUBBLOCK_GAS_QUOTA` | `gas_limit / 20` | per subblock (committee total: 20% of block gas) |
```
alessandromazza98: nit: gas limit / 20 is not 20% of block gas, it's 5%. Or am I not understanding this correctly?

wips/wip-1008.md
```diff
@@ -0,0 +1,268 @@
...
+
+The mandatory access lists are what make step 3 parallel rather than sequential. On receipt of each subblock — before the deadline, off the block-building critical path — the builder MUST verify it under WIP-1007: materialize each transaction's pre-state from the FBAL, execute all of the subblock's transactions in parallel against isolated snapshots, and check that the re-derived access list hashes to `access_list_hash` (validity condition 5). A subblock that fails verification is invalid, MUST be excluded, and does not bind attestors. At sealing time the builder stitches the pre-verified regions: where a subblock's FBAL pre-state claims match the actual parent state of block `N` and no earlier-ordered subblock wrote to overlapping state, the cached execution results and state diffs are applied directly with no re-execution; only transactions whose pre-state diverges (stale `parent_hash`, cross-subblock overlap) are re-executed sequentially under the skip rule of step 3. The executed subblock region is emitted as **flashblock index 0** with its own composite FBAL, so downstream flashblock consumers are unaffected and the 200 ms streaming cadence is preserved.
```
alessandromazza98, line 145: am I understanding it right that only the first subblock will be applied without re-execution. The other 3 subblocks (considering a committee of 4) need to be executed sequentially because they should build on top of the previous one.

wips/wip-1008.md
```diff
@@ -0,0 +1,268 @@
...
+    pub participation: BitVec,
+}
+```
+
+#### Attestation validity policy
+
+Upon importing unsafe block `N` (via flashblocks or op-node gossip), an active validator MUST sign and broadcast an attestation iff **all** hold:
+
+1. The block is valid under World Chain execution rules (the validator's node has executed it).
+2. The block's subblock region matches the deterministic expected region: every valid subblock the validator observed on the P2P network at least `SUBBLOCK_OBSERVATION_MARGIN` (provisionally 200 ms) before `SUBBLOCK_DEADLINE` is present, correctly ordered, and correctly executed.
```
alessandromazza98, line 193: validators should then also execute subblocks sequentially to verify this correct? Otherwise how can they be sure that all subblocks are correct? The first one is easy because it's exactly equal to the one they've received through p2p as well. But 2-3-4 subblocks may be different due to the fact that they're now built on top of the previous subblock

alessandromazza98 (approved):
generally lgtm👍

My main question is whether it's worth it or not. The work is not super small and pretty time sensitive. Right now the batcher posts almost every 3-4 minutes, therefore the time window in which a block goes from unsafe to safe is already small.

That said, it's an interesting wip

wips/wip-1008.md
0xOsiris: I see, yes I overlooked this. I don't think it's possible to provide access lists without a lot of sequential coordination

PR #867, opened 2026-07-17
feat: authenticate proof games through factory registry

Kemperino:
This PR adds:
- One-time factory registration on the anchor registry.
- Authentication of factory-created games through isFactoryGame.
- Parent validation during proposals, including factory, registry, domain, blacklist, state, and block-range checks.
- Immutable transition context on every game, including starting root/block and L1 origin
- Factory enumeration through gameCount() and gameAt() so stateless services can recover after restart.
- Stronger registry checks before accepting finalized games.

> [!NOTE]
> **High Risk**
> Changes L1 anchor acceptance and proposal gating for the proof system; misconfiguration or logic bugs could block valid roots or weaken game authentication.
> 
> **Overview**
> **Binds the anchor registry to a single proof factory** and only accepts anchor updates from games that factory created and registered.
> 
> The registry gains one-time owner `initializeFactory`, stores `proofSystemFactory`, and `setAnchorState` now checks the game’s factory/registry match, `factory.isGame(game)`, and existing finalize/parent rules. The factory constructor takes an immutable `IWorldChainAnchorStateRegistry`, tracks `isGame` / `gameCount` / `gameAt`, and runs `_validateParent` on `propose` (registry not paused, factory initialized on registry, valid parent chain, exact `blockInterval` step, target L2 block ahead of current anchor).
> 
> Each game is stamped at deploy with `factory`, `anchorStateRegistry`, `attempt`, and starting anchor/parent root metadata. Devnet deploy calls `initializeFactory` after factory deploy; tests cover init-once, propose without init, parent/interval rules, and cross-factory anchor rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f3387c2548611c136a1f3adc398d79804360d27. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/contracts/src/proofs/WorldChainProofSystemFactory.sol
+++ b/pkg/contracts/src/proofs/WorldChainProofSystemFactory.sol
@@ -59,7 +59,7 @@ contract WorldChainProofSystemFactory {
     IWorldChainStakingRegistry public immutable stakingRegistry;
 
     mapping(bytes32 proposalKey => address game) public games;
-    mapping(address game => bool created) public isGame;
+    mapping(address game => bool created) public isFactoryGame;
     address[] private allGames;
 
     constructor(
@@ -151,7 +151,7 @@ contract WorldChainProofSystemFactory {
             )
         );
         games[key] = game;
-        isGame[game] = true;
+        isFactoryGame[game] = true;
         allGames.push(game);
 
         _emitGameCreated(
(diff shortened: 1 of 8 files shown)


Kemperino:
@codex review

pkg/contracts/src/proofs/WorldChainProofSystemFactory.sol
```diff
@@ -49,9 +59,13 @@ contract WorldChainProofSystemFactory {
     IWorldChainStakingRegistry public immutable stakingRegistry;
 
     mapping(bytes32 proposalKey => address game) public games;
+    mapping(address game => bool created) public isFactoryGame;
+    // Lets stateless services reconstruct all games after restart without relying on historical logs.
+    address[] private allGames;
```
Kemperino, line 64: for these the alternative is building some kind of indexer into the propser ^^

alessandromazza98 (approved):
lgtm. Ideally in the future we should be able to have a better enum that shows invalidation reasons. Because we only want to retry a game if the invalidation reason is proof threshold not met (if the reason is different for example blacklisted or parent invalid, we should not be able to retry) 

EDIT: actually it's already like this (i.e. if the parent is not valid, it reverts)

PR #876, opened 2026-07-17
feat: world chain proof system game: pull based value resolving

Kemperino:
## Summary

Converts proof-game bond settlement from push payments during `resolve()` to pull-based claims:

- `resolve()` records bond credits without making external ETH calls.
- Adds `claimable(recipient)` and permissionless `withdraw(recipient)`.
- Withdrawals always pay the named recipient, never the caller.
- Protects withdrawals with `ReentrancyGuardTransient`.
- Preserves existing finalization and invalidation bond economics.
- Documents the keeper/automation withdrawal flow.
- Exposes the new methods through `IWorldChainProofSystemGame`.

## Why

A reverting proposer or challenger could previously block game resolution because settlement transferred ETH synchronously. Separating resolution from payment keeps game progression independent of recipient behavior.

## Bond distribution

- Finalization: the proposer receives the proposer bond and all challenger bonds.
- Proof timeout: the first challenger receives the proposer bond; every challenger recovers their principal.
- Invalid parent or blacklist: the proposer and all challengers recover their principal.

## Tests

Adds coverage for:

- Finalization and invalidation claim accounting
- First-challenger rewards after proof timeout
- Invalid-parent and blacklist refunds
- Permissionless keeper withdrawals
- Prevention of withdrawal redirection
- Resolution involving recipients that reject ETH
- Preservation of claims after failed withdrawals

## Risk

High: this changes funds-critical bond accounting and introduces a new withdrawal path. The settlement API from the base branch remains intact, with `claimable` and `withdraw` added for payout collection.

### DIFF
--- a/pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
+++ b/pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
@@ -194,12 +194,11 @@ contract WorldChainProofSystemGame is ReentrancyGuardTransient {
     ///      the caller cannot redirect funds away from `recipient`.
     function withdraw(address payable recipient) external nonReentrant {
         address account = recipient;
-        uint256 credit = payoutCredits[account];
-        uint256 refundablePrincipal = _refundableChallengerPrincipal(account);
-        uint256 amount = credit + refundablePrincipal;
+        uint256 amount = _claimable(account);
         if (amount == 0) revert NoClaim(account);
 
-        if (credit != 0) payoutCredits[account] = 0;
+        uint256 refundablePrincipal = _refundableChallengerPrincipal(account);
+        payoutCredits[account] = 0;
         if (refundablePrincipal != 0) postedChallengerBond = 0;
 
         _transfer(recipient, amount);
@@ -360,6 +359,7 @@ contract WorldChainProofSystemGame is ReentrancyGuardTransient {
         uint256 balance = address(this).balance;
         if (reason == WorldChainProofLib.InvalidationReason.PROOF_TIMEOUT) {
             payoutCredits[challenger] += balance;
+            // The full-balance credit already includes the challenger bond, so it is no longer separately refundable.
             postedChallengerBond = 0;
         } else {
             uint256 proposerRefund = balance - postedChallengerBond;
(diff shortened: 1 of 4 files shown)


Kemperino:
@cursoragent refresh pr description
@codex review

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -110,8 +113,12 @@ contract WorldChainProofSystemGame {
     WorldChainProofLib.RootState public state;
     WorldChainProofLib.InvalidationReason public invalidationReason;
 
+    // TODO: Revisit whether the game should allow only one challenger; additional challengers do not
```
Kemperino: /otto create ticket in PROTO for proof system project

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -148,10 +170,10 @@ contract WorldChainProofSystemGame {
         }
         if (!stakingRegistry.isStaked(msg.sender)) revert UnstakedChallenger(msg.sender);
         if (msg.value != challengerBond) revert InvalidBond(challengerBond, msg.value);
-        if (challengerBonds[msg.sender] != 0) revert DuplicateChallenge(msg.sender);
+        if (challenger != address(0)) revert DuplicateChallenge(challenger);
```
alessandromazza98, line 173: question: ok so we're only making it possible to challenge once - it makes sense to me.

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -148,10 +170,10 @@ contract WorldChainProofSystemGame {
...
 
-        challengerBonds[msg.sender] = msg.value;
-        challengers.push(msg.sender);
+        challenger = payable(msg.sender);
+        postedChallengerBond = msg.value;
```
alessandromazza98, line 176: question: do you need this variable if previously you're checking that `msg.value` must be equal to `challengerBond`? Therefore the value transferred is always equal to `challengerBond` at this point

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
```diff
@@ -192,55 +235,138 @@ contract WorldChainProofSystemGame {
...
-        }
-
-        uint256 forfeited = address(this).balance;
-        if (forfeited != 0 && challengers.length != 0) {
-            _transfer(payable(challengers[0]), forfeited);
-        } else if (forfeited != 0) {
-            _transfer(proposer, forfeited);
+        uint256 balance = address(this).balance;
+        if (reason == WorldChainProofLib.InvalidationReason.PROOF_TIMEOUT) {
+            payoutCredits[challenger] += balance;
```
alessandromazza98, line 361: question: `challenger` here is always initialized?

alessandromazza98 (approved):
lgtm, I left some non-blocking questions

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
alessandromazza98: ok I think yes it's always initialized at this point

pkg/contracts/src/proofs/WorldChainProofSystemGame.sol
alessandromazza98: EDIT: I see it's used in the code, I get the point of this variable 👍

PR #877, opened 2026-07-18
feat: Anchor State Registry jumping intermediate finalized games

Kemperino:
## Summary

- Allow the registry to jump directly to a finalized descendant game.
- Require the candidate’s ancestry to reach the current anchor root and block.
- Reject parallel branches and blacklisted ancestors.
- Add a post-resolution finality delay and game validity helpers.
- Add `game.closeGame()` as a convenience wrapper around `setAnchorState()`.

This removes the now-unused `acceptedGames` and `acceptedRoots` mappings.

### DIFF
--- a/pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
+++ b/pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
@@ -80,8 +80,8 @@ contract WorldChainAnchorStateRegistry {
 
     /// @notice Returns whether a game is recognized by this registry and currently has a valid finalized claim.
     /// @dev Anchor advancement additionally requires a newer L2 block number.
-    /// TODO: Before withdrawal integration, define Base-compatible retirement and emergency anchor recovery semantics,
-    /// implement the OP IAnchorStateRegistry surface, and wire OptimismPortal to this predicate.
+    /// TODO: Before withdrawal integration, define OP-compatible retirement and emergency anchor recovery semantics,
+    /// implement the IAnchorStateRegistry surface, and wire OptimismPortal to this predicate.
     function isGameClaimValid(address game) external view returns (bool) {
         address factory = proofSystemFactory;
         if (paused || factory == address(0) || blacklistedGames[game]) return false;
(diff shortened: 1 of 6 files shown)


pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
```diff
@@ -33,8 +32,6 @@ contract WorldChainAnchorStateRegistry {
     uint256 public currentL2BlockNumber;
     address public anchorGame;
 
-    mapping(address game => bool accepted) public acceptedGames;
-    mapping(bytes32 rootId => bool accepted) public acceptedRoots;
```
Kemperino, line 37: do we need these? acceptedGames so far for trailing though the games, but not sure there is any other use for this storage that we don't get by indexing? cc: @alessandromazza98

alessandromazza98: ok to remove them now

pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
```diff
@@ -70,6 +71,38 @@ contract WorldChainAnchorStateRegistry {
...
+        try proofGame.finalizedAt() returns (uint64 finalizedAt) {
+            return block.timestamp >= uint256(finalizedAt) + finalityDelay;
+        } catch {
+            return false;
+        }
+    }
+
+    /// @notice Returns whether a game is recognized by this registry and currently has a valid finalized claim.
+    /// @dev Anchor advancement additionally requires a newer block and continuity with the current accepted anchor.
+    /// TODO: Before withdrawal integration, define Base-compatible retirement and emergency anchor recovery semantics,
```
alessandromazza98: why Base compatible? it's OP compatible right?

pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
```diff
@@ -26,19 +27,19 @@ contract WorldChainAnchorStateRegistry {
 
     address public owner;
     address public proofSystemFactory;
+    uint64 public immutable finalityDelay;
```
alessandromazza98: what's the reason for this `finalityDelay`? If we set reasonable finality timestamps, why would we need an additional finality delay here?

alessandromazza98 (approved):

pkg/contracts/src/proofs/WorldChainAnchorStateRegistry.sol
Kemperino: removed for now.

PR #898, opened 2026-07-21
refactor: world chain proposer

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4959/proposer-refactor-proposer-for-the-new-contract-design

This PR refactors the `world-chain-proposer` to account for the new contract design.

### Next Steps

- add withdraw credit logic: https://linear.app/worldcoin/issue/PROTO-4976/proposer-add-withdraw-credits-logic
- refactor challenger
- refactor defender

---

> [!NOTE]
> **Medium Risk**
> Changes L1 proposer behavior and on-chain interaction paths (resolve/close vs implicit finalize); mistakes could stall proposals or anchor advancement, though fakes and integration tests cover the new flow.
> 
> **Overview**
> Refactors **`world-chain-proposer`** for the new proof-system contract lifecycle: games are no longer finalized when proof threshold is met on-chain; the keeper must **`resolve()`** positively resolvable games and **`closeGame()`** on the highest finalized canonical game to advance the anchor.
> 
> The main loop is now **scan canonical line → resolve → advance anchor → propose/retry**, replacing **`propose_once` / `prepare_next_proposal`**. **`anchor_and_canonical_line`** walks from the anchor, classifies invalidated games (retry on proof timeout, wait on negative resolution, block on other invalidations), and returns **`CanonicalScan` / `NextProposalAction`**. **`ProposerClient`** and **`AlloyProofSystemClient`** gain **`resolution_status`**, **`resolve_game`**, and **`close_game`**; Solidity bindings swap **`finalize`/`invalidate`** for **`resolutionStatus`**, **`resolve`**, and **`closeGame`**.
> 
> Primitives add **`ResolutionStatus`**, **`InvalidationReason`**, and updated game/registry ABI surface. **`FakeExecution`** mirrors contract semantics (proofs only set bitmap; finalize via resolve; anchor via close). Integration tests wait on **`has_threshold`** then **`settle_with_proposer`**, plus a dedicated resolution-semantics test. Prover-service tests expire locks via SQL instead of sleeping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21e6558a5725356d625ebcc2782efee0b0ecc34c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proofs/proposer/src/alloy.rs
+++ b/proofs/proposer/src/alloy.rs
@@ -3,10 +3,13 @@ use alloy_provider::Provider;
 use async_trait::async_trait;
 use world_chain_proofs::{
     IWorldChainAnchorStateRegistry, IWorldChainProofSystemFactory, IWorldChainProofSystemGame,
-    ProposalCommitment,
+    ProposalCommitment, ResolutionStatus,
 };
 
-use crate::{ParentRef, Proposal, ProposalSubmission, ProposerClient, ProposerError};
+use crate::{
+    ParentRef, Proposal, ProposalSubmission, ProposerClient, ProposerError,
+    types::{CloseGameSubmission, ResolveSubmission},
+};
 
 /// Alloy-backed implementation of [`ProofSystemClient`].
 #[derive(Debug, Clone)]
@@ -88,6 +91,79 @@ where
             .map_err(|error| ProposerError::Contract(error.to_string()))?;
 
         Ok((game != Address::ZERO).then_some(game))
+    }
+
+    async fn resolution_status(&self, game: Address) -> Result<ResolutionStatus, ProposerError> {
+        let game = IWorldChainProofSystemGame::IWorldChainProofSystemGameInstance::new(
+            game,
+            self.provider.clone(),
+        );
+        let resolution_status_result = game
+            .resolutionStatus()
+            .call()
+            .await
+            .map_err(|err| ProposerError::Contract(err.to_string()))?;
+        let resolvable = resolution_status_result.resolvable;
+        let root_state = resolution_status_result
+            .outcome
+            .try_into()
+            .map_err(|err| ProposerError::Contract(err.to_string()))?;
+        let invalidation_reason = resolution_status_result
+            .reason
+            .try_into()
+            .map_err(|err| ProposerError::Contract(err.to_string()))?;
+        let resolution_status = ResolutionStatus {
+            resolvable,
+            root_state,
+            invalidation_reason,
+        };
+        Ok(resolution_status)
+    }
+
+    async fn resolve_game(&self, game: Address) -> Result<ResolveSubmission, ProposerError> {
+        let game = IWorldChainProofSystemGame::IWorldChainProofSystemGameInstance::new(
+            game,
+            self.provider.clone(),
+        );
+        let pending = game
+            .resolve()
+            .send()
+            .await
+            .map_err(|error| ProposerError::Contract(error.to_string()))?;
+
+        let tx_hash = *pending.tx_hash();
+        let receipt = pending
+            .get_receipt()
+            .await
+            .map_err(|error| ProposerError::Contract(error.to_string()))?;
+        if !receipt.status() {
+            return Err(ProposerError::Revert(tx_hash));
+        }
+
+        Ok(ResolveSubmission { tx_hash })
+    }
+
+    async fn close_game(&self, game: Address) -> Result<CloseGameSubmission, ProposerError> {
+        let game = IWorldChainProofSystemGame::IWorldChainProofSystemGameInstance::new(
+            game,
+            self.provider.clone(),
+        );
+        let pending = game
+            .closeGame()
+            .send()
+            .await
+            .map_err(|error| ProposerError::Contract(error.to_string()))?;
+
+        let tx_hash = *pending.tx_hash();
+        let receipt = pending
+            .get_receipt()
+            .await
+            .map_err(|error| ProposerError::Contract(error.to_string()))?;
+        if !receipt.status() {
+            return Err(ProposerError::Revert(tx_hash));
+        }
+
+        Ok(CloseGameSubmission { tx_hash })
     }
 
     async fn submit_proposal(
--- a/proofs/proposer/src/alloy.rs
+++ b/proofs/proposer/src/alloy.rs
@@ -3,7 +3,7 @@ use alloy_provider::Provider;
 use async_trait::async_trait;
 use world_chain_proofs::{
     IWorldChainAnchorStateRegistry, IWorldChainProofSystemFactory, IWorldChainProofSystemGame,
-    ProposalCommitment, ResolutionStatus,
+    InvalidationReasonError, ProposalCommitment, ResolutionStatus, RootStateError,
 };
 
 use crate::{
@@ -107,11 +107,11 @@ where
         let root_state = resolution_status_result
             .outcome
             .try_into()
-            .map_err(|err| ProposerError::Contract(err.to_string()))?;
+            .map_err(|err: RootStateError| ProposerError::Contract(err.to_string()))?;
         let invalidation_reason = resolution_status_result
             .reason
             .try_into()
-            .map_err(|err| ProposerError::Contract(err.to_string()))?;
+            .map_err(|err: InvalidationReasonError| ProposerError::Contract(err.to_string()))?;
         let resolution_status = ResolutionStatus {
             resolvable,
             root_state,
--- a/proofs/proposer/src/proposer.rs
+++ b/proofs/proposer/src/proposer.rs
@@ -1,9 +1,9 @@
 use alloy_primitives::B256;
-use tracing::{debug, info, warn};
+use tracing::{info, warn};
 use world_chain_proofs::ConsensusProvider;
 
 use crate::{
-    ParentRef, Proposal, ProposalSubmission, ProposerClient, ProposerConfig, ProposerError,
+    ParentRef, Proposal, ProposerClient, ProposerConfig, ProposerError,
     types::{CanonicalLine, ResolvedGames},
 };
 
--- a/proofs/proposer/src/proposer.rs
+++ b/proofs/proposer/src/proposer.rs
@@ -1,10 +1,10 @@
 use alloy_primitives::B256;
 use tracing::{info, warn};
-use world_chain_proofs::ConsensusProvider;
+use world_chain_proofs::{ConsensusProvider, RootState};
 
 use crate::{
     ParentRef, Proposal, ProposerClient, ProposerConfig, ProposerError,
-    types::{CanonicalLine, ResolvedGames},
+    types::{CanonicalLine, FinalizedGames},
 };
 
 /// World Chain Proposer.
@@ -103,8 +103,8 @@ where
     pub async fn resolve_games(
         &self,
         canonical_line: &CanonicalLine,
-    ) -> Result<ResolvedGames, ProposerError> {
-        let mut resolved_games = ResolvedGames::default();
+    ) -> Result<FinalizedGames, ProposerError> {
+        let mut finalized_games = FinalizedGames::default();
         for game in canonical_line.games() {
             let resolution_status = self
                 .execution_provider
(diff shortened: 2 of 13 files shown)


proofs/prover-service/src/tests.rs
```diff
@@ -50,6 +50,23 @@ async fn service(config: ProverServiceConfig) -> Option<TestService> {
     })
 }
 
+async fn expire_proof_lock(service: &ProverService, id: ProofRequestId, lock_id: LockId) {
```
alessandromazza98, line 53: there was a test here that was flaky due to some timing issue, this fixes it. It's unrelated to the PR scope

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+                        NextProposalAction::RetryTimedOut {
+                            proposal,
+                            invalidated_game: next_game_addr,
+                        }
+                    } else {
+                        NextProposalAction::BlockedByInvalidation {
+                            game: next_game_addr,
+                            reason: resolution_status.invalidation_reason,
+                        }
+                    };
```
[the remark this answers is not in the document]
alessandromazza98: this is fine because the `world-chain-challenger` will take care of resolving that game. It's not proposer responsibility to resolve that game

proofs/proposer/src/alloy.rs
```diff
@@ -90,6 +93,79 @@ where
...
+        );
+        let pending = game
+            .resolve()
+            .send()
+            .await
+            .map_err(|error| ProposerError::Contract(error.to_string()))?;
+
+        let tx_hash = *pending.tx_hash();
+        let receipt = pending
+            .get_receipt()
```
Kemperino, line 136: i think it would make sense to extract a default setting for the provider on number of confirmations. e.g. lets wait for 4 blocks, then we have less hiccups in new ticks.

proofs/proposer/src/proposer.rs
```diff
@@ -108,28 +239,23 @@ where
...
-                        target_block,
-                        l2_finalized_block, "waiting for L2 to finalize the next proposal height"
-                    );
-                }
-                Err(error) => {
-                    warn!(%error, "proposal attempt failed");
-                }
+            let iteration: Result<(), ProposerError> = async {
+                // 1. refresh the anchor and canonical line
+                let canonical_scan = self.anchor_and_canonical_line().await?;
```
Kemperino, line 243: wait, i thought we want to record an error in one phase but still try to do the other ones? 

like this the async { } block would return on the first failed phase?

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+    /// Resolves positively resolvable games parent-first and returns all finalized games.
+    ///
+    /// Games finalized by an earlier iteration or another keeper are included so anchor
+    /// advancement can be retried.
+    pub async fn resolve_games(
+        &self,
+        canonical_line: &CanonicalLine,
+    ) -> Result<FinalizedGames, ProposerError> {
+        let mut finalized_games = FinalizedGames::default();
+        for game in canonical_line.games() {
```
Kemperino, line 143: only as long as we have per-tick transaction limit according to notion spec right?

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+                );
+                return Ok(());
+            }
+            NextProposalAction::CaughtUp { .. } => return Ok(()),
+        };
+
         let submission = self
             .execution_provider
-            .submit_proposal(&proposal, self.config.proposer_bond)
+            .submit_proposal(proposal, self.config.proposer_bond)
```
Kemperino, line 221: shouldnt the necessary proposer bond be read from the contract? or if the contract doesn't have this yet let's implement it this way?

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
     E: ProposerClient,
     C: ConsensusProvider,
```
Kemperino, line 38: do you think it would be worth clarifying that ExecutionProvider basically = L1 and that Consensus Layer is the canonical *L2* state?

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+            let next_l2_block_number = cursor
                 .l2_block_number
                 .checked_add(self.config.block_interval)
                 .ok_or(ProposerError::BlockNumberOverflow {
-                    parent_block: parent.l2_block_number,
+                    parent_block: cursor.l2_block_number,
                     block_interval: self.config.block_interval,
                 })?;
 
             let latest_finalized_l2_block =
```
Kemperino: do you think we should move this outside of the loop {} this makes the scan more deterministic and easier to reason about.

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+                        target_block: next_l2_block_number,
+                        finalized_block: latest_finalized_l2_block,
+                    },
+                ));
             }
 
             let root_claim = self
                 .consensus_provider
-                .output_root_at_block(l2_block_number)
+                .output_root_at_block(next_l2_block_number)
```
Kemperino, line 75: a bit more documentation would be nice here.

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+            NextProposalAction::AwaitNegativeResolution { game, reason } => {
+                info!(
+                    game_address = %game,
+                    invalidation_reason = ?reason,
+                    "waiting for challenger to resolve game with negative outcome"
+                );
+                return Ok(());
+            }
+            NextProposalAction::BlockedByInvalidation { game, reason } => {
+                warn!(
```
Kemperino, line 209: this probably should err

proofs/proposer/src/proposer.rs
```diff
@@ -36,69 +37,199 @@ where
...
+    /// challenger, and non-retryable invalidations stop with a governance warning.
+    pub async fn propose(&self, scan: &CanonicalScan) -> Result<(), ProposerError> {
+        let (proposal, retry_of) = match scan.next_action() {
+            NextProposalAction::Propose(proposal) => (proposal, None),
+            NextProposalAction::RetryTimedOut {
+                proposal,
+                invalidated_game,
+            } => (proposal, Some(*invalidated_game)),
+            NextProposalAction::AwaitNegativeResolution { game, reason } => {
+                info!(
```
Kemperino, line 201: this should be warn! i believe

proofs/proposer/src/proposer.rs
Kemperino: just to be clear this basically means our proposer computed the same proposal key but it was invalidated right?

proofs/proposer/src/proposer.rs
alessandromazza98: I was thinking that as well, but we do need some of the previous values to do later steps. We can also revisit this approach if we find out that it causes too many issues

proofs/proposer/src/alloy.rs
alessandromazza98: are you worried about reorgs?

proofs/proposer/src/proposer.rs
alessandromazza98: open to it. My idea is:

- execution provider: it is a execution client connection (in this case it's L1)
- consensus provider: it is a consensus client connection (L2 in this case)

proofs/proposer/src/proposer.rs
alessandromazza98: yes good point - I'm gonna do it

proofs/proposer/src/proposer.rs
alessandromazza98: what do you mean here?

proofs/proposer/src/proposer.rs
alessandromazza98: yes, this scenario means a valid game has been blacklisted - this requires a governance intervention (or an emergency protocol)

alessandromazza98: we don't have a way to solve this problem for now - we need to add an emergency protocol

proofs/proposer/src/proposer.rs
alessandromazza98: yes it can be read from the contract - I'll change this so that the config value is read from contract at startup

alessandromazza98: https://linear.app/worldcoin/issue/PROTO-4981/proposer-read-config-values-from-contracts

proofs/proposer/src/proposer.rs
alessandromazza98: let me create a ticket for this: https://linear.app/worldcoin/issue/PROTO-4982/proposer-add-more-docs-for-anchor-and-canonical-line-fn

proofs/proposer/src/proposer.rs
alessandromazza98: done

proofs/proposer/src/proposer.rs
alessandromazza98: potentially yes - it's an issue that requires intervention. It means that a valid game is going to become invalidated

proofs/proposer/src/proposer.rs
```diff
@@ -108,28 +238,23 @@ where
...
-                }
+            let iteration: Result<(), ProposerError> = async {
+                // 1. refresh the anchor and canonical line
+                let canonical_scan = self.anchor_and_canonical_line().await?;
+                // 2. resolve positive-ready games parent-first
+                let finalized_games = self.resolve_games(canonical_scan.canonical_line()).await?;
+                // 3. advance the anchor to the highest finalized canonical game
+                self.advance_anchor(finalized_games).await?;
+                // 4. attempt a new canonical proposal or retry
+                self.propose(&canonical_scan).await?;
```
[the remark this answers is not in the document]
alessandromazza98: this is fine because we're using finalized l2 blocks - so they cannot be reorged out and the root_claim cannot change for them

proofs/proposer/src/proposer.rs
alessandromazza98: but this will eventually go into the `BlockedByInvalidation` once the challenger resolve it

proofs/proposer/src/proposer.rs
alessandromazza98: actually we have - we can unblacklist the game. But still... we need an emergency protocol to handle critical issues

proofs/proposer/src/alloy.rs
0xOsiris: Yeah actually this is going to default to 1 confirmation in alloy. Could cause a bunch of problems downstream

alessandromazza98: Sure - even though a L2 reorg is an uncommon event. That said we could wait for more confirmations here (and in other places).

In general though I'll think a bit more about reorgs and eventually create patches in follow ups

0xOsiris: The risk here would be an Ethereum re-org no?

alessandromazza98: Yes I see your point

proofs/proposer/src/proposer.rs
Kemperino: yeah, worth a warn imo

proofs/proposer/src/proposer.rs
Kemperino: on our notion spec we defined a budget per tick to avoid blocking too long on one iteration. i dont think it's needed necessarily but wanted to point out it's missing here atm.

Kemperino (approved):

proofs/proposer/src/proposer.rs
alessandromazza98: I understand - let me create a ticket for this

alessandromazza98: https://linear.app/worldcoin/issue/PROTO-4985/defender-add-budget-per-tick-to-avoid-starvation-on-a-single-operation

proofs/proposer/src/alloy.rs
alessandromazza98: I created a ticket so that I can later think about a long term solution: https://linear.app/worldcoin/issue/PROTO-4986/l1-reorgs-how-to-handle-reorgs-appropriately

PR #900, opened 2026-07-21
feat(proposer): add withdrawal credits logic

alessandromazza98:
Closes https://linear.app/worldcoin/issue/PROTO-4976/proposer-add-withdraw-credits-logic

### Notes

This PR adds the withdrawal credit logic into the `world-chain-proposer`. Proposed games are kept in-memory and each tick it tries to `withdraw` from them if possible.

If the `world-chain-proposer` shuts down and it gets restarted, historical games are forgotten. Therefore we need a separated binary to handle this case. Further, this binary will be agnostic to proposer because the `world-chain-challenger` will need it too.

### Next Steps

- create binary to handle historical game withdrawals: https://linear.app/worldcoin/issue/PROTO-4979/new-binary-handle-historical-games-withdrawals

---

> [!NOTE]
> **Medium Risk**
> Introduces recurring L1 `withdraw` transactions and bond-recovery logic tied to resolution semantics; incorrect `is_resolved` or receipt parsing could skip or mis-handle funds, though scope is limited to games proposed by this proposer.
> 
> **Overview**
> **Adds automatic withdrawal of proposer credits** for games this instance created. Each poll cycle runs `withdraw_credits` before canonical scanning: for tracked game addresses it checks terminal resolution via `ResolutionStatus::is_resolved()`, calls `claimable` / `withdraw` when balance is non-zero, then drops the game from the in-memory set.
> 
> `WorldChainProposer` is now **mutable** (`proposed_games: HashSet<Address>`); successful `propose` records `game_address` from the submission. `ProposerClient` gains `claimable` and `withdraw`; the Alloy client sends `withdraw`, requires `WalletProvider`, and parses `GameCreated` / `Withdrawn` from receipts. `ProposalSubmission` includes `game_address`. Fakes and integration tests follow the new trait and `mut` proposer.
> 
> **Note:** Tracked games live only in process memory—restart forgets prior proposals (historical withdrawal is planned separately).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d586e1f70e067bc24cf6004c488db5eba355d85a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proofs/proposer/src/proposer.rs
+++ b/proofs/proposer/src/proposer.rs
@@ -6,7 +6,7 @@ use world_chain_proofs::{ConsensusProvider, InvalidationReason, RootState};
 
 use crate::{
     ParentRef, Proposal, ProposerClient, ProposerConfig, ProposerError,
-    types::{CanonicalLine, CanonicalScan,  FinalizedGames, NextProposalAction},
+    types::{CanonicalLine, CanonicalScan, FinalizedGames, NextProposalAction},
 };
 
 /// World Chain Proposer.
@@ -194,6 +194,8 @@ where
     ///
     /// New and timed-out transitions are submitted, negative-ready games wait for the
     /// challenger, and non-retryable invalidations stop with a governance warning.
+    ///
+    /// The newly created game is added to the `self.proposed_games` field.
     pub async fn propose(&mut self, scan: &CanonicalScan) -> Result<(), ProposerError> {
         let (proposal, retry_of) = match scan.next_action() {
             NextProposalAction::Propose(proposal) => (proposal, None),
@@ -237,6 +239,8 @@ where
         Ok(())
     }
 
+    /// Scan all games stored in `self.proposed_games` and withdraw from them
+    /// if `claimable_amount` is greater than zero.
     pub async fn withdraw_credits(&mut self) -> Result<(), ProposerError> {
         let mut withdrawn = Vec::new();
         for &game in &self.proposed_games {
--- a/proofs/proposer/src/proposer.rs
+++ b/proofs/proposer/src/proposer.rs
@@ -1,10 +1,12 @@
-use alloy_primitives::B256;
+use std::collections::HashSet;
+
+use alloy_primitives::{Address, B256, U256};
 use tracing::{info, warn};
 use world_chain_proofs::{ConsensusProvider, InvalidationReason, RootState};
 
 use crate::{
     ParentRef, Proposal, ProposerClient, ProposerConfig, ProposerError,
-    types::{CanonicalLine, CanonicalScan, FinalizedGames, NextProposalAction},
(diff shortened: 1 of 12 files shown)


proofs/proposer/src/proposer.rs
```diff
@@ -223,33 +229,80 @@ where
...
+                        "withdrew claimable credits"
+                    );
+                }
+
+                Ok(true)
+            }
+            .await;
+
+            match result {
+                Ok(true) => {
```
Kemperino, line 270: this arm also matches if resolved but challenger got payout right? worth commenting

Kemperino (approved):

proofs/proposer/src/proposer.rs
```diff
@@ -223,33 +229,80 @@ where
...
+                    self.proposed_games.remove(&game);
```
Kemperino, line 271: here a reorg WOULD be problematic though right? :)

proofs/proposer/src/proposer.rs
alessandromazza98: yes

proofs/proposer/src/proposer.rs
alessandromazza98: an l2 reorg you mean?

Kemperino: no l1. then we remove it from proposed game thinking we did the payout? but then it might reorg out?

alessandromazza98: that's a valid point. I was thinking that the scan of these games can happen up to the last finalized l1 block so no reorgs may change anything.

creating ticket: https://linear.app/worldcoin/issue/PROTO-4983/proposer-scan-games-up-to-the-last-l1-finalized-block-to-avoid-l1

EDIT: actually this doesn't solve the problem at all. Yes we take the risk of l1 reorgs. It's quite rare that after a reorg a tx is not included. If that happens we can restart the proposer - or eventually we can just withdraw manually

alessandromazza98: I'm gonna think about it a bit more though

Kemperino: honestly i am probably even confident enough with 2/3 confirmations or at least writing a metric so that we notice if funds are missing and can back-check. (the probabilty of deeper reorg + that this happens on our game is quite quite low)

alessandromazza98: I created a ticket for this as well

proofs/proposer/src/proposer.rs
```diff
@@ -222,33 +228,80 @@ where
...
```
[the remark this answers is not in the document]
alessandromazza98: I have a ticket for this and I'll work on it

PR #923, opened 2026-07-27
Add latest flashblock simulation mode

karankurbur:
## Summary

- Add optional `useLatestFlashblock` request parameter to `simulate_unsignedUserOp`, defaulting to `false`.
- Wire the simulate RPC to use the latest executed flashblock state when requested, while preserving the existing latest/full-block execution path otherwise.
- Validate flashblock usage at the endpoint boundary: reject when flashblocks are disabled, when no latest flashblock exists yet, or when an explicit `block` is also provided.
- Add endpoint-level integration tests for the new flashblock guard behavior.

## Validation

- `cargo fmt`
- `cargo test -p world-chain-rpc`
- `cargo check -p world-chain-node`
- `cargo check`
- `git diff --check`

`cargo clippy -p world-chain-rpc --all-targets -- -D warnings` could not run because this pinned nightly reports `cargo-clippy` is not applicable to `nightly-2026-07-01-aarch64-apple-darwin`.

---

> [!NOTE]
> **Medium Risk**
> Changes how simulation state is chosen for a new RPC flag and depends on flashblock pending state staying consistent with ETH pending behavior; default behavior is unchanged.
> 
> **Overview**
> Adds optional **`useLatestFlashblock`** (default `false`) on `simulate_unsignedUserOp` so simulations can run against the **latest executed flashblock** instead of a canonical block.
> 
> When the flag is set, block/state resolution reads the pending flashblock from a shared `watch` receiver, uses that block’s header, and builds state from the parent block plus a `BlockState` overlay over the executed flashblock. The existing block-id path is unchanged when the flag is off.
> 
> The node threads the same `pending_block()` receiver used for flashblocks ETH APIs into `WorldChainAddOns` and the `Simulate` RPC via `with_latest_flashblock`. Requests are rejected if flashblocks are disabled, no flashblock is available yet, or `block` is set together with `useLatestFlashblock`.
> 
> Unit and integration tests cover deserialization, param validation, and the RPC error paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a018d3e658cdc13b0c7bc0334fdb70f3622c42e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


karankurbur:
Addressed the parent state lookup review note in `latest_flashblock_state`: the flashblock overlay now uses `state_by_block_hash(header.parent_hash)` instead of `history_by_block_hash(...)`, so it can resolve live tip/in-memory parent state.

Local follow-up validation:
- `cargo fmt` passed
- `git diff --check` passed
- Cargo compile/test rerun was blocked in this fresh checkout by local disk space while building `librocksdb-sys` (`No space left on device`, filesystem had ~1.1GiB free).

karankurbur:
Backwards compatibility pass pushed in `97f95b63`:

- Restored the original `Simulate::new(client, evm_config, task_pool, task_guard)` signature.
- Restored the original `Simulate::from_eth_api(client, evm_config, eth_api)` signature.
- Restored the original `WorldChainAddOns::new(...)` signature.
- Added explicit `.with_latest_flashblock(...)` opt-in methods for the new flashblock receiver wiring.
- Added an endpoint-level integration test proving requests that omit `useLatestFlashblock` still default to the latest full-block path and do not require flashblocks.

`cargo fmt`, `cargo metadata --no-deps`, and `git diff --check` pass locally. Full cargo test/check reruns remain blocked by local disk space while compiling `librocksdb-sys` in the fresh checkout (`No space left on device`, about 1.0GiB free).

crates/node/src/add_ons.rs
```diff
@@ -290,6 +310,7 @@ where
             simulate_enabled,
             witness,
         )
+        .with_latest_flashblock(pending_block)
```
crystalt, line 313: nit: is there a way we could avoid adding this plumbing type code to all these places? codex said we could add a trait to `FlashblocksEthApi`? Something like this

```
pub trait PendingFlashblockSource {
    fn pending_flashblock_receiver(&self) -> Option<LatestFlashblockReceiver>;
}
```

codex said, 

> `FlashblocksEthApi` could implement this by cloning its optional watch receiver. Then `Simulate::from_eth_api` could require `SpawnBlocking + PendingFlashblockSource` and obtain the receiver from `registry.eth_api()`. That should eliminate the `pending_block` field and `with_latest_flashblock` method on `WorldChainAddOns`, including the repeated forwarding through its generic builder transformations.

just a nit though.

crystalt (approved):

Kemperino (approved):

crates/rpc/src/simulate.rs
```diff
@@ -472,6 +477,9 @@ pub trait SimulateApi {
 /// Cross-request cache for resolved token metadata. LRU-bounded.
 type MetadataCache = Arc<Mutex<LruCache<Address, AssetInfo>>>;
 
+pub type LatestFlashblockReceiver =
```
0xOsiris, line 480: Remove this type alias, it's just an indirection that makes the logic harder to read

crates/node/src/add_ons.rs
0xOsiris: agreed

crates/node/src/add_ons.rs
```diff
@@ -152,10 +154,21 @@ where
...
             simulate_enabled,
+            pending_block: None,
             witness,
             _tx: PhantomData,
         }
     }
+
+    /// Configures access to the latest flashblock-backed pending block, when
+    /// flashblocks are enabled.
+    pub fn with_latest_flashblock(
```
0xOsiris, line 165: Remove this, just add it to WorldChainAddOns::new

crates/node/src/add_ons.rs
```diff
@@ -194,6 +208,7 @@ where
             simulate_enabled,
             witness,
         )
+        .with_latest_flashblock(pending_block)
```
0xOsiris, line 211: Remove

crates/node/src/add_ons.rs
```diff
@@ -226,6 +242,7 @@ where
...
```
0xOsiris, line 245: Remove

crates/node/src/add_ons.rs
```diff
@@ -258,6 +276,7 @@ where
...
```
0xOsiris, line 279: Remove

crates/node/src/add_ons.rs
0xOsiris, line 313: Remove

0xOsiris (changes requested):

crates/rpc/src/simulate.rs
```diff
@@ -56,6 +57,10 @@ pub struct SimulateUnsignedUserOpRequest {
     /// Block to simulate against. Defaults to `latest` when omitted.
     #[serde(default)]
     pub block: Option<BlockId>,
+    /// Whether to simulate against the latest flashblock state instead of the
+    /// latest full block state. Defaults to `false`.
+    #[serde(default)]
+    pub use_latest_flashblock: bool,
```
0xOsiris, line 63: You are breaking the wire format backwards compatibility of the request by not making this an Option

PR #932, opened 2026-07-29
feat: flashblock sentry node selection

0xForerunner:
Feature to reduce the number of connections to available sentry nodes. This should minimize required bandwidth to serve flashblocks.

---

> [!NOTE]
> **Medium Risk**
> Changes default P2P trust and ban behavior for flashblocks nodes; misconfiguration or selection bugs could reduce flashblock connectivity, though policy is covered by new tests.
> 
> **Overview**
> Flashblocks clients no longer trust every default flashblocks bootnode for RLPx. Defaults are loaded into a **sentry pool** (`flashblocks.sentry_peers`, renamed from bootnode constants), and the network layer picks up to **`--flashblocks.max-sentry-connections`** peers (default **1**) per node.
> 
> **Selection** uses rendezvous hashing on the persisted local `PeerId`, so choice is stable and spreads load across the pool. **All** pool members stay UDP discovery bootnodes; only the chosen sentries are added as trusted RLPx peers. Non-selected sentries are peer-manager banned for RLPx so they cannot be dialed via persisted/basic peers, while discovery can still use them.
> 
> Chain defaults for World mainnet/Sepolia populate `sentry_peers` at config time instead of extending `network.trusted_peers` immediately—sentry policy runs when the network is built and the local identity is known. Builders can raise `max-sentry-connections` to connect to every sentry. CLI reference and unit/e2e test fixtures were updated for the new flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b984eff7347e9c0c7e0534361a9b9709cfc16e7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/cli/src/cli.rs
+++ b/crates/cli/src/cli.rs
@@ -25,9 +25,9 @@ pub use builder::*;
 pub use p2p::*;
 pub use pbh::*;
 
-pub const DEFAULT_FLASHBLOCKS_BOOTNODES: &str = "enode://78ca7daeb63956cbc3985853d5699a6404d976a2612575563f46876968fdca2383a195ee7db40de348757b2256195996933708f351169ca3f3fe93ab2a774608@16.62.98.53:30303,enode://c96dcadf4cdea4c39ec3fd775637d9e67d455b856b1514cfcf55b72f873a34b96d69e47ccea9fc797a446d4e6948aa80f6b9d479a1727ca166758a900b08f422@16.63.14.166:30303,enode://15688a7b281c32a4da633252dcc5019d60f037ee9eb46d05093dd3023bdd688b9b207d10a39e054a5ed87db666b2cb75696f6537de74d1e1f8dcabc53dc8d2ab@16.63.123.160:30303";
+pub const DEFAULT_FLASHBLOCKS_SENTRIES: &str = "enode://78ca7daeb63956cbc3985853d5699a6404d976a2612575563f46876968fdca2383a195ee7db40de348757b2256195996933708f351169ca3f3fe93ab2a774608@16.62.98.53:30303,enode://c96dcadf4cdea4c39ec3fd775637d9e67d455b856b1514cfcf55b72f873a34b96d69e47ccea9fc797a446d4e6948aa80f6b9d479a1727ca166758a900b08f422@16.63.14.166:30303,enode://15688a7b281c32a4da633252dcc5019d60f037ee9eb46d05093dd3023bdd688b9b207d10a39e054a5ed87db666b2cb75696f6537de74d1e1f8dcabc53dc8d2ab@16.63.123.160:30303";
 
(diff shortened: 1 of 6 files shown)


crates/cli/src/cli.rs
```diff
@@ -341,6 +352,15 @@ impl WorldChainArgs {
...
         }
 
+        if let Some(flashblocks) = &self.flashblocks {
+            let bootnodes = config.network.bootnodes.get_or_insert_default();
+            for sentry in &flashblocks.sentry_peers {
+                if !bootnodes.iter().any(|bootnode| bootnode.id == sentry.id) {
+                    bootnodes.push(sentry.clone());
+                }
+            }
+        }
```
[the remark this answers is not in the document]
alessandromazza98: this comment seems relevant

alessandromazza98 (approved):
approving to not block you, but I don't fully understand what's the reason for having this change. Can you add a brief description in the PR descr?

Is it done to prevent a node to connect with too many nodes?

Note that the CI is red, you can fix it before merging it

crates/cli/src/cli/builder.rs
```diff
@@ -7,6 +7,9 @@ use std::path::PathBuf;
 
 use crate::cli::p2p::FanoutArgs;
 
+/// Default number of flashblocks sentries a client maintains as trusted RLPx peers.
+pub const DEFAULT_MAX_SENTRY_CONNECTIONS: usize = 1;
```
0xOsiris, line 14: Can you explain why this is a sensible default in the comment?

crates/cli/src/cli/builder.rs
```diff
@@ -160,6 +163,33 @@ pub struct FlashblocksArgs {
...
+    /// retained as a UDP discovery bootnode, including sentries that are not selected for RLPx.
+    #[arg(
+        long = "flashblocks.sentry-peers",
+        alias = "flashblocks.sentry_peers",
+        env = "FLASHBLOCKS_SENTRY_PEERS",
+        value_delimiter = ',',
+        value_name = "ENODE",
+        required = false
+    )]
+    pub sentry_peers: Vec<TrustedPeer>,
```
0xOsiris, line 183: I think this may get confusing. e.g. How do I as a user know what should be a bootnode / sentry-peer? Is there a difference?

crates/node/src/context.rs
```diff
@@ -87,8 +88,110 @@ impl WorldChainNetworkBuilder {
...
+                FLASHBLOCKS_SENTRY_SELECTION_DOMAIN.len() + local_peer_id.len() + sentry.id.len(),
+            );
+            input.extend_from_slice(FLASHBLOCKS_SENTRY_SELECTION_DOMAIN);
+            input.extend_from_slice(local_peer_id.as_slice());
+            input.extend_from_slice(sentry.id.as_slice());
+            (keccak256(input), sentry.id)
+        })
+        .collect::<Vec<_>>();
+
+    ranked.sort_unstable_by(|(left_score, left_id), (right_score, right_id)| {
```
0xOsiris, line 130: If the rank is purely random. What's the point of sorting in the first place. Why not just take `max_connections` capacity from the original vector without the persistent identifier.

crates/cli/src/cli/builder.rs
0xForerunner: It is a bit confusing. Currently they are the same, but we may eventually have different deployments for them.

crates/node/src/context.rs
0xForerunner: Ideally our selection is deterministic.

0xForerunner:
> approving to not block you, but I don't fully understand what's the reason for having this change. Can you add a brief description in the PR descr?
> 
> Is it done to prevent a node to connect with too many nodes?
> 
> Note that the CI is red, you can fix it before merging it

edited

PR #941, opened 2026-07-30
feat(proofs): submit proofs after game creation

Kemperino:
- move initial proof generation from proposer to defender
- require proof support for unchallenged success
- make canonical parent unique -> only allow ASR anchor if no compatible anchor of current game type exists
- let defender negatively resolve to enable propser to re-propose

> [!NOTE]
> **High Risk**
> Changes L1 dispute-game resolution, bond routing, and parent/canonical rules in MultiProofGame v2.0.0, plus coordinated proposer/defender/prover behavior; mistakes affect withdrawals, anchoring, and fund custody.
> 
> **Overview**
> Moves **proof submission to after** `DisputeGameFactory.create`: the proposer only opens games; the defender (and workers) supply proofs on-chain afterward.
> 
> **`MultiProofGame` v2.0.0** tightens resolution: an unchallenged proposal needs **at least one valid proof lane** before the challenge window can end in a defender win; a proposal with **no proof** times out and forfeits the proposer bond to **`proofTimeoutRecipient`**. Challenged games still need the configured multi-lane threshold. Adds **`canonicalAnchorParent()`** and stricter parent checks (**`NonCanonicalAnchorParent`**) so proposals chain off the right anchor parent across domain upgrades and retired games. **`submitProofLane`** is allowed while still *proposed* (deadline = challenge deadline) and after challenge (proof deadline).
> 
> **Offchain services** align with that model: **`world-chain-proposer`** drops prover-service / pre-create Nitro proofing and posts proposals directly; canonical scanning uses a single parent from **`canonicalAnchorParent()`**. **`world-chain-defender`** submits an initial **TEE** lane for valid proofless games, requests **SP1 + Nitro** only when challenged, ABI-encodes lane payloads (including **`public_key`** on Nitro), and **`resolve()`** negatively resolvable games (e.g. proof timeouts) so retries can proceed.
> 
> **Devnet / deploy / e2e**: defender always runs when the proof stack is enabled; SP1 worker is optional for escalation; mock Nitro proofs use **`TransitionPublicValues`**; withdrawal test waits for **`proofBitmap != 0`**; deploy script sets mock verifiers to accept proofs and configures proof timeout recipient.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbd9e5d6bfeb4d7628b2cb681b64b82345e00f4c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/contracts/src/proofs/MultiProofGame.sol
+++ b/pkg/contracts/src/proofs/MultiProofGame.sol
@@ -477,11 +477,10 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
             return (false, state(), ProofLib.InvalidationReason.NONE);
         }
 
-        // TODO: Decide whether reaching PROOF_THRESHOLD should make an unchallenged game
-        // immediately resolvable too. The current WIP guarantees the full challenge window even
-        // with threshold support, while challenged games resolve as soon as they reach the same
-        // threshold; update the WIP together with this branch if threshold support is intended
-        // to provide fast finality in both states.
+        if (ProofLib.hasThreshold(proofBitmap, PROOF_THRESHOLD)) {
+            return (true, ProofLib.RootState.FINALIZED, ProofLib.InvalidationReason.NONE);
+        }
+
         if (challenger == address(0)) {
             if (block.timestamp < challengeDeadline) {
                 return (false, ProofLib.RootState.PROPOSED, ProofLib.InvalidationReason.NONE);
@@ -491,9 +490,6 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
                 : (true, ProofLib.RootState.INVALIDATED, ProofLib.InvalidationReason.PROOF_TIMEOUT);
         }
 
-        if (ProofLib.hasThreshold(proofBitmap, PROOF_THRESHOLD)) {
-            return (true, ProofLib.RootState.FINALIZED, ProofLib.InvalidationReason.NONE);
-        }
         return block.timestamp >= proofDeadline
             ? (true, ProofLib.RootState.INVALIDATED, ProofLib.InvalidationReason.PROOF_TIMEOUT)
             : (false, ProofLib.RootState.CHALLENGED, ProofLib.InvalidationReason.NONE);
@@ -522,6 +518,10 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
         } else if (parentStatus == GameStatus.IN_PROGRESS) {
             // A proposed or challenged parent must resolve before its descendant.
             revert ParentGameNotResolved();
+        } else if (ProofLib.hasThreshold(proofBitmap, PROOF_THRESHOLD)) {
+            // Threshold support provides fast finality whether or not the game was challenged.
+            status = GameStatus.DEFENDER_WINS;
+            normalModeCredit[gameCreator()] = totalBonds;
         } else if (challenger == address(0) && proofBitmap != 0) {
             // Any configured proof lane may support the optimistic path. The offchain defender
             // uses TEE attestations by policy, but the protocol does not privilege a lane.
@@ -534,11 +534,6 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
             status = GameStatus.CHALLENGER_WINS;
             invalidationReason = ProofLib.InvalidationReason.PROOF_TIMEOUT;
             normalModeCredit[proofTimeoutRecipient] = totalBonds;
-        } else if (ProofLib.hasThreshold(proofBitmap, PROOF_THRESHOLD)) {
-            // A challenged game finalizes as soon as enough independent proof lanes support it.
-            // The proposer takes the challenger bond.
-            status = GameStatus.DEFENDER_WINS;
-            normalModeCredit[gameCreator()] = totalBonds;
         } else if (block.timestamp >= proofDeadline) {
             // A challenged game below threshold times out once its proof window expires. The
             // challenger takes the proposer bond.
--- a/pkg/contracts/src/proofs/MultiProofGame.sol
+++ b/pkg/contracts/src/proofs/MultiProofGame.sol
@@ -44,9 +44,10 @@ import {ISemver} from "@optimism-bedrock/interfaces/universal/ISemver.sol";
 /// @title MultiProofGame
 /// @notice A multi-proof dispute game created through the stock Optimism `DisputeGameFactory`
 ///         using the Clone-With-Immutable-Args (CWIA) pattern. Proposals chain parent-to-parent
-///         at a fixed block interval; a challenged proposal finalizes only once enough
-///         independent proof lanes (validity proof, TEE attestation, security council) support
-///         it. Bond custody uses `DelayedWETH` with the two-phase unlock/withdraw claim flow.
+///         at a fixed block interval. An unchallenged proposal requires one valid proof lane;
+///         a challenged proposal requires enough independent lanes (validity proof, TEE
+///         attestation, security council) to reach the configured threshold. Bond custody uses
+///         `DelayedWETH` with the two-phase unlock/withdraw claim flow.
 /// @dev Structure follows `ZKDisputeGame`; challenge/lane semantics are World Chain specific.
 contract MultiProofGame is Clone, ISemver, IMultiProofGame {
     ////////////////////////////////////////////////////////////////
@@ -69,6 +70,7 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
     uint64 public immutable proofPeriod;
     uint256 public immutable proposerBond;
     uint256 public immutable challengerBond;
+    address public immutable proofTimeoutRecipient;
 
     IWorldChainProofVerifier public immutable validityProofVerifier;
     IWorldChainProofVerifier public immutable teeVerifier;
@@ -83,8 +85,8 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
     ////////////////////////////////////////////////////////////////
 
     /// @notice Semantic version.
-    /// @custom:semver 1.0.0
-    string public constant version = "1.0.0";
+    /// @custom:semver 2.0.0
+    string public constant version = "2.0.0";
 
     Timestamp public createdAt;
     Timestamp public resolvedAt;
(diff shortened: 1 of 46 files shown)


proofs/prover-service/src/types.rs
```diff
@@ -200,6 +200,8 @@ pub enum ProofData {
         public_values: Bytes,
         /// Enclave signature over the proven outputs.
         signature: Bytes,
+        /// SEC1-uncompressed public key certified by the attestation document.
+        public_key: Bytes,
```
Kemperino, line 204: will remove this public key dependency in a follow up PR by changing the TEE verifier

pkg/contracts/src/proofs/MultiProofGame.sol
```diff
@@ -219,6 +222,28 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
...
+        ) {
+            return address(anchorStateRegistry);
+        }
+
+        // A previous implementation registered under this game type may predate `domainHash`.
+        try IMultiProofGame(address(anchorGame_)).domainHash() returns (bytes32 anchorDomainHash) {
+            if (anchorDomainHash == domainHash) return address(anchorGame_);
+        } catch {}
+
+        return address(anchorStateRegistry);
```
Kemperino: todo: check this

pkg/contracts/src/proofs/MultiProofGame.sol
```diff
@@ -219,6 +222,28 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
...
+    /// @inheritdoc IMultiProofGame
+    function canonicalAnchorParent() public view returns (address parent_) {
+        IDisputeGame anchorGame_ = anchorStateRegistry.anchorGame();
+        if (address(anchorGame_) == address(0)) return address(anchorStateRegistry);
+        if (anchorGame_.gameType().raw() != GameTypes.MULTI_PROOF_GAME_TYPE.raw()) {
+            return address(anchorStateRegistry);
+        }
+        if (
+            anchorGame_.status() == GameStatus.CHALLENGER_WINS || !anchorGame_.wasRespectedGameTypeWhenCreated()
+                || anchorStateRegistry.isGameBlacklisted(anchorGame_) || anchorStateRegistry.isGameRetired(anchorGame_)
```
alessandromazza98: this condition should not ever be true. The ASR already ensures these conditions when it sets a new game as the anchor

pkg/contracts/src/proofs/MultiProofGame.sol
```diff
@@ -305,9 +330,12 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
 
         (Hash anchorRoot, uint256 anchorL2BlockNumber) = anchorStateRegistry.getAnchorRoot();
         address parentRef_ = parentRef();
+        address canonicalAnchorParent_ = canonicalAnchorParent();
```
alessandromazza98, line 311: with this implementation we're not solving the problem of having a unique `parentRef` though. It can be both the real parent game address or the anchor state registry and this creates 2 possible lines which then creates problem for the proposer.

alessandromazza98:
I'd separate the PR with contracts changes to the offchain services to make it a bit smaller to review

EDIT: we agreed on making a unique PR for this work

proofs/prover-service/src/store.rs
```diff
@@ -881,24 +881,19 @@ fn proof_id_bytes(id: ProofRequestId) -> Vec<u8> {
...
-/// backend session from the previous attempt unusable while still eligible for resumption.
+/// The L1 head is immutable game metadata. A mismatch must create a different request rather
+/// than resuming a backend session whose proof cannot satisfy the game verifier.
 fn request_matches(row: &PgRow, request: &ProofRequest) -> Result<bool, ProofRequestError> {
     let stored_backend: &str = row.get("backend");
     let stored_game: &[u8] = row.get("game");
     let stored_root_claim: &[u8] = row.get("root_claim");
     let stored_l2_block_number: i64 = row.get("l2_block_number");
     let stored_l1_head = b256_from_bytes(row.get("l1_head"))?;
-    let l1_head_matches =
```
alessandromazza98, line 894: nice🤝

proofs/prover-service/src/types.rs
alessandromazza98: sounds good to me

proofs/defender/src/defender.rs
```diff
@@ -364,6 +427,11 @@ where
                 continue;
             }
 
+            // TODO: The proposer adopts an existing canonical transition regardless of who
+            // created it, so an exact-UUID front-run is followed but currently receives no proof
+            // support here. Reconcile the policies by defending the game selected by the
+            // proposer's canonical-line rules; accepting every foreign game with a valid root
+            // would let parallel branches amplify proof-generation work.
```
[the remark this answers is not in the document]
alessandromazza98: we discussed this and yes we need to adjust our defender to defend all (with valid root claim) games

proofs/defender/src/lane.rs
```diff
@@ -226,30 +238,40 @@ fn proof_request(game: &GameMetadata, backend: ProofBackend) -> ProofRequest {
...
-            public_values.as_ref(),
-            attestation.as_ref(),
-            signature.as_ref(),
-        ]
-        .concat()
-        .into(),
+            public_key,
+        } => {
+            let transition = <TransitionPublicValues as SolValue>::abi_decode(public_values)
+                .map_err(|error| DefenderError::ProofEncoding(error.to_string()))?;
```
alessandromazza98, line 267: why do we decode and then encode (later in abi_encode)? cannot we just use public_values as we do for sp1?

alessandromazza98 (approved):
I like it

proofs/defender/src/lane.rs
alessandromazza98: you can do:

```rust
/// Encodes the verifier-specific payload passed to `submitProofLane`.
fn encode_proof(metadata: &GameMetadata, proof: &ProofData) -> Result<Bytes, DefenderError> {
    let l1_origin_number = U256::from(metadata.l1_origin_number);
    match proof {
        ProofData::Sp1 {
            proof,
            public_values,
        } => Ok((
            metadata.domain_hash,
            metadata.parent_ref,
            l1_origin_number,
            public_values.clone(),
            proof.clone(),
        )
            .abi_encode()
            .into()),
        ProofData::Nitro {
            attestation: _,
            public_values,
            signature,
            public_key,
        } => {
            Ok((
                metadata.domain_hash,
                metadata.parent_ref,
                l1_origin_number,
                public_values.clone(),
                signature.clone(),
                public_key.clone(),
            )
                .abi_encode()
                .into())
        }
    }
}
```

and then you can even remove the `Result` wrapper and directly return the `Bytes`

proofs/defender/src/defender.rs
```diff
@@ -239,34 +244,44 @@ where
...
                     return Ok(DefenseProgress::DeadlineElapsed);
                 }
                 return Ok(DefenseProgress::Closed);
             }
+            GameObservation::Proposed {
+                proof_bitmap,
+                has_initial_support,
+            } => {
+                if has_initial_support {
+                    return Ok(DefenseProgress::Complete);
```
alessandromazza98, line 175: an important point here: this is correct. But this game can later be challenged and another proof would be necessary. I think our defender already "sort of" handles this by scanning past `game_scan_lookback` (defaults to 100) games every tick and therefore it should be able to handle these scenarios.

In fact 100 games in a normal condition includes more than 4 days of games. So I think we can say it's already handled, I just wanted to point this out loud

proofs/defender/src/defender.rs
```diff
@@ -341,6 +357,54 @@ where
         self.handle_defense_progress(defense_results);
     }
 
+    async fn resolve_negative_games(&self) {
```
alessandromazza98: isn't this already handled by our `challenger`? why should the defender do it too?

alessandromazza98: the only difference I can see is that our `challenger` only negative resolve for owned challenge games, while this defender negative resolve all games. So the difference is that with this defender we negative resolve all games, even not owned games, therefore accounting money to the protocol account for example if a game has been created without proof attached

alessandromazza98: further, I think we will need some service that acts as bond manager for the protocol account as well (that withdraws money from games with no proof attached)

pkg/contracts/src/proofs/MultiProofGame.sol
Kemperino: fixed in https://github.com/worldcoin/world-chain/pull/943

pkg/contracts/src/proofs/MultiProofGame.sol
Kemperino: https://github.com/worldcoin/world-chain/pull/943

proofs/prover-service/src/types.rs
Kemperino: https://github.com/worldcoin/world-chain/pull/944

proofs/defender/src/defender.rs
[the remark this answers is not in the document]
Kemperino: this is happening already with this PR now

proofs/proposer/src/proposer.rs
```diff
@@ -51,241 +32,174 @@ impl<E, C, P> WorldChainProposer<E, C, P> {
...
         let (proposal, retry_of) = match scan.next_action() {
             NextProposalAction::Propose(proposal) => (*proposal, None),
             NextProposalAction::RetryTimedOut {
                 proposal,
                 invalidated_game,
-            } => (*proposal, Some(*invalidated_game)),
-            NextProposalAction::AwaitNegativeResolution { game, reason } => {
-                warn!(
+            } => {
+                error!(
```
alessandromazza98, line 185: nit: I'd put a warn! here instead of a error!. The idea is that this is still correct and valid - we may need to eventually resolve some game manually

proofs/defender/src/defender.rs
```diff
@@ -120,116 +95,61 @@ where
...
+                .game_metadata(selected.game.address)
+                .await?;
+            if metadata.parent_ref != selected.transition.parent_ref
+                || metadata.root_claim != selected.transition.root_claim
+                || metadata.l2_block_number != selected.transition.l2_block_number
+            {
+                return Err(DefenderError::Contract(format!(
+                    "selected game {} does not match its transition commitment",
+                    selected.game.address
+                )));
```
alessandromazza98, line 115: what is this check for? do we ever hit this error? I don't think so because the lineage.games() are read through the execution_provider and consensus_provider just above, so it should always check. I think you can get rid of it

alessandromazza98 (approved):
it looks good to me. I've done a pretty deep review and I think we can merge and work on smaller improvements

proofs/defender/src/lane.rs
Kemperino: /otto first create a ticket for this, then check if we can actually do this?

proofs/defender/src/defender.rs
Kemperino: i agree!

PR #967, opened 2026-08-04
fix(txpool): reject replayed PBH transactions before validation

alessandromazza98:
### Problem

Replayed PBH transactions could trigger expensive proof verification before duplicate nullifiers or already-imported transaction hashes were rejected, enabling unnecessary CPU consumption.

### Solution

Check for duplicate nullifiers before PBH proof verification and reject transaction hashes already present in the pool before invoking pool validation. Retain the pool’s insertion-time duplicate check to handle concurrent submissions.

---

> [!NOTE]
> **Medium Risk**
> Changes ordering and early-exit paths in PBH proof validation and eth raw-tx submission; behavior for duplicates should match prior outcomes but timing and error surfacing on RPC replays differ slightly from running full validation.
> 
> **Overview**
> Hardens PBH bundle handling and raw-tx RPC against **replay-style CPU waste**.
> 
> **PBH validator:** Payloads are decoded sequentially, then **duplicate nullifier hashes** are rejected **before** parallel cryptographic proof verification. That stops a bundle from repeating the same `(UserOp, payload)` pair to multiply expensive proof work. A test now uses an invalid proof to assert duplicate detection runs ahead of proof validation.
> 
> **`send_raw_transaction`:** After recovering the tx, if the hash is **already in the pool**, the handler returns `AlreadyImported` immediately instead of running full pool validation again. Concurrent duplicates still rely on the pool’s insertion-time check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b51df4e0f1094298b5c68f15ac4df7fc9b564bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/pool/src/validator.rs
+++ b/crates/pool/src/validator.rs
@@ -8,7 +8,7 @@ use crate::{
 use alloy_eips::BlockId;
 use alloy_primitives::Address;
 use alloy_sol_types::{SolCall, SolValue};
-use rayon::iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator};
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use reth_evm::ConfigureEvm;
 use reth_optimism_forks::OpHardforks;
 use reth_optimism_node::txpool::OpTransactionValidator;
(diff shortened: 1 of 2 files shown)


crates/pool/src/validator.rs
```diff
@@ -185,14 +185,40 @@ where
...
+                .into_iter()
                 .zip(aggregated_ops.userOps)
+                .map(|(payload, op)| {
+                    PbhPayload::try_from(payload)
+                        .map(|payload| (payload, op))
+                        .map_err(|_| {
+                            WorldChainPoolTransactionError::from(
+                                PBHValidationError::InvalidCalldata,
+                            )
+                        })
```
Dzejkop: could we do something like

```suggestion
                    let payload = PbhPayload::try_from(payload)?; // <- will likely need some `.map_error`
                    Ok((payload, op))
```

alessandromazza98: yes, doing it now

alessandromazza98: done

Dzejkop (approved):

PR #992, opened 2026-08-05
chore: reth 2.4.1

0xForerunner:
> [!NOTE]
> **Medium Risk**
> Touches execution, payload assembly, and state-root/trie wiring across flashblock build and validate paths; regressions could affect pending block publication or state consistency even though behavior is largely API adaptation.
> 
> **Overview**
> Aligns World Chain with **op-reth v2.4.1** (and matching **reth** git pins / lockfile), including **op-alloy** and related Optimism crates on the same tag.
> 
> Application code is updated for upstream API shifts around built payloads and chain state: **`BuiltPayloadExecutedBlock`** now carries **`changed_paths`**, and flashblock **builder** / **validator** paths construct that struct when attaching executed data to **`OpBuiltPayload`**. **`into_executed_payload`** converts **`BuiltPayloadExecutedBlock`** into **`ExecutedBlock`** by sorting hashed state and trie updates in parallel, then building **`ComputedTrieData`** via **`new_with_changed_paths`** for **`ExecutedBlock::new`** (replacing the older separate trie fields). Transaction recovery in the validator follows the **`alloy_consensus::transaction`** **`SignerRecoverable`** / **`Recovered`** layout expected by the new stack.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1fb1d980e4970bb3875419c0d45c4100216fb306. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/build.rs
+++ b/crates/node/build.rs
@@ -1,7 +1,7 @@
 use std::{env, error::Error};
 
-use vergen::{BuildBuilder, CargoBuilder, Emitter};
-use vergen_git2::Git2Builder;
+use vergen::{Build, Cargo, Emitter};
+use vergen_git2::Git2;
 
 fn main() -> Result<(), Box<dyn Error>> {
     println!("cargo:rerun-if-env-changed=VERGEN_GIT_SHA");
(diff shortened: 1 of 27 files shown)


crates/node/build.rs
```diff
@@ -11,20 +11,15 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    emitter.add_instructions(&Cargo::builder().features(true).target_triple(true).build())?;
     if supplied_sha.is_none() {
         emitter.add_instructions(
-            &Git2Builder::default()
+            &Git2::builder()
                 .describe(false, true, None)
                 .dirty(true)
-                .sha(true)
-                .build()?,
+                .sha(false)
```
0xOsiris, line 21: Why false?

0xOsiris, line 21:
```suggestion
                .sha(true)
```

0xForerunner: I don't think this makes a difference

0xForerunner: false just means it uses the short version, but that's not actually what we want.

crates/validator/src/validator.rs
```diff
@@ -38,7 +39,23 @@ use world_chain_evm::{
 pub fn into_executed_payload(
     payload: BuiltPayloadExecutedBlock<OpPrimitives>,
 ) -> ExecutedBlock<OpPrimitives> {
-    let trie_data = DeferredTrieData::sort(payload.hashed_state, payload.trie_updates);
+    let hashed_state = payload.hashed_state;
```
0xOsiris:
```suggestion
    let _span = debug_span!(target: "flashblocks::validator", "sort_trie_inputs").entered();
    let hashed_state = payload.hashed_state;
```

0xOsiris (approved):

PR #1013, opened 2026-08-07
fix: validate MultiProofGame 

Kemperino:
> [!NOTE]
> **Medium Risk**
> Changes dispute-game creation invariants on a security-sensitive path, but the diff is narrow and covered by a new regression test.
> 
> **Overview**
> Tightens **parent validation** on `MultiProofGame` creation so using the anchor registry as `parentRef` is only allowed while there is **no** `anchorGame` yet. After the anchor advances, new games must parent off a **factory-registered, respected, live** dispute game via `_isValidParent`.
> 
> Also **deduplicates bond custody** in `_depositBond` (proposer init + challenger), documents **lazy `closeGame()`** inside `claimCredit`, and adds **`test_Create_RejectsAnchorSentinelAfterAnchorAdvances`** to reject late bootstrap-style parents.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0098797ea922030bc41a385d7d4d34140371733. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/contracts/src/dispute/MultiProofGame.sol
+++ b/pkg/contracts/src/dispute/MultiProofGame.sol
@@ -338,10 +338,9 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
         Claim rootClaim_ = rootClaim();
         uint256 attempt_ = attempt();
 
-        // INVARIANT: The parent must be a registered, respected, live game of this type.
-        if (parentRef_ != address(anchorStateRegistry) && !_isValidGame(IDisputeGame(parentRef_))) {
-            revert InvalidParentGame();
-        }
+        // INVARIANT: The registry sentinel is only valid while bootstrapping from the starting
+        // anchor. Once an anchor game exists, the parent must be an eligible game.
+        if (!_isValidParent(parentRef_)) revert InvalidParentGame();
 
         // INVARIANT: Each proposal must advance its parent by the configured cadence.
         (, uint256 startingBlockNumber_) = startingProposal();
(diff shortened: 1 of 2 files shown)


pkg/contracts/src/dispute/MultiProofGame.sol
```diff
@@ -407,13 +403,27 @@ contract MultiProofGame is Clone, ISemver, IMultiProofGame {
...
-    /// @notice Checks if the game is registered, respected, not blacklisted, not retired, and not challenged.
+    /// @notice Checks if the game is factory-registered, was respected when created, and has not
+    ///         been invalidated by blacklisting, retirement, or a challenger win.
     function _isValidGame(IDisputeGame game) internal view returns (bool) {
         return anchorStateRegistry.isGameRegistered(game) && anchorStateRegistry.isGameRespected(game)
             && !anchorStateRegistry.isGameBlacklisted(game) && !anchorStateRegistry.isGameRetired(game)
             && (game.status() != GameStatus.CHALLENGER_WINS);
     }
 
+    function _isValidParent(address parentRef_) internal view returns (bool) {
```
Kemperino, line 414: @alessandromazza98 we had this check before basically to ensure we are able to point to ASR for very first game.

I am not sure that would be possible with current code.

alessandromazza98: what do you mean? I don't understand.

This seems correct to me, but even the previous version was correct

alessandromazza98 (approved):

PR #1026, opened 2026-08-10
feat(nitro-proof-worker): add aws kms support

alessandromazza98:
- add aws kms support to `nitro-proof-worker`

---

> [!NOTE]
> **Medium Risk**
> Changes how L1 registration transactions are signed in production (KMS IAM and key configuration); mistakes could block registration or use the wrong funding account, but it does not alter on-chain attestation or enclave proof-signing logic.
> 
> **Overview**
> Adds **AWS KMS** as an alternative to a local hex key for the **host** account that pays gas and signs `NitroEnclaveKeyRegistry.registerKey` (unchanged on-chain semantics; still separate from enclave proof-signing keys).
> 
> Registration flows now take `signer_secret` plus `SignerType` (`Local` or `AwsKms`) and build the L1 wallet via `world_chain_proof_tx_signer::build_transaction_signer`, replacing the previous local-only `PrivateKeySigner` path in `register_enclave_key`.
> 
> **CLI and ops:** `world-chain-prover-nitro register`, `nitro-worker register`, and `nitro-worker run --auto-register` accept `--kms-key-id` / `REGISTER_KMS_KEY_ID`, mutually exclusive with `REGISTER_PRIVATE_KEY`. `PRIVATE_KEY` remains a fallback only when neither dedicated registration signer is set (so KMS can win over a leftover `PRIVATE_KEY`). `just proof-register-key` mirrors this and unsets both signer env vars inside `kubectl exec` so Deployment auto-register config cannot leak a second signer.
> 
> Docs describe the KMS key type (`ECC_SECG_P256K1`) and required IAM actions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d3f62a372836b7154a499ed9eb6d2f85004ac9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -642,7 +642,8 @@ proof-verify-pcrs env="alphanet":
 #            it to NitroEnclaveKeyRegistry. Idempotent: a no-op if already registered.
 #            `registerKey` is NOT owner-gated, so any funded key works.
 #
-# Required: L1_RPC_URL, and a funding key via REGISTER_PRIVATE_KEY or PRIVATE_KEY.
+# Required: L1_RPC_URL, and a funding signer via REGISTER_PRIVATE_KEY,
+#           REGISTER_KMS_KEY_ID, or the legacy PRIVATE_KEY fallback.
 # Optional: NITRO_ENCLAVE_KEY_REGISTRY (else read from the {{env}}-nitro.json deployment),
 #           PCR0/PCR1/PCR2 (else host-side attestation checks are skipped; the on-chain
 #           verifier still enforces the approved PCR allowlist).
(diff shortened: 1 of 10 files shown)


Justfile
```diff
@@ -700,7 +711,11 @@ proof-register-key env="alphanet":
         printf 'export ENCLAVE_CID=%s\n' "$(shq "$ENCLAVE_CID")"
         printf 'export NITRO_ENCLAVE_KEY_REGISTRY=%s\n' "$(shq "$NITRO_ENCLAVE_KEY_REGISTRY")"
         printf 'export L1_RPC_URL=%s\n' "$(shq "$L1_RPC_URL")"
-        printf 'export REGISTER_PRIVATE_KEY=%s\n' "$(shq "$REGISTER_KEY")"
+        if [ -n "${REGISTER_KMS_KEY_ID:-}" ]; then
+            printf 'export REGISTER_KMS_KEY_ID=%s\n' "$(shq "$REGISTER_KMS_KEY_ID")"
+        else
+            printf 'export REGISTER_PRIVATE_KEY=%s\n' "$(shq "$REGISTER_KEY")"
+        fi
```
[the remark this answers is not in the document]
alessandromazza98: fisxed

proofs/backends/nitro/enclave/src/register.rs
```diff
@@ -312,9 +314,20 @@ pub struct RegisterParams {
     pub l1_rpc_url: String,
     /// `NitroEnclaveKeyRegistry` contract address on L1 (hex, `0x`-prefixed).
     pub registry: String,
-    /// Hex-encoded private key used to sign (and pay gas for) the `registerKey` tx.
-    /// `registerKey` is **not** owner-gated, so any funded key works.
-    pub private_key: String,
+    /// Local private key or AWS KMS key ID used to sign (and pay gas for) the
+    /// `registerKey` tx. `registerKey` is **not** owner-gated, so any funded key works.
+    pub signer_secret: String,
```
0xOsiris, line 319: nit: Can we just make this a single enum field for simplicity

e.g.

```rust 
pub enum Signer {
     /// Hex-encoded local secp256k1 private key.
    Local(String),
    /// AWS KMS key ID or alias.
    AwsKms(String),
}


impl Signer {
    pub fn into_signer(self, url: &str) -> Result<...> { ... }
}
```

alessandromazza98: we can, but this would be slightly different to what we currently do for all the other services. I think we can do what you propose in a follow up and apply it to all services at the same time

cichaczem (approved):

PR #1037, opened 2026-08-12
fix(proofs): reproducibly build SP1 worker ELFs

Kemperino:
> [!NOTE]
> **High Risk**
> Changes how production SP1 guest ELFs and on-chain vkeys are built and verified—governance-critical proof measurements. A build or pin mistake could ship a worker whose embedded ELFs diverge from committed `vkeys.json`.
> 
> **Overview**
> Makes the production **SP1 worker** embed guest ELFs from a **digest-pinned** SP1 image and canonical `/root/program` layout, so on-chain vkeys no longer drift from host-path-dependent local builds.
> 
> `Dockerfile.prover` adds a dedicated `sp1-worker` target: guests are built once in `sp1-guests`, copied into the host builder with `SP1_SKIP_PROGRAM_BUILD=true`, then checked against `vkeys.json` both as intermediate ELF hashes and via `world-chain-proof-sp1-worker vkeys --check` on the linked binary. The generic `final` target now refuses to build the SP1 worker.
> 
> Also pins the SP1 image by digest in `build.rs`, adds shared `embedded_vkey_manifest` helpers plus the worker `vkeys` subcommand, and wires CI/`docker-proof.yml` to the new target (with QEMU on arm64 for the amd64 guest compiler).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fc37a17b5076f259dca846267494e3bd52528fb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/docker-proof.yml
```diff
@@ -88,12 +85,13 @@ jobs:
       - name: Build and push digest
         uses: ./.github/actions/docker-build-push-digest
         with:
-          platform: ${{ matrix.platform.platform }}
+          platform: linux/amd64
```
0xOsiris: Can we keep it cross platform?

0xOsiris: Most of these do not need to be single platform reproducible. We should isolate out the ELF into it's own workflow if it does

0xOsiris (changes requested):

0xOsiris (approved):

PR #1049, opened 2026-08-17
refactor(nitro): delegate cert path validation to webpki

0xOsiris:
---

> [!NOTE]
> **Medium Risk**
> Touches security-critical attestation verification with deliberate acceptance changes (reordered intermediates); overall hardening via audited path validation and mandatory leaf certificates.
> 
> **Overview**
> Replaces the hand-rolled RFC 5280 certificate walk in Nitro attestation verification with **`rustls-webpki`**, anchored on the pinned AWS Nitro root. Signature linkage, validity, name chaining, basic constraints, and `pathLenConstraint` now come from that library; **`x509-parser` stays only for AWS-specific `keyUsage` checks** (`keyCertSign` on CAs, `digitalSignature` on the leaf).
> 
> **`verify_cose_sign1_signature`** no longer skips crypto when the `certificate` field is missing—those documents are rejected instead of letting PCR/`user_data` checks pass without hardware proof.
> 
> **Intentional behavior changes:** intermediate order in `cabundle` is no longer required to be strict (webpki builds the path); reordered real AWS intermediates now verify. Tests were updated for webpki error variants and the path-length case now constrains an intermediate, not the trust anchor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d15e86f277c856f0a11dfd634fb44e68f703f3d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proofs/backends/nitro/enclave/src/attestation.rs
+++ b/proofs/backends/nitro/enclave/src/attestation.rs
@@ -522,6 +522,27 @@ fn verify_root_ca_against(
     Ok(())
 }
 
+/// Checks the pinned root's own validity window.
+fn check_root_validity(root_der: &[u8], now_secs: i64) -> Result<(), AttestationError> {
+    use x509_parser::prelude::{ASN1Time, FromDer as _, X509Certificate};
+
+    let (_, root) = X509Certificate::from_der(root_der)
+        .map_err(|e| AttestationError::CertChain(format!("pinned root CA parse: {e}")))?;
+
+    let now = ASN1Time::from_timestamp(now_secs).map_err(|e| {
+        AttestationError::CertExpired(format!("current time is unrepresentable: {e}"))
+    })?;
+
+    if !root.validity().is_valid_at(now) {
+        return Err(AttestationError::CertExpired(format!(
+            "pinned root CA outside its validity window (notBefore {}, notAfter {}, now {now_secs})",
+            root.validity().not_before.timestamp(),
+            root.validity().not_after.timestamp(),
+        )));
+    }
+    Ok(())
+}
+
 /// AWS signs every Nitro certificate with ecdsa-with-SHA384; naming only that pins the chain
 /// away from weaker primitives.
 static NITRO_SIG_ALGS: &[&dyn SignatureVerificationAlgorithm] = &[webpki::ring::ECDSA_P384_SHA384];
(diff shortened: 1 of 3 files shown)


proofs/backends/nitro/enclave/src/attestation.rs
```diff
@@ -599,95 +576,66 @@ fn verify_cert_chain_against(
...
+    // Shape check only: the anchor below comes from `expected_root`, never from the document.
     verify_root_ca_against(cabundle, expected_root)?;
 
-    // Every element of cabundle is a CA in the path, root included: its own pathLenConstraint
-    // bounds what may follow it. Only the signature and issuer checks are skipped for the root,
-    // which is self-signed and anchored above.
-    for i in 0..cabundle.len() {
-        let (_, cert) = X509Certificate::from_der(&cabundle[i])
+    let root = CertificateDer::from(expected_root);
+    let anchor = webpki::anchor_from_trusted_cert(&root)
```
0xForerunner, line 605: Do we not need to check root validity first?

0xForerunner (changes requested):

proofs/backends/nitro/enclave/src/attestation.rs
0xOsiris: WDYM `root` here is the hardcoded `AWS_NITRO_ROOT_CA_PEM`

0xOsiris: `cabundle[0]` is checked against the `AWS_NITRO_ROOT_CA_PEM` in `verify_root_ca_against`

piohei (approved):

Kemperino (approved):

proofs/backends/nitro/enclave/src/attestation.rs
0xForerunner: Right, `verify_root_ca_against` checks that it’s the pinned root. I mean its validity window. `anchor_from_trusted_cert` treats it as a prevalidated trust anchor, so `verify_for_usage` doesn’t check the root’s notBefore/notAfter. So I'm wondering if we need to retain an explicit validity check for expected_root?

0xForerunner: Although now I see that the pinned root expires in 2049, so maybe not worrying about haha.

0xForerunner (approved):

PR #1059, opened 2026-08-19
feat (proofs): automatic cancel + retries on auction timeout of SP1 proofs

Kemperino:
- resubmits timed out and auction timed out proofs with back off (1,2,5 mins if provers are unavailable)
- uses wait_proof for polling instead of our own looping around get (which didnt detect proof / auction time out)
- removes poll() and donwload() from `WorldSuccinctProver`

> [!NOTE]
> **Medium Risk**
> Changes affect live SP1 Network job handling and retry semantics for proof generation; misconfigured timeouts or resubmission limits could delay or duplicate network requests, though scope is limited to the SP1 worker path.
> 
> **Overview**
> **SP1 Network proving** now uses the SDK’s `wait_proof` path (with optional auction and proof deadlines) instead of a manual poll loop, and maps auction/proof timeouts and other terminal network failures into `SuccinctProverError` so callers can tell resubmittable timeouts from hard failures.
> 
> The **SP1 worker** records failed sessions, then automatically resubmits range or aggregation requests after **60s / 120s / 300s** delays, up to **three** resubmissions per worker attempt; a finished range proof is kept when only aggregation needs a retry. Worker CLI/env adds `SP1_AUCTION_TIMEOUT_SECONDS` and `SP1_PROOF_TIMEOUT_SECONDS`, folded into `NetworkProofRequestConfig`.
> 
> **Docs** describe the new timeouts and retry behavior; **prover-service** tests cover replacing a failed proof session with a new running one.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d895258626281c3f24667fb8027313271c8228c6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proofs/workers/sp1/src/backend.rs
+++ b/proofs/workers/sp1/src/backend.rs
@@ -311,10 +311,11 @@ impl<P: WorldSuccinctProver + Send + Sync, G: ProofGameProvider> Sp1Backend<P, G
         request: RangeProofRequest,
         retrying_existing: bool,
     ) -> anyhow::Result<RangeProofArtifact> {
-        let mut resubmission = usize::from(retrying_existing);
-        loop {
-            self.wait_before_resubmission(SessionType::Stark, resubmission)
-                .await;
+        for (resubmission, delay) in network_request_attempts(retrying_existing) {
+            if let Some(delay) = delay {
+                self.wait_before_resubmission(SessionType::Stark, resubmission, delay)
+                    .await;
+            }
             let session_id = self
                 .submit_session(
                     job,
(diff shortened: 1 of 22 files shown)


proofs/workers/sp1/src/backend.rs
```diff
@@ -196,112 +200,171 @@ impl<P: WorldSuccinctProver, G: ProofGameProvider> Sp1Backend<P, G> {
...
             }
+            resubmission = next_resubmission(SessionType::Snark, resubmission)?;
+        }
+    }
+
+    async fn wait_before_resubmission(&self, session_type: SessionType, resubmission: usize) {
+        if resubmission == 0 {
+            return;
         }
+        let delay = NETWORK_REQUEST_RETRY_BACKOFFS[resubmission - 1];
```
piohei: I know the array range is checked in other function but it is easy to change something in the future and by accident errored here. What do you think about using `find` over the `NETWORK_REQUEST_RETRY_BACKOFFS` instead of infinite loop with return inside?

piohei (approved):

PR #1062, opened 2026-08-20
fix(proposer): settle obsolete-domain games

Kemperino:
makes the bond manager "upgrade-safe" in the sense that it would even resolve and collect bonds from games which were, but are no longer in the current lineage due to a change in e.g. domain hash / a fork

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 7ec4713b7cde62908472bccf6d239532c3aa4c98. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proofs/services/proposer/src/bond_manager.rs
+++ b/proofs/services/proposer/src/bond_manager.rs
@@ -5,7 +5,7 @@ use tracing::{info, warn};
 
 use crate::{BondManagerClient, BondManagerConfig, ProposerError};
 
-/// Discovers proposer-owned games, resolves invalid-parent descendants, and recovers bond credits.
+/// Discovers proposer-owned games, resolves abandoned games, and recovers bond credits.
 #[derive(Debug)]
 pub struct BondManager<E> {
     config: BondManagerConfig,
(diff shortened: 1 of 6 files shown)


proofs/services/proposer/src/bond_manager.rs
```diff
@@ -82,30 +82,37 @@ where
...
-                    if resolution_status.invalid_parent_resolvable() {
+                    // Resolve games abandoned by retries or domain upgrades; active-domain direct
+                    // outcomes stay with the proposer to avoid racing retry creation.
+                    let obsolete_domain_resolvable = if resolution_status.resolvable {
+                        self.execution_provider.game_domain_hash(game).await? != active_domain_hash
+                    } else {
+                        false
+                    };
+                    if resolution_status.invalid_parent_resolvable() || obsolete_domain_resolvable {
                         let submission = self.execution_provider.resolve_game(game).await?;
```
0xOsiris, line 107: I'm not sure I quite understand the intuition. Why do we not want to resolve a game whose domain hash is historical. Like what would be the side effect of this?

0xOsiris (approved):

proofs/services/proposer/src/bond_manager.rs
Kemperino: this is only resolving if for example we change the game implementation we still want to collect the proposer bond from a previously proposed game of "older lineage"

PR #1063, opened 2026-08-20
feat(proofs): route jobs by verifier values

Kemperino:
add the vkeys and pcr0 hash as parameters on the proof table, safe it per game and advertise it accordingly on the proof workers.

this is needed to still defend old games on cutover when a new game is deployed.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 37c5068c624c785a61e03477c83189836a218915. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


proofs/services/prover-service/migrations/20260619_proof_requests.sql
```diff
@@ -28,9 +30,9 @@ CREATE TABLE proof_requests (
    
 );
 
-CREATE INDEX proof_requests_queued_idx
-    ON proof_requests (backend, created_at)
-    WHERE job_status = 'PENDING';
+CREATE INDEX proof_requests_verifier_queue_idx
+    ON proof_requests (backend, verifier_id, range_vkey_commitment, l2_block_number, created_at)
+    WHERE job_status IN ('PENDING', 'CLAIMED');
```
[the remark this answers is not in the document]
Kemperino: its pre v0, we will reset :)

proofs/workers/nitro/src/lib.rs
```diff
@@ -72,6 +72,10 @@ where
         ProofBackend::Nitro
     }
 
+    fn verifier_id(&self) -> B256 {
+        keccak256(self.config.expected_pcrs.pcr0)
```
0xOsiris, line 76: The PCR's should not be supplied over CLI. It should be fetched via the attestation document of the enclave it's running

0xOsiris (approved):

proofs/workers/nitro/src/lib.rs
Kemperino: yes I agree. it's already doing the roundtrip at startup, will fetch and verify there.

PR #1064, opened 2026-08-20
feat(provers): release flow

0xOsiris:
- Isolates Nitro Enclave outside of Cargo workspace
- Adds `proof-releases.lock` which tracks all enclave EIF or Sp1 Program changes to a specific release (rc or stable)
- Adds Gates to CI that ensure `releases.lock` is up to date, and well formed

https://linear.app/worldcoin/issue/PROTO-5039/proof-releases

> [!NOTE]
> **High Risk**
> Touches proof-system governance (SP1 vkeys, Nitro PCRs, tee_image_id), GHCR publishing, and tag/release automation with contents/packages write. A gate or registry bug could block merges or ship mismatched on-chain measurements.
> 
> **Overview**
> Adds **`proof-releases.lock`** as the source of truth for SP1 vkeys, ELF hashes, enclave PCRs, and `tee_image_id`. PRs that change those measurements cannot merge without a new append-only entry (`just proof-release --version X.Y.Z`).
> 
> **CI gates:** `vkeys.yml` rebuilds SP1 measurements against the current registry entry; new `proof-gate.yml` validates the lockfile (append-only, `tee_image_id = keccak256(pcr0)`) and rebuilds the EIF when enclave inputs change. On merge, `proof-release-tag.yml` creates `proofs/vX.Y.Z` and dispatches `release-proof.yml`, which re-verifies measurements, reuses `docker-proof.yml` (now `workflow_call`) to push `proofs-v*` images, and opens a **draft** GitHub release with `manifest.json` + EIF.
> 
> The Nitro enclave crate is **excluded from the root workspace** (own lockfile/version) so root `Cargo.lock` churn cannot rotate PCRs; Rust CI now tests/lints that workspace separately. Worker/image checks use `--check-registry` instead of committed `vkeys.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6d803c558cb1f1a3b2fd3e30e29117fb17c3617. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1065, opened 2026-08-20
refactor(proofs): isolate measured dependency graph

0xOsiris:
Stacked on #1064.

Isolates the proof-measurement dependency graph so unrelated root-workspace churn cannot rotate the enclave PCRs or SP1 vkeys.

**Extract host-side registration.** `register.rs` never ran inside the enclave — it talks to a running enclave over vsock and submits `registerKey` to L1 — but was compiled into the measured crate because one `enclave` feature served three roles (guest, host prover, registration). Moving it to `world-chain-proof-nitro-register` drops the alloy provider/signer stack and tx-signer from the measured graph: **653 → 493 packages (-160)**.

**Pin measured deps in their own workspace.** `proofs/core` and `proofs/kona/client` move to `proofs/measured/{core,kona-client}` with their own `[workspace.dependencies]` and lockfile, excluded from root. Acceptance test: root `Cargo.toml` is now absent from proof-gate's enclave-input filter, so a reth/revm/tokio bump for the node side can no longer rotate measurements.

Adds `tests`/`fmt`/`clippy` steps for the new workspace — its 9 tests were previously covered by root `--workspace` and would otherwise have silently stopped running.

Measurements rotate once as a result (SP1 guest ELFs embed crate paths), so this cuts `1.0.0-rc.3`.


0xOsiris:
Closing in favor of #1073

PR #1088, opened 2026-08-26
feat(contracts): add ERC-20 staking vault

Kemperino:
## Summary

Adds a token-generic `ERC20StakingVault` so WIP-1006 proposer and challenger bonds use a configured ERC-20 instead of ETH/`DelayedWETH`, without changing the stock Dispute Game Factory. World Chain can configure the vault with WLD, while the contract remains reusable with other ERC-20 tokens.

## Changes

- Add a singleton, deposit-backed `ERC20StakingVault` with reusable available balances and delayed two-phase withdrawals.
- Lock proposer bonds from `gameCreator` during factory initialization and challenger bonds from the challenger during `challenge()`.
- Settle the complete game bond pot into reusable vault balances from `closeGame()`.
- Set the type-1006 factory init bond to zero ETH and document the ERC-20 custody and settlement model.

Bindings, broader devnet tests, and off-chain service updates land in the stacked PRs.


Kemperino:
@codex review

pkg/contracts/scripts/devnet/DeployProofSystem.s.sol
```diff
@@ -172,9 +157,10 @@ contract DeployProofSystem is Script {
         config.anchorStateRegistry = IAnchorStateRegistry(vm.envAddress("ANCHOR_STATE_REGISTRY"));
         config.systemConfig = ISystemConfig(vm.envAddress("SYSTEM_CONFIG"));
         config.proxyAdmin = IProxyAdmin(vm.envAddress("OP_CHAIN_PROXY_ADMIN"));
-        config.delayedWethDelay = vm.envOr("DELAYED_WETH_DELAY", uint256(300));
-        config.proxyAdminOwnerKey = vm.envUint("OP_CHAIN_PROXY_ADMIN_OWNER_PRIVATE_KEY");
-        config.dgfOwnerKey = vm.envUint("DGF_OWNER_KEY");
+        config.wld = IERC20(vm.envAddress("WLD_TOKEN"));
+        config.existingBondVault = IWLDStakingVault(vm.envOr("WLD_STAKING_VAULT", address(0)));
+        config.wldWithdrawalDelay = vm.envOr("WLD_WITHDRAWAL_DELAY", uint256(300));
+        config.proxyAdminOwnerKey = vm.envOr("OP_CHAIN_PROXY_ADMIN_OWNER_PRIVATE_KEY", uint256(0));
```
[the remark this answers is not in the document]
Kemperino: we adapt devnet in followup

pkg/contracts/src/dispute/interfaces/IERC20StakingVault.sol
0xOsiris, line 1: Let's put Errors into their own `Errors.sol`. Natspec all types in here.

pkg/contracts/src/dispute/ERC20StakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+/// @custom:proxied true
+/// @title WLDStakingVault
+/// @notice A one-to-one WLD ledger for WIP-1006 proposal and challenge bonds.
+contract WLDStakingVault is Initializable, IWLDStakingVault {
+    using SafeERC20 for IERC20;
+
+    /// @notice Semantic version.
+    string public constant version = "1.0.0";
+
+    uint256 internal immutable DELAY_SECONDS;
```
0xOsiris, line 49: Immutable's should not be initialized via the constructor for upgradeable contracts. As every inheriting contract will have to repeat the constructors initializer with the same value. 

You should hardcode the value, or not make it immutable and set via the initializer.

0xOsiris (changes requested):

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+/// @notice A one-to-one WLD ledger for WIP-1006 proposal and challenge bonds.
+contract WLDStakingVault is Initializable, IWLDStakingVault {
+    using SafeERC20 for IERC20;
+
+    /// @notice Semantic version.
+    string public constant version = "1.0.0";
+
+    uint256 internal immutable DELAY_SECONDS;
+
+    IERC20 public wld;
```
0xOsiris: Natspec

pkg/contracts/src/dispute/ERC20StakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+
+    IERC20 public wld;
+    ISystemConfig public systemConfig;
```
0xOsiris, line 55: Natspec all of these storage variables

pkg/contracts/src/dispute/ERC20StakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        ) {
+            revert InvalidVaultConfiguration();
+        }
+        wld = wld_;
+        systemConfig = systemConfig_;
+        disputeGameFactory = disputeGameFactory_;
+    }
+
+    /// @inheritdoc IWLDStakingVault
+    function delay() external view returns (uint256) {
```
0xOsiris, line 93: Just make `DELAY_SECONDS` public instead of internal and remove this view function. So, we can just call `DELAY_SECONDS`

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        return wld.balanceOf(address(this)) >= totalLiabilities;
+    }
+
+    /// @inheritdoc IWLDStakingVault
+    function deposit(uint256 amount) external {
+        _depositFor(msg.sender, amount);
+    }
+
+    /// @inheritdoc IWLDStakingVault
+    function depositFor(address account, uint256 amount) external {
```
0xOsiris: I personally don't think we need both of these. Better to have a simple API with only `deposit(address account, uint256 amount)` imo

pkg/contracts/src/dispute/ERC20StakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+    /// @inheritdoc IWLDStakingVault
+    function requestWithdrawal(uint256 amount) external {
+        if (amount == 0) revert InvalidWithdrawal();
+        uint256 available = availableBalance[msg.sender];
+        if (available < amount) revert InsufficientBalance(msg.sender, available, amount);
+        availableBalance[msg.sender] = available - amount;
+
+        WithdrawalRequest storage request = withdrawals[msg.sender];
+        request.amount += amount;
+        request.timestamp = block.timestamp;
```
0xOsiris, line 123: This will overwrite the timestamp on an already existent withdrawal correct?

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        if (systemConfig.paused()) revert WithdrawalPaused();
+
+        WithdrawalRequest storage request = withdrawals[msg.sender];
+        if (request.amount < amount || request.timestamp == 0) revert InvalidWithdrawal();
+        uint256 availableAt = request.timestamp + DELAY_SECONDS;
+        if (block.timestamp < availableAt) revert WithdrawalDelayNotMet(availableAt);
+
+        request.amount -= amount;
+        if (request.amount == 0) request.timestamp = 0;
+        totalLiabilities -= amount;
```
0xOsiris: Total liabilities is not actually doing anything is it? We just increase it on deposit / decrease it on finalized withdrawal but the `Request` amount is the source of truth metric for the liabilities. 

This is just a value we increment / decrement in parallel for seemingly no other reason than an `isSolvent()` API. 

I say we just remove it for simplicity

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        }
+        (IDisputeGame registered,) = disputeGameFactory.games(gameType_, rootClaim_, extraData_);
+        if (address(registered) != gameAddress || address(game.bondVault()) != address(this)) {
+            revert GameNotRegistered(gameAddress);
+        }
+    }
+
+    // Bond locking lets factory-selected game code allocate participant balances, so the factory
+    // owner must be the same governance authority that controls this vault.
+    function _assertAlignedOwners() internal view {
```
0xOsiris: Can we make this a modifier

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        if (msg.sender != address(admin) && msg.sender != admin.owner()) revert NotProxyAdminOwner(msg.sender);
+    }
+
+    function _debitAvailable(address account, uint256 amount) internal {
+        uint256 available = availableBalance[account];
+        if (available < amount) revert InsufficientBalance(account, available, amount);
+        availableBalance[account] = available - amount;
+    }
+
+    function _assertProxyAdminOwner() internal view returns (address owner) {
```
0xOsiris: Can we make this a modifier

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+
+    // Bond locking lets factory-selected game code allocate participant balances, so the factory
+    // owner must be the same governance authority that controls this vault.
+    function _assertAlignedOwners() internal view {
+        address factoryOwner = disputeGameFactory.owner();
+        address vaultOwner = proxyAdminOwner();
+        if (factoryOwner != vaultOwner) revert OwnerMismatch(factoryOwner, vaultOwner);
+    }
+
+    function _assertProxyAdminOrOwner() internal view {
```
0xOsiris: Can we make this a modifier

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        address recipient = _assertProxyAdminOwner();
+        uint256 balance = wld.balanceOf(address(this));
+        uint256 recovered = amount < balance ? amount : balance;
+        // Liabilities intentionally remain unchanged: recovery may seize backing for active
+        // games without blocking their later internal settlement.
+        wld.safeTransfer(recipient, recovered);
+        emit Recovered(recipient, recovered);
+    }
+
+    function _depositFor(address account, uint256 amount) internal {
```
0xOsiris: Remove this abstraction and just make this the logic of `deposit`. Just adds indirection without much value

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+        if (fromWithdrawal != 0) {
+            withdrawals[account].amount -= fromWithdrawal;
+            if (withdrawals[account].amount == 0) withdrawals[account].timestamp = 0;
+        }
+        availableBalance[recipient] += amount;
+        emit AccountHeld(account, recipient, amount);
+    }
+
+    /// @inheritdoc IWLDStakingVault
+    function recover(uint256 amount) external {
```
0xOsiris: Why do we want this? User's should be able to recover their own bond. The proxy admin should not be able to unilaterally steal all of the bonds.

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,305 @@
...
+
+        request.amount -= amount;
+        if (request.amount == 0) request.timestamp = 0;
+        totalLiabilities -= amount;
+        wld.safeTransfer(msg.sender, amount);
+        emit Withdrawn(msg.sender, amount);
+    }
+
+    /// @inheritdoc IWLDStakingVault
+    function lockProposerBond() external {
```
0xOsiris: We may consider a pattern where the game itself deploys the WLD vault escrow within `initialize`. 

Then the vault is no longer "trusting" that the game is in fact authentically created from the vault. The trust assumptions are implicit in the existence of the vault. Then each WLD vault is only in charge of settling a single games bonds. This may be a simpler, and more intuitive approach. 

Then it is on the proposer to `approve(game, bond)`, and the game simply deploys the ephemeral vault -> `wld.transferFrom(proposer, game, bond)`

pkg/contracts/src/dispute/ERC20StakingVault.sol
Kemperino: correct, same with DelayedWeth.

pkg/contracts/src/dispute/WLDStakingVault.sol
Kemperino: it was a security measure to double check if any inconsistency was there before recovering. but i agree, removed.

pkg/contracts/src/dispute/WLDStakingVault.sol
Kemperino: they are right now with DelayedWeth. I just kept this in line with this. Since I can see more utility than "just" proposing flow on this vault in the long run though I'd be fine removing it though.

going to remove.

pkg/contracts/src/dispute/ERC20StakingVault.sol
Kemperino: I agree with you but [DelayedWeth](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/dispute/DelayedWETH.sol) has the same.

so there is a theoretical case of APIs or similar breaking and that's why I would prefer to keep it this way!

pkg/contracts/src/dispute/ERC20StakingVault.sol
Kemperino: no. they should be - if they are intended to be protocol fixed even after a proxy upgrade while choosing it at impl deployment time. which is the case here for OP Stack.

this is the same as DelayedWeth, happy to discuss if we want to diverge from it though.

pkg/contracts/src/dispute/WLDStakingVault.sol
Kemperino: i prefer the current design. a single vault hub with overview about stakes/balances, easier fund management, easier to not miss any locked funds, lower deployment cost (since no extra create), and we have the DGF as a trust anchor either way.

also dont have to approve for every new game, which I much prefer. You just deposit into the vault and are ready to propose/challeng.

I can see follow up features being built on top of this hub as well.

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,258 @@
...
+import {IProxyAdmin} from "@optimism-bedrock/interfaces/universal/IProxyAdmin.sol";
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import {ERC1967Utils} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Utils.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {LibClone} from "@solady/utils/LibClone.sol";
+
+/// @custom:proxied true
+/// @title WLDStakingVault
+/// @notice A one-to-one WLD ledger for WIP-1006 proposal and challenge bonds.
```
0xOsiris:
```suggestion
/// @title WLDStakingVault
/// @author World Contributors
/// @notice A one-to-one WLD ledger for WIP-1006 proposal and challenge bonds.
/// @custom:security-contact security@toolsforhumanity.com
```

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,258 @@
...
+contract WLDStakingVault is Initializable, IWLDStakingVault {
```
0xOsiris:
```suggestion
contract ERC20StakingVault is Initializable, IWLDStakingVault {
```

pkg/contracts/src/dispute/WLDStakingVault.sol
```diff
@@ -0,0 +1,258 @@
...
+    using SafeERC20 for IERC20;
+
+    /// @notice Semantic version.
+    string public constant version = "1.0.0";
+
+    /// @notice Delay required between requesting and executing a WLD withdrawal.
+    uint256 internal immutable DELAY_SECONDS;
+
+    /// @notice Canonical WLD token backing the vault's internal balances.
+    IERC20 public wld;
```
0xOsiris:
```suggestion
    IERC20 public token;
```

pkg/contracts/src/dispute/interfaces/IWLDStakingVault.sol
```diff
@@ -0,0 +1,124 @@
...
+pragma solidity 0.8.28;
+
+import {IDisputeGameFactory} from "@optimism-bedrock/interfaces/dispute/IDisputeGameFactory.sol";
+import {ISystemConfig} from "@optimism-bedrock/interfaces/L1/ISystemConfig.sol";
+import {IProxyAdmin} from "@optimism-bedrock/interfaces/universal/IProxyAdmin.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/// @title IWLDStakingVault
+/// @notice Custodies WLD used by WIP-1006 proposers, challengers, and reward recipients.
+interface IWLDStakingVault {
```
0xOsiris:
```suggestion
interface IERC20StakingVault {
```

0xOsiris (approved):

PR #1092, opened 2026-08-26
feat(sp1): dynamically sized range proofs in sp1 worker

0xOsiris:
> [!NOTE]
> **Medium Risk**
> Changes core proof orchestration (multi-range chaining, session persistence, bisection, and scaled network authorization), so mis-planned ranges or validation gaps could fail jobs or produce incorrect aggregated transitions.
> 
> **Overview**
> The SP1 worker no longer proves each job as a single range proof. It **plans one or more contiguous ranges** from cumulative L2 gas (via batched `eth_getBlockByNumber`), caps block count and cycle headroom, **submits ranges concurrently**, then aggregates all range artifacts. Plans are **serialized in the job’s Stark session slot** so restarts resume in-flight network requests (legacy bare session IDs still decode as a single range).
> 
> When the SP1 Network marks a range **unexecutable**, the worker **bisects at the block midpoint** (bounded by `SP1_MAX_RANGE_SPLITS`), records **`sp1_range_bisections`**, and retries; timeouts keep the existing resubmit backoff. Aggregation requests now carry **multiple range proofs**, and **network aggregation cycle/gas limits scale linearly** with range count.
> 
> The standalone **`prove_validity` helper** accepts `split_count > 1`, pins a shared L1 head, and validates aggregation against the **first and last** range metadata. New CLI/env knobs (`SP1_CYCLES_PER_GAS`, `SP1_MAX_BLOCKS_PER_RANGE`, `SP1_MAX_RANGE_SPLITS`) and docs describe the derived gas target and per-range aggregation ceiling behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74dc8903cd7f08318bdd325b312cfd645c2c93ca. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


proofs/workers/sp1/src/backend.rs
```diff
@@ -225,31 +213,216 @@ impl<P: WorldSuccinctProver + Send + Sync, G: ProofGameProvider> Sp1Backend<P, G
...
+                    }
+                    let delay = NETWORK_REQUEST_RETRY_BACKOFFS[*attempts];
+                    *attempts += 1;
+                    self.wait_before_resubmission(SessionType::Stark, *attempts, delay)
+                        .await;
+                    plan.ranges[index].session_id = None;
+                    self.persist_plan(job, plan, BackendSessionStatus::Running, None)
+                        .await?;
+                }
+                Some(session_error @ SuccinctProverError::RequestUnexecutable { .. }) => {
```
Kemperino, line 351: can we add a metric for this split / unexecutable?

it will cost 'a lot' and we should know if this happens :)

Kemperino (approved):

proofs/workers/sp1/src/backend.rs
Kemperino: ah its there

PR #1124, opened 2026-09-08
feat(proofs): log local SP1 limit estimation progress

Kemperino:
## Summary
- add 15-minute liveness updates while the SP1 SDK performs local limit estimation
- log and measure SP1 range and aggregation witness, submission, wait, and resume phases
- log and measure Nitro enclave proving duration; remove raw L1/L2 RPC URLs from logs

All changes are host-only; guest ELFs and vkeys are unchanged.

## Validation
- `cargo fmt --check`
- `cargo check -p world-chain-proof-sp1-worker`
- `cargo check -p world-chain-proof-metrics`
- `cargo check -p world-chain-proof-nitro-worker`

`cargo test -p world-chain-proof-sp1-worker --lib` was started but stopped during dependency compilation to avoid consuming the local machine; no test result is claimed.

### DIFF
--- a/proofs/workers/sp1/src/backend.rs
+++ b/proofs/workers/sp1/src/backend.rs
@@ -310,9 +310,6 @@ impl<P: WorldSuccinctProver + Send + Sync, G: ProofGameProvider> Sp1Backend<P, G
                 {
                     Ok(range_request) => {
                         let duration = witness_started_at.elapsed();
-                        world_chain_proof_metrics::record_witness_collection(
-                            "sp1", "success", duration,
-                        );
                         tracing::info!(
                             proof_id = %request.id(),
                             start_block = range_start,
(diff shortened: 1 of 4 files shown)


proofs/workers/sp1/src/backend.rs
```diff
@@ -283,17 +293,111 @@ impl<P: WorldSuccinctProver + Send + Sync, G: ProofGameProvider> Sp1Backend<P, G
...
+                let witness_started_at = Instant::now();
+                let range_request = match self
                     .build_range_request(range_start, range_end, request)
                     .await
-                    .context("failed to build range proof request")?;
-                let session_id = self
+                {
+                    Ok(range_request) => {
+                        let duration = witness_started_at.elapsed();
+                        world_chain_proof_metrics::record_witness_collection(
```
Dzejkop: duplicate metric? It's already recorded in `build_range_request`?

Dzejkop (approved):

PR #1131, opened 2026-09-10
fix(proofs): cancel obsolete proof jobs

Kemperino:
closes https://linear.app/worldcoin/issue/PROTO-5078/stop-retries-after-security-council-proof-satisfies-game

> [!NOTE]
> **Medium Risk**
> Changes proof job lifecycle and L1 polling in a service the defender depends on; mis-timed cancellations could drop needed work, though L1 errors are designed to retain jobs.
> 
> **Overview**
> Adds **L1-aware cleanup** in the prover-service so queued work does not keep running after a game no longer needs proofs.
> 
> The status poller now takes an L1 provider (required `L1_RPC_URL` on the standalone binary; devnet wires the same loop). On each interval (default **5 minutes**, up from 30s) it multicalls `resolutionStatus` and `claimData` per game with active jobs and marks `CREATED`/`RUNNING` requests **`CANCELLED`** when the game is terminal, resolved, or the contract reports sufficient proof support (`ProposalStatus::has_sufficient_proof_support()`). Failed L1 reads leave jobs unchanged; exhausted worker attempts are still failed in the same loop.
> 
> **Cancelled** is a first-class `ProofStatus` (DB constraint + API). Cancellation revokes leases and surfaces as failed to clients, but **re-requesting a cancelled proof does not bump the failure retry counter** (unlike true failures). The defender treats cancelled like failed and re-requests lanes; game observation now keys off proposal status instead of raw bitmap counts. Tests cover cancellation/restart, poller behavior, and re-proving after a supported game is challenged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9242e6b50635502a59fd86b0b3d70c45776faf0d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


proofs/services/prover-service/migrations/20260619_proof_requests.sql
```diff
@@ -24,7 +24,7 @@ CREATE TABLE proof_requests (
     finished_at         TIMESTAMPTZ NULL,
 
     CONSTRAINT proof_requests_status_check
-        CHECK (proof_status IN ('CREATED', 'RUNNING', 'SUCCEEDED', 'FAILED')),
+        CHECK (proof_status IN ('CREATED', 'RUNNING', 'SUCCEEDED', 'FAILED', 'CANCELLED')),
```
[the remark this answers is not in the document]
piohei: Cursor is right here. You need to alter or drop/create the constraint.

proofs/services/prover-service/src/config.rs
```diff
@@ -14,7 +14,7 @@ pub const DEFAULT_MAX_RETRIES: u32 = 3;
 pub const DEFAULT_BACKEND_POLL_INTERVAL: Duration = Duration::from_secs(30);
 
 /// Default delay between status-poller scans.
-pub const DEFAULT_STATUS_POLLER_INTERVAL: Duration = Duration::from_secs(30);
+pub const DEFAULT_STATUS_POLLER_INTERVAL: Duration = Duration::from_mins(5);
```
piohei: Why do you bump it?

proofs/services/prover-service/src/store.rs
```diff
@@ -277,19 +278,48 @@ impl ProverServiceStore {
...
+        .fetch_all(&self.pool)
+        .await?;
+        games.into_iter().map(address_from_bytes).collect()
+    }
+
+    pub(crate) async fn cancel_game_proofs(&self, game: Address) -> Result<u64, sqlx::Error> {
+        let result = sqlx::query(
+            r#"
+            UPDATE proof_requests
+            SET proof_status = 'CANCELLED', job_status = 'FAILED',
```
piohei: What about setting values here using bind? like `bind(ProofStatus::Cancelled.as_str()` and `bind("game no longer requires proof support")`. Could help in later looking for status usage and ensure data formatting in database (as we use ORM here).

piohei (changes requested):
Migration comment is critical (otherwise we need to wipe database which might be fine on dev).

Kemperino:
> Migration comment is critical (otherwise we need to wipe database which might be fine on dev).

yes planning to reset it since this is all devnet phase.

piohei (approved):

proofs/services/prover-service/src/config.rs
Kemperino: this obsoletion check doesn't have to run that often. though as it's shared with the other status polling i will og back 30 secs.

proofs/services/prover-service/src/store.rs
Kemperino: thx! that's better.

